PR #1, opened 2017-01-30
Update to K8S 1.5.2 and trim down master size

alexsomesan:
@fabxc @brancz 

After this, it will spin up 1.5.2 clusters.
Also trimmed down the instance size for master since it's not going to be that busy.


brancz:
:+1: (we would also be doing good if we test the latest kops beta release, as it makes sure kubernetes components are labelled correctly, this is irrelevant to these changes, but wanted to mention it)

alexsomesan:
Yeah, i noticed there is a beta out now.
I'll give it try.

PR #2, opened 2017-02-04
Extend by declerative benchmarking scenarios

fabxc:
@alexsomesan @brancz @mxinden as discussed, this holds all state in a per-cluster directory.
All configuration is done via a YAML spec that allows setting a flexible conditions and multiple Prometheus servers to be compared. A default instrumented application is deployed to simulate load.


alexsomesan:
Looks good

PR #9, opened 2017-03-06
Simulator

mxinden:
The simulator app helps performance testing Prometheus by scaling target deployments up and down on an interval basis. It takes a configmap "k8s-simulator-config" which specifies the targets deployment name, interval and upper and lower bound. It will reload the configmap every 30 seconds.


mxinden:
@fabxc Could you try this out?

I will squash the commits before merge.

apps/simulator/main.py
```diff
@@ -0,0 +1,55 @@
...
+    global prevScalingUp
+    if prevScalingUp:
+        scaleDeployment(config["low"])
+    else:
+        scaleDeployment(config["high"])
+    prevScalingUp = not prevScalingUp
+
+getConfig()
+schedule.every(30).seconds.do(getConfig)
+schedule.every(config["interval"]).seconds.do(alternateUpDownScaling)
```
fabxc: This should probably be minutes. I can't see a case where we'd want it to be below 1 minute.

apps/simulator/main.py
```diff
@@ -0,0 +1,55 @@
...
+def alternateUpDownScaling():
+    global prevScalingUp
+    if prevScalingUp:
+        scaleDeployment(config["low"])
+    else:
+        scaleDeployment(config["high"])
+    prevScalingUp = not prevScalingUp
+
+getConfig()
+schedule.every(30).seconds.do(getConfig)
```
mxinden: @fabxc Currently the config is updated every 30 seconds to see if there are changes. Is this relevant for your testcases?

fabxc: Configuration within a run is theoretically immutable. Regardless, it cannot pick up the actual interval here. Partially updating changes is semantically odd as well. I'd just leave it out for now.

If this is ever necessary as manual intervention just killing the pod and have it be restarted gets does the same.

spec.example.yaml
```diff
@@ -23,4 +23,10 @@ prometheus:
...
   count: 6
   machineType: c4.large
-  appInstances: 180
\ No newline at end of file
+  appInstances: 180
+simulator: |
+  name: fake-webserver
+  high: 200
+  low: 100
+  interval: 10
```
fabxc: Should be "intervalMinutes" for sake of clarity.

fabxc: How about naming this "loadGenerator" instead of "simulator"?

Then a sub-section "scaler" for the properties specified here so we can add a separate "querier" section.
What seems like a good approach is having functionality for either be provided by the same container for update simplicity, but only have one of them ever running. Both are ran by just having two distinct deployments. That makes it easy to turn either on or off on demand.

fabxc:
Merging here – I'll check what we can do for querying.

Issue #10, opened 2017-03-06
Add query load generator


fabxc:
We want a way to launch queries against the running Prometheus servers to observe their behavior under load.
To be predictable and reproducible it should be part of the YAML specification. Possibly:

```yaml
loadGenerator:
  querier:
    # `name` fields of Prometheus server to actually run against
    targets:
    - name: prometheus-dev-2-0
    - name: prometheus-v1-5-2
    # Groups of queries that are launched independently.
    queryGroups:
    # Regularly show instant queries. A group is run sequentially every <interval>.
    - interval: 15s
      queries:
      - rate(some_metric[1m])
      - sum by(x)(some_metric)
    # If range window is provided, launch range queries.
    # Range queries always go from now - <duration> to now.
    - interval: 3m
      range: 2h
      queries:
      - rate(some_metric[1m])
      - sum by(x)(some_metric)

PR #11, opened 2017-03-07
Refactor scaler to accomodate querier in same script

fabxc:
@mxinden 

As described in #10 it's probably best to implement this one in the same script/container but not run them in same process.

Made a few tweaks to prepare for that and changed/fixed a few namings.
I removed the `scheduler` as it still left us with a while loop regardless. I hope that's okay with you.


mxinden:
@fabxc Just two small changes from my side. One of them fixing an accidental tmp file push by me.

Tested scaling in k8s cluster. ✔️ 
Still issues with configmap generation with template engine. But I think this is due to my local setup.

mxinden:
@fabxc Feel free to revert 1d3c2dd, or comment if I should.

fabxc:
Cool, thanks

PR #12, opened 2017-03-08
Wait for all nodes and k8s cluster before deploying k8s manifests

mxinden:
This is an old PR that was targeted to the tmpl branch - not master branch. First tmpl was merged into master and then this PR was merged into tmpl. Thereby this code never made it into master.

@fabxc Could you merge these changes if they are still relevant for the project.


fabxc:
Oops, sorry about that.

PR #13, opened 2017-03-09
Add query load generator

fabxc:
First simple version of querying. Always queries on most recent data and should be extended by proper range queries and configurable time offsets.

@mxinden


manifests/k8s/load-generator-configmap.yaml
```diff
@@ -1,9 +0,0 @@
-apiVersion: v1
```
mxinden: @fabxc Why are load-generator-{configmap,deployment}.yaml deleted?

fabxc: I collapsed them into one file which I forgot to add to the commit. Fixed now.

fabxc:
GitHub got all confused and now it shows 0 commits ahead of master... reopening

PR #14, opened 2017-03-10
Add query load generator

fabxc:


spec.example.yaml
```diff
@@ -24,12 +24,37 @@ prometheus:
 workers:
   count: 6
   machineType: c4.large
-  appInstances: 180
-loadGenerator:
```
mxinden: @fabxc Why are we configuring the *load generator* in the "workers" section. I would prefer if we keep the "loadGenerator" section. The *querier* configuration e.g. has nothing to do with the *workers*, right?

fabxc: That was accidental and wouldn't even work. Fixed

PR #15, opened 2017-03-10
Add query group name, catch connection errors and enable range queries

mxinden:
1. Add query group name
2. Try catch requests.exceptions.ConnectionError when querying prometheus intances
3. Enables specification of range queries with start, end and step parameters next to normal queries

### DIFF
--- a/apps/load-generator/main.py
+++ b/apps/load-generator/main.py
@@ -55,17 +55,17 @@ class Querier(object):
     """
     Querier launches groups of queries against a Prometheus service.
     """
-    def __init__(self, i, t, qg, hist):
+    def __init__(self, t, qg, hist):
         self.url = "http://prometheus-test-%s.default:9090/api/v1/query?" % t
         self.interval = qg["intervalSeconds"]
         self.queries = qg["queries"]
-        self.i = i
+        self.groupName = qg["name"]
         self.t = t
 
         self.query_time = hist
 
     def run(self):
-        print("run querier %s %s" % (self.t, self.i))
+        print("run querier %s %s" % (self.t, self.groupName))
 
         while True:
             start = time.time()
@@ -84,7 +84,7 @@ class Querier(object):
         dur = time.time() - start
         print("query %s %s, status=%s, size=%d, dur=%d" %(self.t, q, resp.status_code, len(resp.text), dur))
 
-        self.query_time.labels(self.t, str(self.i), q).observe(dur)
+        self.query_time.labels(self.t, str(self.groupName), q).observe(dur)
 
 
 def main():
@@ -119,7 +119,7 @@ def main():
     for t in config["querier"]["targets"]:
         i = 0
         for g in config["querier"]["queryGroups"]:
-            p = threading.Thread(target=Querier(i, t["name"], g, hist).run)
+            p = threading.Thread(target=Querier(t["name"], g, hist).run)
             p.start()
             i += 1
 
--- a/apps/load-generator/main.py
+++ b/apps/load-generator/main.py
@@ -78,13 +78,16 @@ class Querier(object):
                 time.sleep(wait)
 
     def query(self, q):
-        start = time.time()
-        resp = requests.get(self.url, params={"query": q})
-        
-        dur = time.time() - start
-        print("query %s %s, status=%s, size=%d, dur=%d" %(self.t, q, resp.status_code, len(resp.text), dur))
+        try:
+            start = time.time()
+            resp = requests.get(self.url, params={"query": q})
 
-        self.query_time.labels(self.t, str(self.groupName), q).observe(dur)
+            dur = time.time() - start
+            print("query %s %s, status=%s, size=%d, dur=%d" %(self.t, q, resp.status_code, len(resp.text), dur))
+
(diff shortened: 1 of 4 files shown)


apps/load-generator/main.py
```diff
@@ -84,7 +84,7 @@ def query(self, q):
         dur = time.time() - start
         print("query %s %s, status=%s, size=%d, dur=%d" %(self.t, q, resp.status_code, len(resp.text), dur))
 
-        self.query_time.labels(self.t, str(self.i), q).observe(dur)
+        self.query_time.labels(self.t, str(self.groupName), q).observe(dur)
```
fabxc: The `str()` can be removed then.

mxinden:
@fabxc Now adding range queries as well. Could you do a review again?

spec.example.yaml
```diff
@@ -38,24 +38,44 @@ loadGenerator:
...
-      - go_goroutines
-      - container_memory_rss
-      - kube_pod_container_info
-      - codelab_api_http_requests_in_progress
-      - codelab_api_requests_total
-    - intervalSeconds: 10
+      - type: query_range
+        queryString: go_goroutines
+        start: 1489165548
+        end: 1489165569
```
fabxc: The time range should be relative to `now`.
Fixed timestamps cause us to always query the same data. We also have to adjust them in every new deployment. More realistic would be to query a window like `[rel_end-dur:rel_end]`, where `rel_end` is a duration delta to now and `dur` the total window.

apps/load-generator/main.py
```diff
@@ -78,13 +78,22 @@ def run(self):
...
+                params["step"] = q["step"]
+            resp = requests.get(self.url + q["type"] + "?", params)
+
+            dur = time.time() - start
+            print("query %s %s, status=%s, size=%d, dur=%d" %(self.t, q, resp.status_code, len(resp.text), dur))
 
-        self.query_time.labels(self.t, str(self.i), q).observe(dur)
+            self.query_time.labels(self.t, self.groupName, q).observe(dur)
+        except requests.exceptions.ConnectionError as e:
+            print("Could not query prometheus instance %s. \n %s" %(self.url, e))
```
fabxc: We should add two counter metrics here tracking a) total queries and b) failed queries.
The `except` should probably just catch everything. I can't really think of a scenario where we want to crash the thread.

spec.example.yaml
fabxc: Looking at it, we probably also want to make "instant" vs "range" as well as the range parameters for the entire group for simplicities sake.
That should simulate things like dashboards very well where multiple widgets fire queries for the same time range at the same resolution

mxinden:
@fabxc Could you review my new changes?

apps/load-generator/main.py
```diff
@@ -112,14 +144,21 @@ def main():
...
 
-    hist = Histogram("loadgen_query_duration_seconds", "Query duration", 
-            ["prometheus", "group", "query"],
-            buckets=(0.01, 0.05, 0.1, 0.3, 0.7, 1.5, 3, 6, 12, 18, 28))
+    queryTimeHist = Histogram("loadgen_query_duration_seconds", "Query duration",
+        ["prometheus", "group", "queryString", "queryType", "start", "end", "step"],
+        buckets=(0.01, 0.05, 0.1, 0.3, 0.7, 1.5, 3, 6, 12, 18, 28))
+    totalQueriesCnt = Counter('loadgen_total_queries', 'Total amount of queries',
+        ["prometheus", "group", "queryString", "queryType", "start", "end", "step"],
+    )
```
fabxc: Counters always should end in `_total`. So `loadgen_queries_total` and `loadgen_failed_queries_total`.

apps/load-generator/main.py
```diff
@@ -112,14 +144,21 @@ def main():
         Scaler(config["scaler"], url, req_kwargs).run()
         return
 
-    hist = Histogram("loadgen_query_duration_seconds", "Query duration", 
-            ["prometheus", "group", "query"],
-            buckets=(0.01, 0.05, 0.1, 0.3, 0.7, 1.5, 3, 6, 12, 18, 28))
+    queryTimeHist = Histogram("loadgen_query_duration_seconds", "Query duration",
+        ["prometheus", "group", "queryString", "queryType", "start", "end", "step"],
+        buckets=(0.01, 0.05, 0.1, 0.3, 0.7, 1.5, 3, 6, 12, 18, 28))
```
fabxc: Adding too many buckets its expensive but as query time has such a large variance, let's extend this even further and have more precision in the upper buckets. The lower ones can be reduced somewhat, because we don't care all that much about well-performing ones.

Actually, let's just split our metrics into range and instant queries explicitly:

* `loagen_query_range_duration_seconds` with buckets `(0.01, 0.1, 0.3, 0.8, 1.2, 2.5, 5, 9, 15)`
* `loadgen_query_instant_duration_seconds` with buckets `(0.5, 1, 2.5, 4, 8, 12, 16, 20, 25, 30, 36, 45)`

fabxc:
Merging, will add those changes quickly myself.

PR #16, opened 2017-03-15
fake-webserver: Do not collect go_* and process_* metrics

beorn7:
Those collections are relatively expensive (the former reads the
memstats, which requires a stop-the-world, the latter has to read and
parse a number of files in /proc). In high-frequency-scrape scenarios,
this might have a significant impact on performance.

@fabxc as discussed.

### DIFF
--- a/apps/fake-webserver/main.go
+++ b/apps/fake-webserver/main.go
@@ -6,7 +6,7 @@ import (
 	"net/http"
 	"time"
 
-	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


fabxc:
Can we make that optional?

In CPU stats of worker nodes I can see that there's a curve in CPU usage, which might be the cummulative curve of all the timers running. So those are probably not negligible either.

While you are tweaking around anyway, could you enable the debug endpoints so we can check this further in the future.

beorn7:
If this gets more involved, I'll come back to it later. CloudNativeCon preps and "normal day job" take priority right now.

beorn7:
PHAL

fabxc:
Great, thanks 👍

PR #18, opened 2017-07-13
Add the required --subnet flag

gouthamve:
The newer versions of kops require this. make fails without.

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/prombench/18)


krasi-georgiev:
outdated as we are moving to a completely new design.

PR #20, opened 2018-02-14
*: Various updates and stability changes

brancz:
Various changes that I had to do for testing the CPU performance regressions in 2.1.

@fabxc @Gouthamve @krasi-georgiev

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,17 +1,23 @@
 CLUSTER_NAME  ?= prom-test-$(shell whoami)
-DOMAIN        ?= dev.coreos.systems
+DOMAIN        ?= monitoring.team.coreos.systems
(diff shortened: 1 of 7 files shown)


krasi-georgiev:
shouldn't  we wait and see what is the reply on https://github.com/crosscloudci/crosscloudci/issues/11 before we spend any more efforts here?

brancz:
I had these changes laying around, so for the time being I figured it's better to share in case anyone wants/needs to spin up a cluster *now*.

krasi-georgiev:
I guess if Fabian doesn't have any objections and you have access,I would say merge at will so it doesn't get lost in the pile and depending on the development with the CNCF CI can revisit this later if needed.

fabxc:
👍

PR #21, opened 2018-05-14
Update README with new location of Kubernetes manifests

jeromefroe:
It seems #2 moved the Kubernetes manifests from under `manifests/prometheus` to `manifests/k8s`. This small PR just updates the README to reflect that change.


krasi-georgiev:
we are refactoring the whole prombench so this would be outdated.

PR #25, opened 2018-07-03
Automate prombench for GKE

sipian:
PR for modifying prombench into GKE
Although I have pushed all changes in 1 commit, the entire commit history can be seen [here](https://github.com/sipian/benchmark/pull/1)


README.md
```diff
@@ -1,6 +1,14 @@
+# prometheus-test-environment
+A Kubernetes cluster preconfigured for E2E testing and benchmarking Prometheus
+
+The design details and instructions on how to run can be found [here](design.md)
```
krasi-georgiev: I think this should be like (blob/master/design.md)

README.md
```diff
@@ -1,6 +1,14 @@
...
+
 # A benchmarking tool preconfigured for testing Prometheus
 1. Create the k8s cluster described by a configuration file.
 2. Apply a k8s resource file to create all required pods,services and config maps.
+3. Create a service account with Kubernetes Admin access 
+4. Follow [this](https://github.com/coreos/prometheus-operator/issues/357#issuecomment-363498455) to get primaryemail of the serviceaccount user and add this to \_rbac.yaml
+
```
krasi-georgiev: do we still need this step?

cmd/prombench/main.go
```diff
@@ -19,35 +19,35 @@ func main() {
...
-	k8sGKECluster := k8sGKE.Command("cluster", "Create or delete k8s clusters")
-	k8sGKECluster.Command("create", "gke cluster create -a key.json  -c ../../config/cluster.yaml").
-		Action(g.ClusterCreate)
-	k8sGKECluster.Command("delete", "gke cluster delete -a key.json  -c ../../config/cluster.yaml").
-		Action(g.ClusterDelete)
+	k8sGKECluster := k8sGKE.Command("cluster", "Scale up or down a k8s clusters by creating node-pools")
+	k8sGKECluster.Command("scaleUp", "gke cluster scaleUp -a key.json  -c ../../config/node-pool.yaml").
+		Action(g.NodePoolCreate)
+	k8sGKECluster.Command("scaleDown", "gke cluster scaleDown -a key.json  -c ../../config/node-pool.yaml").
+		Action(g.NodePoolDelete)
```
krasi-georgiev: can we  keep as there is a chance to need all of them at some point.

```
cluster create
cluster delete
nodePool create
nodePool delete
```

design.md
```diff
@@ -0,0 +1,25 @@
+# Prombench
+
+The proposal for Prometheus CI project can be found [here](https://docs.google.com/document/d/1aCGHS0hOrh3LiQLuOa1EWA6knF7HmqWbhp3ev66hB7Y/edit?ouid=118160464041419930165&usp=docs_home&ths=true)
+
+The TODO list can be found [here](https://github.com/sipian/prombench/issues/5)
```
krasi-georgiev: lets more all to the github issues in the prombench repo and remove this line.

prow-files/pr-image-builder/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+		software-properties-common \
+		git \
+		make \
+		&& rm -rf /var/lib/apt/lists/*
+
+RUN curl -fsSL https://download.docker.com/linux/debian/gpg | apt-key add -
+RUN add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/debian $(lsb_release -cs) stable"
+
+RUN apt-get update
+RUN apt-get -y install docker-ce
```
krasi-georgiev: what are we using this for and can't we use : https://hub.docker.com/_/docker/ ?

README.md
sipian: The ReadME is outdated. I'll update it

manifests/7_ingress.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: ingress-prombench-9-prometheus-meta
```
sipian: I accidentally committed prombench-9 instead of prombench-{{.PR_NUMBER}}
Will fix this in next commit

krasi-georgiev: git commit --amend :)

manifests/6_grafana.yaml
```diff
@@ -0,0 +1,74 @@
...
+        #     cpu: 100m
+        #     memory: 100Mi
+        env:
+          # The following env variables set up basic auth twith the default admin user and admin password.
+          - name: GF_PATHS_PROVISIONING
+            value: "/opt/grafana-provision-files"
+          - name: GF_SECURITY_ADMIN_PASSWORD
+            value: admin
+          - name: GF_SERVER_ROOT_URL
+            value: "http://{{.INGRESS_IP}}/prombench-{{ .PR_NUMBER }}/grafana/"
```
krasi-georgiev: where do you set the INGRESS_IP var?

manifests/7_ingress.yaml
sipian: :grin:

manifests/6_grafana.yaml
sipian: There are 2 ways of setting this
- we save ingress-ip in the prow config & prow saves this as an environment variable in the prowjob
- (preferred) we use a hostname in-place of INGRESS_IP, which will be resolved by an external DNS.

README.md
sipian: But we still need a service-account to authenticate the GKE API

manifests/6_grafana.yaml
krasi-georgiev: will this need to be manually set or we query this from some API?

cmd/prombench/main.go
sipian: [bb340ed](https://github.com/prometheus/prombench/pull/25/commits/bb340ede8db305abfa0256ac1052fd406ead00b6) adds these

cmd/prombench/main.go
```diff
@@ -16,38 +16,53 @@ func main() {
...
+		Required().
+		PlaceHolder("service-account.json").
 		Short('a').
 		ExistingFileVar(&g.AuthFile)
+	k8sGKE.Flag("config", "GKE cluster-config yaml file").
+		PlaceHolder("cluster.yaml").
+		Short('c').
+		Default("config/cluster.yaml").
+		ExistingFileVar(&g.ConfigFile)
+	k8sGKE.Flag("pool", "GKE Nodepool-config yaml file").
```
krasi-georgiev: I would say lets keep both configs in `cluster.yaml` keeps it less confusing and no need for a separate flag.

sipian: I kept it separate because when benchmarking is triggered, just prombench node-pools should be made, not prow. Since [createClusterRequest](cloud.google.com/go/container/apiv1) & [createNodePoolRequest](https://godoc.org/cloud.google.com/go/container/apiv1#ClusterManagerClient.CreateNodePool) are different structs, I have made 2 different files.

manifests/6_grafana.yaml
sipian: I prefer using a hostname, something like `prombench.prometheus.io`
With this just we just need to set the INGRESS_IP at the DNS server.
So if for some reason, cluster needs to be redeployed & INGRESS_IP changes, there will be no need to change the prow-config

sipian:
@krasi-georgiev 
How about creating 1 prometheus-meta & 1 grafana to monitor all instances of prombench.
This will greatly simplify the logic of allowing anonymous user access to grafana graphs and exposing ingress paths

cmd/prombench/main.go
krasi-georgiev: my idea is to use the same file, but use the parts that a needed for the command that is executed.

krasi-georgiev:
Aaah I just noticed that you have deployed 2. We definitely need only 2 as otherwise triggering alerts and comparing results would be difficult.

cmd/prombench/main.go
krasi-georgiev: so you will parse the entyre file, but just use the `nodepools` for `nodePool create` and the `cluster` for `cluster create`

sipian:
Could you clarify what 2 things have been deployed & what 1 thing do we need?

prow-files/pr-image-builder/Dockerfile
sipian: Since we need an image containing both golang & docker, one thing needs to be manually installed.
That's why I'm installing docker manually

config/cluster.yaml
```diff
@@ -1,38 +1,33 @@
...
+        isolation: prometheus
+  - name: nodes-{{ .PR_NUMBER }}
+    initialnodecount: 1
     config:
-      machinetype: n1-standard-1
+      machinetype: n1-standard-2
       imagetype: COS
       disksizegb: 100
-      localssdcount: 1
+      localssdcount: 0
```
krasi-georgiev: maybe a comment why we don't need a ssd here?

sipian: Fixed by [6e6312f](https://github.com/prometheus/prombench/pull/25/commits/6e6312f8046b6e36bf708bcc88fcdd3952a1d5b6)

config/cluster.yaml
```diff
@@ -1,38 +1,33 @@
...
+        isolation: prometheus
+  - name: nodes-{{ .PR_NUMBER }}
+    initialnodecount: 1
     config:
-      machinetype: n1-standard-1
+      machinetype: n1-standard-2
       imagetype: COS
       disksizegb: 100
-      localssdcount: 1
+      localssdcount: 0  #SSD not needed for fake-webservers (has lesser disk operations)
```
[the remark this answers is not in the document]
sipian: No
Every VM has access to a hard-disk
This flag associates a SSD with the VM to allow for fast risk IO operation.
I don't know if it's necessary for benchmarking. But the previous prombench had it

provider/gke/gke.go
```diff
@@ -48,7 +47,7 @@ type GKE struct {
 	// These are also used when the command requires some variables that are not provided by the deployment file.
 	DeploymentVars map[string]string
 	// DeploymentFile content after substituting the variables filename is used as the map key.
-	deploymentsContent map[string][]byte
+	deploymentsContent []provider.ResourceFile
```
sipian: Using maps, the order of manifest files is not preserved.
For `-f b.yaml -f a.yaml`, we should process b.yaml first.
Also for `-f folder`, we should process in order of files present in the folder.

krasi-georgiev: yes good point , we still need to know the the filename somehow for the logging.

krasi-georgiev: yep good, point, it would be good to put a comment comment somewhere that when a folder is provided the files are  applied in alphabetical order and when multiply files are provided by the -f flag these are also applied by their order.

PR #51, opened 2018-07-25
continue execution on error

krasi-georgiev:
At the moment we can't re-apply services and
this is a workaround to contonue executing
the rest of the manifest definition.


sipian:
lgtm

PR #55, opened 2018-07-26
Replace env var with args

### LINKED ISSUE #52: When passing variables to trigger Prow jobs use arguments instead of env variables.
passing values using env variables works well only when within the same context. When you want to pass data from one application to another it is best to use explicit argument passed when starting the command.

Otherwise we need to keep track of env variable names and it is very easy to brake.

sipian:
Fixes #52

### DIFF
--- a/components/prombench/manifests/benchmark/2_loadgen.yaml
+++ b/components/prombench/manifests/benchmark/2_loadgen.yaml
@@ -73,13 +73,11 @@ spec:
       serviceAccountName: loadgen-scaler
       containers:
       - name: prom-load-generator
-        image: gcr.io/prometheus-test-204522/prometheus-bench-loadgen:v0.1.0
+        image: docker.io/sipian/prombench-loadgen:v2.0.0
         imagePullPolicy: Always
-        env: 
-        - name: PROMBENCH_NAMESPACE
-          value: prombench-{{ .PR_NUMBER }}
         args:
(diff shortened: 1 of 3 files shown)


components/prombench/manifests/benchmark/2_loadgen.yaml
```diff
@@ -73,13 +73,11 @@ spec:
       serviceAccountName: loadgen-scaler
       containers:
       - name: prom-load-generator
-        image: gcr.io/prometheus-test-204522/prometheus-bench-loadgen:v0.1.0
+        image: docker.io/sipian/prombench-loadgen:v2.0.0
```
krasi-georgiev: seems the changes in this file shouldn't belong in this PR

sipian: Since I have also replaced env var with args in loadgen, I think this is valid.

sipian: I have replaced `gcr.io/prometheus-test-204522/prometheus-bench-loadgen` with `docker.io/sipian/prombench-loadgen` because `gcr.io/prometheus-test-204522/prometheus-bench-loadgen` might become inactive after the free trial ends

krasi-georgiev:
Merging so we can continue with the other changes and avoid conflicts.

PR #56, opened 2018-07-26
Refactor /benchmark pr

sipian:
On triggering `/benchmark pr`, prometheus-master will be deployed as it is.
prometheus-pr will be deployed as follows:
- prombench image will be used in the container with args `make start-pr`
-  PR will be cloned & prometheus binaries will be built using `promu`
- `./prometheus` will be executed to start prometheus-pr

GCS is no longer needed in prow because `decorate:true` is not used anymore.


components/prombench/manifests/benchmark/prometheus-pr.yaml
```diff
@@ -0,0 +1,253 @@
...
+      securityContext:
+        runAsUser: 0
+      containers:
+      - name: prometheus
+        image: {{ .PROMETHEUS_2_IMAGE }}
+        imagePullPolicy: Always
+        command:
+        - "make"
+        args:
+        - "start-pr"
```
sipian: args is the difference between `prometheus-pr` & `prometheus-release`

Makefile
```diff
@@ -1,33 +1,84 @@
 PROMBENCH_CMD        = /bin/prombench
 DOCKER_TAG = docker.io/sipian/prombench:v2.0.0
 
-deploy:
+create-nodepool:
+	printf ">> Creating NodePools"
 	$(PROMBENCH_CMD) gke nodepool create -a /etc/serviceaccount/service-account.json \
 		-v ZONE:${ZONE} -v PROJECT_ID:${PROJECT_ID} -v CLUSTER_NAME:${CLUSTER_NAME} -v PR_NUMBER:${PR_NUMBER} \
-		-f  components/prombench/nodepools.yaml
+		-f  /prombench/components/prombench/nodepools.yaml
```
krasi-georgiev: can't we have these relative so that the makefile can be also used without building a docker image?

Makefile
```diff
@@ -1,33 +1,84 @@
...
+
+DIR := "/go/src/github.com/${REPO_OWNER}/${REPO_NAME}"
+start-pr:
+	mkdir -p /go/src/github.com
+	printf "\n\n>> Fetching Pull Request"
+	git clone https://github.com/${REPO_OWNER}/${REPO_NAME} ${DIR}
```
krasi-georgiev: I don't see any reason why we need these?
lets not use env var unless 100% needed. Just to make it more clear what is happening.

sipian: These values will not be set as env variables but Makefile variables
ex: `make start-pr REPO_OWNER=prometheus REPO_NAME=prometheus PR_NUMBER=56`

Makefile
sipian: Since `WORKDIR /prombench` is in [Dockefile](https://github.com/prometheus/prombench/blob/master/Dockerfile#L19), these can be made relative

Makefile
```diff
@@ -1,33 +1,84 @@
...
+          		 --storage.tsdb.path=/data \
+				 --web.console.libraries=${DIR}/console_libraries \
+            	 --web.console.templates=${DIR}/consoles
+
+deploy-pr: create-nodepool
+	printf ">> Deploying Prombench components for PR"
+	$(PROMBENCH_CMD) gke resource apply -a /etc/serviceaccount/service-account.json \
+		-v ZONE:${ZONE} -v PROJECT_ID:${PROJECT_ID} -v CLUSTER_NAME:${CLUSTER_NAME} -v PR_NUMBER:${PR_NUMBER} \
+		-v PROMETHEUS_1_NAME:${PROMETHEUS_1_NAME} -v PROMETHEUS_1_IMAGE:${PROMETHEUS_1_IMAGE} \
+		-v PROMETHEUS_2_NAME:${PROMETHEUS_2_NAME} -v PROMETHEUS_2_IMAGE:${DOCKER_TAG} \
```
krasi-georgiev: It is not clear where PROMETHEUS_1_NAME and PROMETHEUS_2_NAME are set , can you add some comments somewhere.

components/prombench/manifests/benchmark/prometheus-pr.yaml
```diff
@@ -0,0 +1,253 @@
...
+  - name: prometheus
+    port: 9090
+  selector:
+    app: prometheus
+    prometheus: test-{{ .PROMETHEUS_1_NAME }}
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: prometheus-test-{{ .PROMETHEUS_2_NAME }}
```
krasi-georgiev: Lets try to keep the env variables to a minimum to make it easyer to understand what is happening.
I think it is fine to hard core PROMETHEUS_1_IMAGE and PROMETHEUS_2_IMAGE to make it easyer to understand .

Makefile
krasi-georgiev: yeah my point was why do you need to set them  as args in
```
args:
        - "start-pr"
        - "REPO_OWNER={{ .REPO_OWNER }}"
        - "REPO_NAME={{ .REPO_NAME }}"
        - "PR_NUMBER={{ .PR_NUMBER }}"
```

why not just set them at the beginning of the makefile?

```
REPO_OWNER=prometheus 
REPO_NAME=prometheus
```

Makefile
sipian: Prow config has the following args set in it's config
- PROJECT_ID
- ZONE
- CLUSTER_NAME

When the start-benchmark prow-job is created, the benchmark plugin add the following args
- ACTION 
- PR_NUMBER
- PROMETHEUS_1_NAME
- PROMETHEUS_1_IMAGE
- PROMETHEUS_2_NAME
- PROMETHEUS_2_IMAGE
The values of these args are not constant are are dependent on `/benchmark pr|release`

components/prombench/manifests/benchmark/prometheus-pr.yaml
sipian: The values of these args are not constant are are dependent on /benchmark pr|release

krasi-georgiev: these are only used in w places

sipian:
Closed by #66

PR #57, opened 2018-07-27
add template function to allow normalising variable values.

krasi-georgiev:
k8s objects can't have dots(.) se we added a custom function to allow normalising the variable values.


sipian:
lgtm

PR #63, opened 2018-07-29
Dockerfile and bash script to build and run prometheus from PR

sipian:

### DIFF
--- a/temp/prometheus-builder/README.md
+++ b/temp/prometheus-builder/README.md
@@ -1,0 +1,26 @@
+### Prometheus-Builder
+
+This is used for building prometheus binaries from Pull Requests and running them on containers.  
+Prombench uses this to build binaries for the Pull Request being benchmarked.
+
+### How to run
+
+#### Kubernetes Pod
+
+A sample deployment config can be found [here](components/prombench/manifests/benchmark/3_prometheus-test.yaml#L176)
+
+#### Docker Container
+
+```
+mkdir -p config
+echo \
+"global:
+  scrape_interval:     15s
+scrape_configs:
+  - job_name: 'prometheus'
+    scrape_interval: 5s
+    static_configs:
+      - targets: ['localhost:9090']" > config/prometheus.yaml
+
+docker run --rm -p 9090:9090 -v /absolute-path/to/config:/etc/prometheus/config -v data:/data sipian/prometheus-builder:v1.0.0 <PR_NUMBER> 
+```
--- a/temp/prometheus-builder/README.md
+++ b/temp/prometheus-builder/README.md
@@ -22,5 +22,6 @@ scrape_configs:
     static_configs:
       - targets: ['localhost:9090']" > config/prometheus.yaml
 
+docker build -t sipian/prometheus-builder:v1.0.0 .
 docker run --rm -p 9090:9090 -v /absolute-path/to/config:/etc/prometheus/config -v data:/data sipian/prometheus-builder:v1.0.0 <PR_NUMBER> 
 ```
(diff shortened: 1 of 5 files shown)


temp/prometheus-builder/build.sh
```diff
@@ -0,0 +1,25 @@
...
+DIR="/go/src/github.com/prometheus/prometheus"
+
+printf "\n\n>> Fetching Pull Request\n\n"
+git clone https://github.com/prometheus/prometheus.git $DIR
+
+cd $DIR
+git fetch origin pull/$PR_NUMBER/head:pr-branch
+git checkout pr-branch
+printf "\n\n>> Creating prometheus binaries\n\n"
+make build
```
krasi-georgiev: Where is the makefile for this step?

sipian: This uses the cloned Prometheus [Makefile](https://github.com/prometheus/prometheus/blob/master/Makefile.common#L96)

krasi-georgiev: put a copy inside the same folder and later will decide what to do with this docker image.

sipian:
PTAL

krasi-georgiev:
I am getting PR NUMBER is missing in argument
docker run --rm -it prometheus-builder 4268

in my case it worked after I changed it to
ENTRYPOINT ["/go/src/github.com/build.sh"]


It would be good to add a simple README.md with a full example how to use it.
including the files/folders that need to be mounted.

sipian:
I have tested this image in minikube

krasi-georgiev:
would you mind also adding a short readme with an example how to use

krasi-georgiev:
I mean something like 
docker run -v ..... args... etc

sipian:
I'll add this in the readme.
But in prow, these things will be taken care by k8s

krasi-georgiev:
yes I meant in  temp/prometheus-builder/README.md
just a simple example so that someone looking the Dockerimage know how and what we are using it for.

sipian:
PTAL

temp/prometheus-builder/README.md
```diff
@@ -0,0 +1,26 @@
...
+"global:
+  scrape_interval:     15s
+scrape_configs:
+  - job_name: 'prometheus'
+    scrape_interval: 5s
+    static_configs:
+      - targets: ['localhost:9090']" > config/prometheus.yaml
+
+docker run --rm -p 9090:9090 -v /absolute-path/to/config:/etc/prometheus/config -v data:/data sipian/prometheus-builder:v1.0.0 <PR_NUMBER> 
+```
```
krasi-georgiev: this would be better as 
```
docker build -t prometheus-builder ./
docker run --rm -p 9090:9090 -v /absolute-path/to/config:/etc/prometheus/config prometheus-builder <PR_NUMBER>
```

temp/prometheus-builder/README.md
```diff
@@ -22,5 +22,6 @@ scrape_configs:
     static_configs:
       - targets: ['localhost:9090']" > config/prometheus.yaml
 
+docker build -t sipian/prometheus-builder:v1.0.0 .
```
krasi-georgiev: I meant making it more generic by not including your repo name in the docker path.

krasi-georgiev: if you want to include the name than make is as 
`promettheus/prometheus-builder`

if we are to keep this than it would end up in the official prometheus repo.

sipian: Oh my bad :stuck_out_tongue:

sipian: I think just `prometheus-builder` is fine for now

PR #66, opened 2018-07-30
Modify files for different benchmark comments

sipian:
Benchmark comments are now

- /benchmark (If no release version is given, master is considered by default)
- /benchmark master
- /benchmark 2.3.0
- /benchmark stop

Prow does'nt seem to require GCS_BUCKET now

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,9 +9,9 @@ RUN \
         make \
 	&& rm -rf /var/lib/apt/lists/*
 
-COPY prombench /bin/prombench
 RUN mkdir -p /prombench/components/prombench/manifests
 
+COPY prombench /prombench/prombench
 COPY Makefile /prombench/Makefile
 COPY components/prombench/nodepools.yaml /prombench/components/prombench/nodepools.yaml
 COPY components/prombench/manifests/benchmark /prombench/components/prombench/manifests/benchmark
--- a/Makefile
+++ b/Makefile
@@ -1,15 +1,6 @@
 PROMBENCH_CMD        = ./prombench
 DOCKER_TAG = docker.io/sipian/prombench:v2.0.0
 
-#Prow config has the following args set in it's configuration during deployment
-#	ZONE
-#	PROJECT_ID
-#	CLUSTER_NAME
-#
-#When the start-benchmark/cancel-benchmark prow-job is created, the benchmark plugin adds the following args
-#	PR_NUMBER
-#	RELEASE - which release version to benchmark PR with (Default: master)
-
 deploy:
 	$(PROMBENCH_CMD) gke nodepool create -a /etc/serviceaccount/service-account.json \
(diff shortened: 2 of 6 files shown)


Dockerfile
```diff
@@ -9,9 +9,9 @@ RUN \
         make \
 	&& rm -rf /var/lib/apt/lists/*
 
-COPY prombench /bin/prombench
 RUN mkdir -p /prombench/components/prombench/manifests
 
+COPY prombench /prombench/prombench
 COPY Makefile /prombench/Makefile
 COPY components/prombench/nodepools.yaml /prombench/components/prombench/nodepools.yaml
 COPY components/prombench/manifests/benchmark /prombench/components/prombench/manifests/benchmark
```
krasi-georgiev: Is there no way we can mount any config files  when starting the docker container?

Makefile
```diff
@@ -1,22 +1,28 @@
...
 DOCKER_TAG = docker.io/sipian/prombench:v2.0.0
 
+#Prow config has the following args set in it's configuration during deployment
+#	ZONE
+#	PROJECT_ID
+#	CLUSTER_NAME
+#
+#When the start-benchmark/cancel-benchmark prow-job is created, the benchmark plugin adds the following args
+#	PR_NUMBER
+#	RELEASE - which release version to benchmark PR with (Default: master)
```
krasi-georgiev: Why do you think we need these comment?

This should be  just s simple Dockerimage which expects these variables.
I think we should try to separate it as much as possible from any other notions. Otherwise need to constantly update comments when we change the logic somewhere else.

components/prow/manifests/prow_internals_2.yaml
```diff
@@ -168,7 +168,6 @@ data:
           entrypoint: "registry.svc.ci.openshift.org/ci/entrypoint:latest"
           sidecar: "registry.svc.ci.openshift.org/ci/sidecar:latest"
         gcs_configuration:
-          bucket: "{{ .GCS_BUCKET }}"
           path_strategy: "explicit"
           default_org: "{{ .GITHUB_ORG }}"
           default_repo: "{{ .GITHUB_REPO }}"
```
krasi-georgiev: do we need the rest of the gcs config?

sipian: `hook` & `plank` were crashing when I removed gcs config completely

Dockerfile
sipian: In our case, k8s will mount secret `serviceaccount` at `/etc/serviceaccount/service-account.json`, when it creates the [prowjob](https://github.com/sipian/prombench/blob/refactor/components/prow/manifests/prow_internals_2.yaml#L209)

Makefile
sipian: i have removed the entire comment

Dockerfile
krasi-georgiev: I meant something like a volume for all config files, but lets leave it like this for now and will revisit.

components/prow/manifests/prow_internals_2.yaml
krasi-georgiev: would you mind opening a an issue upstream to explain that it should crash when we don't need a GCS_BUCKET

PR #68, opened 2018-07-31
change the bench machine types and loader configs

krasi-georgiev:
It is time to start testing PR's so setting higher machine types
and higher load settings for the loadgenertor.

### DIFF
--- a/components/prow/cluster.yaml
+++ b/components/prow/cluster.yaml
@@ -10,5 +10,6 @@ cluster:
   - name: prow
     initialnodecount: 1
     config:
-      machinetype: n1-standard-2
+      machinetype: n1-standard-4
       imagetype: COS
+      disksizegb: 300
--- a/components/prombench/manifests/benchmark/2_loadgen.yaml
+++ b/components/prombench/manifests/benchmark/2_loadgen.yaml
@@ -7,9 +7,9 @@ data:
   config.yaml: |
     scaler:
       name: fake-webserver
-      high: 5
-      low: 2
-      intervalMinutes: 5
+      high: 700
+      low: 100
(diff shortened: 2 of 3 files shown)


components/prow/cluster.yaml
```diff
@@ -10,5 +10,6 @@ cluster:
   - name: prow
     initialnodecount: 1
```
sipian: Since `prow` node-pool will deploy

- prow
- prowjobs
- prometheus-meta
- nginx-ingress-pods
- grafana
- alert-manager

`initialnodecount` should be atleast 4

krasi-georgiev: why? we run all of these on the same node, The machine is big enough to handle the load.

krasi-georgiev: none of these should cause a big load.

sipian: Since all these components have `nodeSelector: cloud.google.com/gke-nodepool: prow`
I think for resilience, there should be at-least 2 nodes.
And also, if nodecount is less, we should add resource.limits so that a pod does'nt consume more resource and harm other pods.

krasi-georgiev: that is easy to scale if needed, lets keep it to a minimum and if facing issues will revisit.

krasi-georgiev: it is difficult to know what we need before running some tests so lets test and than adjust as needed. I don't like wasting resources until we are sure we need this. It might happen that it only consumes 10% of CPU and memory.

krasi-georgiev:
would you mind building a new prombench image so I can run a new test and see how it will perform.

PR #73, opened 2018-08-01
remove echo test job

krasi-georgiev:
I don't see how is this useful and it just filling up the reports screen so removing for now and for checking real prow components status can revisit again.


sipian:
:+1:

PR #74, opened 2018-08-01
added PersistentVolumeClaim for prometheus-meta

krasi-georgiev:
it requests and dedicates a hdd
attached to the prometheus-meta pod
for persisisting the data

also added the command to the prombenh tool
to support this deployment type.

### DIFF
--- a/components/prombench/manifests/results/prometheus-meta.yaml
+++ b/components/prombench/manifests/results/prometheus-meta.yaml
@@ -1,4 +1,15 @@
 apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: prometeus-meta
+spec:
+  accessModes:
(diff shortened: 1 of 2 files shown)


components/prombench/manifests/results/prometheus-meta.yaml
```diff
@@ -112,8 +121,7 @@ spec:
         args:
         - "--config.file=/etc/prometheus/config/prometheus.yaml"
         - "--storage.tsdb.path=/data"
-        - "--storage.tsdb.min-block-duration=15m"
-        - "--storage.tsdb.max-block-duration=4h"
```
krasi-georgiev: this should have been in a separate PR
but in general these are added to triger compaction more often.
Withou our long running tests don't think this is necessary.

Issue #77, opened 2018-08-01
Use a service name instead of ingress ip for grafana.


krasi-georgiev:
It would be less prone to error if we use the prometheus-meta service name instead of the ingress ip

We still want to expose prometheus-meta, but for any connection between the internal components would be best to use service names instead of external IP addresses.

geekodour:
@krasi-georgiev with ref to you comment on #188 related to using the k8s api to set the `INGRESS_IP`, do you think by replacing the `INGRESS_IP` with `prombench.prometheus.io` in the manifest files would be a good idea? then we don't even need to use the `INGRESS_IP`.

also, I think making `prombench.prometheus.io` be set using a DeploymentVariable will be even better because we'll be able to test prombench in a staging/testing environment with some other custom domain then.



```
λ  prombench (networkingIngress) × grep -rn INGRESS_IP --exclude=README.md
components/prombench/manifests/results/prometheus-meta.yaml:22:    # web.external-url of INGRESS_IP/prometheus-meta/
components/prombench/manifests/results/prometheus-meta.yaml:135:        - "--web.external-url=http://{{ .INGRESS_IP }}/prometheus-meta"
components/prombench/manifests/results/grafana_deployment.yaml:29:            value: "http://{{ .INGRESS_IP }}/grafana"
components/prombench/manifests/results/grafana_datasource.yaml:19:      url: http://{{ .INGRESS_IP }}/prometheus-meta/
```
```
λ  prombench (networkingIngress) × grep -rn prombench.prometheus.io --exclude=README.md
cmd/load-generator/main.py:46:            self.url = "http://prombench.prometheus.io/%s/prometheus-%s/api/v1/query" % (pr_number, target)
cmd/load-generator/main.py:48:            self.url = "http://prombench.prometheus.io/%s/prometheus-%s/api/v1/query_range" % (pr_number, target)
components/prow/manifests/prow_internals_2.yaml:191:      job_url_template: 'http://prombench.prometheus.io/log?job={{"{{"}}.Spec.Job{{"}}"}}&id={{"{{"}}.Status.BuildID{{"}}"}}'
components/prombench/manifests/benchmark/3_prometheus-test.yaml:763:          "--web.external-url=http://prombench.prometheus.io/{{ .PR_NUMBER }}/prometheus-pr",
[...]

PR #78, opened 2018-08-01
replace INGESS_IP with a domain name

krasi-georgiev:
this way we don't need to pass the env variable arround
and less likely to brake.
for local dev env can still use /etc/hosts
to point it to another prow installation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,31 +67,39 @@ kubectl create secret generic hmac-token --from-literal=hmac=$HMAC_TOKEN
 kubectl create secret generic service-account --from-file=service-account.json=$AUTH_FILE
 ```
 
-- Deploy all internal prow components and the [nginx-ingress-controller](https://github.com/kubernetes/ingress-nginx) which will be used to access all public components.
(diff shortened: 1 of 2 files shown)


sipian:
Since we are using domain-name, we can also add prombench.prometheus.io as a host field in [nginx-ingress-spec](https://github.com/prometheus/prombench/blob/master/components/prow/manifests/nginx-controller.yaml#L310).

Also we need to add this in [grafana_datasource](https://github.com/prometheus/prombench/blob/master/components/prombench/manifests/results/grafana_datasource.yaml#L15), [grafana_deployment](https://github.com/prometheus/prombench/blob/master/components/prombench/manifests/results/grafana_deployment.yaml#L38) & [prometheus-meta](https://github.com/prometheus/prombench/blob/master/components/prombench/manifests/results/prometheus-meta.yaml#L117) because both of them are accessed via nginx-reverse-proxy

krasi-georgiev:
I have another idea about grafana. I think it should use the service name to connect to prometheus.

https://github.com/prometheus/prombench/issues/77

as per the issue I think it is a good idea any component on the k8s cluster to use internal service names to connect to each other.

I already did a quick test and grafana is fine by using the prome..-meat service name

PR #87, opened 2018-08-02
Add node-exporter for prow nodepool and Node-Metrics grafana dashboard

sipian:


krasi-georgiev:
that would be useful 
can we show the metrics in Grafana ?

sipian:
This node-exporter is just for the prow nodes.

Current grafana dashboard has a panel for node-exporter-metrics for the nodes of 3_prometheus-test.yaml.

sipian:
a prombench instance deploys new node-exporters on the new deployed nodes.
And prometheus-servers being benchmarked scrape

- fake-web-servers
- node-exporters of the newly deployed nodes
- kubelet of the newly deployed nodes

krasi-georgiev:
sorry I didn't understand is this to show the load on the main prow node and  will it show on the same graph?

I am thinking if it is not too difficult would be better to have a separate dashboard in grafana dedicated just to this.
I am thinking if it is on the same as the others it will get too busy.

sipian:
It's not too difficult.
Need to choose a dashboard from [grafana-dashboards](https://grafana.com/dashboards/) and modify some fields.

krasi-georgiev:
ok lets do it in the same PR than.

sipian:
sample prow-node-metrics grafana graphs can be viewed [here](http://35.237.152.93/grafana/d/rVGW4V5iz/prow-node-metrics?orgId=1).
I have used [this](https://grafana.com/dashboards/405)  template from [grafana.com/dashboards](https://grafana.com/dashboards)

sipian:
Waiting for #104 to be merged before adding grafana-dashboard-prombench.yaml

components/prombench/manifests/results/node-exporter.yaml
```diff
@@ -0,0 +1,70 @@
+apiVersion: apps/v1
+kind: DaemonSet
```
krasi-georgiev: is this suppose to be a DaemonSet?

krasi-georgiev: got it thanks

krasi-georgiev:
I am a bit confused here can you give a bit more detail about each piece and maybe we can come up with some better structure and names. 

 - grafana_dashboard_prow_node.yaml - defines
   * `prow-node.json` config map which is used by `grafana-core`

 - grafana_dashboard.yaml - defines 
    * `prometheus-dashboard.yaml` config map which is used in `grafana-core`
    * `prombench.json` config map which is used in `grafana-core`

sipian:
grafana_dashboard_prow_node.yaml > prow-node.jsonis the dashboard for node-metrics
grafana_dashboard.yaml > prombench.json is the dashboard for prombench. 
grafana_dashboard.yaml > prometheus-dashboard.yaml is the dashboard  config file where we tell grafana about how to [provision dashboards](http://docs.grafana.org/administration/provisioning/#dashboards)

Ideally they can all be in 1 file. called `grafana-dashboard.yaml`.
I did not include them in one file as that will make it a really long file

krasi-georgiev:
yeah I would say keep is separately , but also separate the prombench.json in its own file to make the connection more clear.

grafana_dashboard_prow_node.yaml > prow-nodes.json is the dashboard for prow nodes (since you use a daemon set than lets use plural)
grafana_dashboard_prow_bench.yaml > bench-nodes.json is the dashboard for the bench-ed nodes

`benchmarking` is too long so lets use `bench` :stuck_out_tongue_winking_eye:

krasi-georgiev:
ping me when this is ready for another review.

sipian:
@krasi-georgiev This is finally ready for review :)
A sample grafana deployment with both the dashboards: http://35.231.161.178/grafana
In the Node-Metrics dashboard, initially a user should select the namespace from the `namespace` dropdown.
Then the `node` dropdown will be populated with the node-exporters present in this namespace.
Currently I am using the [node](https://github.com/prometheus/prombench/blob/master/components/prombench/manifests/benchmark/4_node-exporter.yaml#L14) label for listing the nodes in the dropdown. 
This label will always be different for prometheus-release and prometheis-pr.
But this might not work well if there are multiple fake-web-server nodes. 
All will have the same label-value `web-server`.

P.S. I have also added a few tags to the dashboards :)

sipian:
Such graphs will be helpful in choosing which machine-type is good for the deployment

PR #89, opened 2018-08-02
Renamings - remove test prefix and replace with bench where needed.

krasi-georgiev:
the labels would look much better as just 
`pr-4415` and `master` rather than `test-pr-4415` and `test-master`.

I see you used these to separate meta and the tested one, so where needed I replace test with bench to be less confusing. With test it looked like the deployment labels are some temporary holders.

### DIFF
--- a/components/prombench/apps/load-generator/main.py
+++ b/components/prombench/apps/load-generator/main.py
@@ -81,9 +81,9 @@ class Querier(object):
         self.step = qg.get("step", "15s")
 
         if self.type == "instant":
-            self.url = "http://prometheus-test-%s.%s:9090/api/v1/query" % (target, namespace)
+            self.url = "http://prometheus-%s.%s:9090/api/v1/query" % (target, namespace)
         else:
-            self.url = "http://prometheus-test-%s.%s:9090/api/v1/query_range" % (target, namespace)
+            self.url = "http://prometheus-%s.%s:9090/api/v1/query_range" % (target, namespace)
 
     def run(self):
         print("run querier %s %s" % (self.target, self.name))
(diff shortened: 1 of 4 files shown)


sipian:
I am a bit wary of removing `test` because `test` is used in `promql expressions` in [grafana-dashboard.yaml](https://github.com/prometheus/prombench/blob/master/components/prombench/manifests/results/grafana_dashboard.yaml)

ex:
```
Node-Metrics panel
"expr": "node_memory_MemTotal_bytes{job=\"node-exporter\",namespace=\"prombench-[[pr-number]]\",node=~\"(test+).*\"}"
```
```
Prombench panel
"expr": "rate(prometheus_tsdb_compaction_duration_seconds_sum{job=\"prometheus\",namespace=\"prombench-[[pr-number]]\",prometheus=~\"(test+).*\"}[10m])",
```

sipian:
With this change, you can remove `prometheus=~\"(test+).*\"` filter from [grafana_dashboard.yaml](https://github.com/prometheus/prombench/blob/master/components/prombench/manifests/results/grafana_dashboard.yaml).

sipian:
You can make a similar change by removing `test` in [4_node-exporter.yaml](https://github.com/prometheus/prombench/blob/master/components/prombench/manifests/benchmark/4_node-exporter.yaml).
You can then replace `node=~\"(test+).*\"` with `node=~\".+\"` in [grafana_dashboard.yaml](https://github.com/prometheus/prombench/blob/master/components/prombench/manifests/results/grafana_dashboard.yaml) (similar to [this](https://github.com/prometheus/prombench/pull/104/commits/2b2aa3fc4551d9049e1cb5ec6021a7b35c00639b#diff-e3c19a38f2d99a0ee11c44e132a6883fR11))

krasi-georgiev:
@sipian would you mind taking over this PR?

PR #91, opened 2018-08-02
Add hook service-account and appropriate RBAC role

sipian:


krasi-georgiev:
> Since future plans are to report errors instead of waiting for other ProwJobs to finish, get verb will then not be needed for hook.

Sorry I didn't understand what you mean by this?

krasi-georgiev:
I think the current behaviour is ok  with posting comments when another job is running.

PR #92, opened 2018-08-02
separate nodePoolDeletion and nodePoolDeleteStatusCheck

sipian:
Previously `c.clientGKE.DeleteNodePool` used to run in RetryUntilTrue.
This led to `FailedPrecondition` error as recurring DeleteNodePool calls are blocked by `ongoing operation message` for deletion of the same nodepool.
This PR separated deletion and checkingDeletionStatus for nodepools

### DIFF
--- a/provider/gke/gke.go
+++ b/provider/gke/gke.go
@@ -227,7 +227,7 @@ func (c *GKE) NodePoolCreate(*kingpin.ParseContext) error {
 			}
 
 			err = provider.RetryUntilTrue(
-				fmt.Sprintf("checking nodepool running status for:%v", reqN.NodePool.Name)
+				fmt.Sprintf("checking nodepool running status for:%v", reqN.NodePool.Name),
 				func() (bool, error) {
 					return c.nodePoolRunning(reqN.Zone, reqN.ProjectId, reqN.ClusterId, reqN.NodePool.Name)
 				})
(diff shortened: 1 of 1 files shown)


provider/gke/gke.go
```diff
@@ -260,7 +260,7 @@ func (c *GKE) nodePoolCreated(req *containerpb.CreateNodePoolRequest) (bool, err
 		return false, err
 	}
 	log.Printf("cluster node pool status: `%v`", rep.Status)
```
krasi-georgiev: now looking at this again should we make check what is the  rep.Status and compare it to running or created?

krasi-georgiev:
what was the problem with the current logic? 

for me deleting always worked as expected. yes it did return the precondition error check , but it didn't look to bad and the deletion worked well without any extra checks.

sipian:
The precondition error might become confusing to users. It also gives an impression that something is not right.
I moved `codes.NotFound` check from `nodePoolDeleted` funtion into a new function `nodePoolDeleting`.
No new check is added.

krasi-georgiev:
the workflow becomes more entangled with this change even the func names `nodePoolDeleted` , `nodePoolDeleting` are more confusing. 
Unless we have a problem with the execution order lets keep it as for simplicity and will revisit if needed.

krasi-georgiev:
will revisit if needed, but for now lets close

Issue #90, opened 2018-08-02
Change all port references to use port name instead 


krasi-georgiev:
just noticed that we reference the 9090 port in few places. Instead we should set a port name and reference that instead so that when it changes don't need to update all references.

```
serviceName: prometheus-meta
          servicePort: 9090
        path: /prometheus-meta
```
should change to

```
serviceName: prometheus-meta
          servicePort: web
        path: /prometheus-meta
```

krasi-georgiev:
duplicate of https://github.com/prometheus/prombench/issues/105

Issue #82, opened 2018-08-01
Use some sync to make sure both tested prometheus versions start at the same time.


krasi-georgiev:
When one Prometheus starts later it will shift the graphs which will brake the alerting rules.

Quick search shows that we might use initConatiner for that using `/ready` or `/healthy` endpoints to check the status.

something like a simple curl command - `curl localhost:9090/....`

PR #94, opened 2018-08-03
Add ready probe for the builder.

krasi-georgiev:
~~this PR includes 2 changes, but since both are in the same file and  want to avoid merge conflicts I can't think of a better way to do this.~~

 - fixes https://github.com/prometheus/prombench/issues/82 - just moved the order of the deployments and added a ready probe - this way we first start the pr builder and when Prometheus is started the probe will mark the deployment as ready and only than it will continue to deploy the release version.This reduced the graph time shift by only 2 secs 
~~- fixes https://github.com/prometheus/prombench/issues/90 - use named ports for Prometheus and just expose it as port 80. This way we even if we decide to change the port that Prometheus listens to wouldn't affect any other config.~~

UPDATE: the second change was done in https://github.com/prometheus/prombench/pull/106 so I removed these changes here.

### DIFF
--- a/components/prombench/manifests/benchmark/3_prometheus-test.yaml
+++ b/components/prombench/manifests/benchmark/3_prometheus-test.yaml
@@ -92,6 +92,95 @@ data:
         regex: "https-metrics.*"
         target_label: __scheme__
         replacement: https
+
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: prometheus-test-pr-{{ .PR_NUMBER }}
+  namespace: prombench-{{ .PR_NUMBER }}
+  labels:
+    app: prometheus
+    prometheus: test-pr-{{ .PR_NUMBER }}
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: prometheus
+      prometheus: test-pr-{{ .PR_NUMBER }}
+  template:
+    metadata:
+      namespace: prombench-{{ .PR_NUMBER }}
+      labels:
+        app: prometheus
+        prometheus: test-pr-{{ .PR_NUMBER }}
+    spec:
+      serviceAccountName: prometheus
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+          - topologyKey: kubernetes.io/hostname
+            labelSelector:
+              matchExpressions:
+              - key: app
+                operator: In
+                values:
+                - prometheus
+      securityContext:
+        runAsUser: 0
+      containers:
+      - name: prometheus
+        image: docker.io/prombench/prometheus-builder:2.0.0
+        imagePullPolicy: Always
+        # The prometheus-builder takes a while to build and run
+        # so make sure we start it before the release deployment.
+        # Mark it ready only when prometheus is started.
+        # This way we have the least time difference in the scraped metrics.
+        readinessProbe:
+          tcpSocket:
+            port: 9090
+          initialDelaySeconds: 30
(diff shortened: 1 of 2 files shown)


krasi-georgiev:
btw I tried using initContainers but wasn't working well. Still big delay before it detected the other pod as started and sometimes couldn't even resolve the other pod's hostname. I think bacause of some delay in the service's networking deployment.

components/prombench/manifests/benchmark/3_prometheus-test.yaml
```diff
@@ -207,6 +207,13 @@ spec:
                 - prometheus
       securityContext:
         runAsUser: 0
+      // The PR builder image takes a little longer to start due to closing and building te binay from scratch.
+      // So to make sure both pods start at the same time we use the init container to check when the other pod is ready.
+      // Without this the scraping graphs will show slightly time shifted to each other.
+      initContainers:
```
sipian: Should'nt this initContainer be a part of the 1st prometheus-deployment.
We want prometheus-test-{{ normalise .RELEASE }} to wait until PR has been built and started

sipian: sorry I did'nt see the 3rd commit.
where the names are swapped

krasi-georgiev: yeah I tried initContainers, but didn't work well. Still big delay. the ready probe worked best

krasi-georgiev: this is the initial commit, but I removed it later.

sipian:
If we are changing ports to named ports, we might as well do it for [prow-deck](https://github.com/prometheus/prombench/blob/master/components/prow/manifests/prow_internals_2.yaml#L459) and [default-http-backend in nginx-ingress-controller](https://github.com/prometheus/prombench/blob/master/components/prow/manifests/nginx-controller.yaml#L59)

krasi-georgiev:
yep that is a good idea, but lets do it in a separate PR  as this is already too big.

I opened another issue.

components/prombench/manifests/benchmark/3_prometheus-test.yaml
```diff
@@ -129,21 +130,25 @@ spec:
...
-          "--config.file=/etc/prometheus/config/prometheus.yaml",
-          "--storage.tsdb.path=/data",
-          "--storage.tsdb.min-block-duration=15m",
-          "--storage.tsdb.max-block-duration=4h"
-        ]
+        # The prometheus-builder takes a while to build and run
+        # so make sure we start it before the release deployment.
+        # Mark it ready only when prometheus is started.
+        # This way we have the least time difference in the scraped metrics.
+        readinessProbe:
```
sipian: readinessProbe is a good idea. Prometheus-meta won't start pulling metrics from prometheus-pr until it is ready

components/prombench/manifests/benchmark/3_prometheus-test.yaml
```diff
@@ -129,21 +130,25 @@ spec:
...
+          tcpSocket:
```
sipian: Won't the following readinessProbe be a good fit? 
```
readinessProbe:
  httpGet:
    path: /-/ready
    port: 9090
  initialDelaySeconds: 30
  periodSeconds: 1
```

krasi-georgiev: I guess if for some reason Prom doesn't start properly can have an additional Alert rule for that.

PR #96, opened 2018-08-03
show the yaml config text when it can't be parsed.

krasi-georgiev:
yaml error parsing shows the line number of the culprit.
We split the object by "---" and  parse each object separately
so the yaml error line will be relative to the current object.
With this change we also display a small part of the text
so we know where in the main file is the error.


sipian:
lgtm

Issue #97, opened 2018-08-03
Move all docker images to https://hub.docker.com/u/prombench/


krasi-georgiev:
until we figure out where to put all images lets keep them in the org I created on docker hub  `prombench/....` . We both have access to push so that can quickly push any changes.

PR #101, opened 2018-08-04
moved all dockerfile references to the prombench repo

krasi-georgiev:
fixes https://github.com/prometheus/prombench/issues/97

I also changes some docker tags to be all 2.0.0 so we match with the current prombench version.
This way we can always match the current prombench version with the image tag to be used.

All images have been uploaded to the prombench repo.


krasi-georgiev:
I have added you to the repo so can also push changes when any of the images need updating.

PR #64, opened 2018-07-29
add proxy to receive webhooks from the AM and post github comments

krasi-georgiev:
the image `docker.io/krasimir/alertmanager-github-receiver` used in this deployment includes all needed changes from `https://github.com/m-lab/alertmanager-github-receiver` to cover our use case.

I have submited PR's upstream and if these get merged will decide whether to use the upstream image or where to keep this image.

https://github.com/m-lab/alertmanager-github-receiver/pull/16
https://github.com/m-lab/alertmanager-github-receiver/pull/17 
https://github.com/m-lab/alertmanager-github-receiver/pull/18

to test 
```
kubectl apply  -f components/prow/manifests/alertmanager-github-receiver.yaml
```
```
cat <<EOF | kubectl create -f -
kind: Pod
apiVersion: v1
metadata:
  name: debugger
spec: 
  containers:
  - name: debugger
    image: debian
    command:
    - "ping"
    - "google.com" 
    env:
    - name: TERM
      value: xterm
EOF
```
```
kubectl exec -it debugger bash
msg='{
  "version": "4",
  "groupKey": "fakegroupkey",
  "status": "firing",
  "receiver": "http://localhost:9393/v1/receiver",
  "groupLabels": {"alertname": "FoobarIsBroken"},
  "externalURL": "http://localhost:9093",
  "commonAnnotations": {"issue": "4415"},
  "alerts": [
    {
      "labels": {"thing": "value"},
      "status": "firing",
      "startsAt": "2018-06-12T01:00:00Z",
      "endsAt": "2018-06-14T01:00:00Z"
    }
  ]
}'
curl -XPOST --data-binary "${msg}" http://alertmanager-github-receiver/v1/receiver
```
this will post a new comment on `issue=4415`

Signed-off-by: Krasi Georgiev <kgeorgie@redhat.com>

PR #103, opened 2018-08-04
remove newlines from hmac_token & oauth_token secret in readme

sipian:
Presence of newline gives the following error in  [alertmanager-github-receiver](https://github.com/prometheus/prombench/pull/64)
```
Get https://api.github.com/repos/sipian/prometheus/issues/12: net/http: invalid header field value "Bearer <OAUTH_TOKEN>\n" for key Authorization
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,12 +41,13 @@ export GCLOUD_SERVICEACCOUNT_CLIENTID=<client_id from the service-account.json>
 - Add all required tokens as k8s secrets.
   * hmac is used when verifying requests from github.
   * oauth is used when sending requests to the github api.
-  * gke auth is used when scaling up and down the cluster.
+  * gke auth is used when scaling up and down the cluster.  
+  *2 printfs are used to remove the newline at the end of the line.*
 ```
 ./prombench gke resource apply -a $AUTH_FILE -v ZONE:$ZONE -v CLUSTER_NAME:$CLUSTER_NAME \
 -f components/prow/manifests/secrets.yaml \
--v HMAC_TOKEN="$(echo $HMAC_TOKEN | base64)" \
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,12 +41,13 @@ export GCLOUD_SERVICEACCOUNT_CLIENTID=<client_id from the service-account.json>
 - Add all required tokens as k8s secrets.
   * hmac is used when verifying requests from github.
   * oauth is used when sending requests to the github api.
-  * gke auth is used when scaling up and down the cluster.
+  * gke auth is used when scaling up and down the cluster.  
+  *2 printfs are used to remove the newline at the end of the line.*
```
krasi-georgiev: don't think we need this extra comment.

README.md
```diff
@@ -41,12 +41,13 @@ export GCLOUD_SERVICEACCOUNT_CLIENTID=<client_id from the service-account.json>
...
 ```
 ./prombench gke resource apply -a $AUTH_FILE -v ZONE:$ZONE -v CLUSTER_NAME:$CLUSTER_NAME \
 -f components/prow/manifests/secrets.yaml \
--v HMAC_TOKEN="$(echo $HMAC_TOKEN | base64)" \
--v OAUTH_TOKEN="$(echo $OAUTH_TOKEN | base64)" \
+-v HMAC_TOKEN="$(printf $(printf $HMAC_TOKEN | base64))" \
```
krasi-georgiev: why double printf?

sipian: `printf` does not add a newline at the end of string, but piping it with base64 does.
2nd printf is to remove this newline

krasi-georgiev: this should also work

`base64 -w 0`

even echo has `echo -n` but some comments suggest that it is not consistent and some implementation don't have this flag.

sipian: done

PR #106, opened 2018-08-04
change all references to use port names instead of number

### LINKED ISSUE #105: Change all references to use  port names instead of number.
[prow-deck](https://github.com/prometheus/prombench/blob/master/components/prow/manifests/prow_internals_2.yaml#L459),  
[default-http-backend in nginx-ingress-controller](https://github.com/prometheus/prombench/blob/master/components/prow/manifests/nginx-controller.yaml#L59)  
[grafana-ingress-path](https://github.com/prometheus/prombench/blob/master/components/prow/manifests/nginx-controller.yaml#L373)

sipian:
Closes #105 

### DIFF
--- a/components/prombench/manifests/benchmark/3_prometheus-test.yaml
+++ b/components/prombench/manifests/benchmark/3_prometheus-test.yaml
@@ -142,6 +142,9 @@ spec:
           mountPath: /etc/prometheus/config
         - name: instance-ssd
           mountPath: /data
+        ports:
+        - name: prom-port
+          containerPort: 9090
       volumes:
       - name: config-volume
         configMap:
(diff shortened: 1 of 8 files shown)


components/prombench/manifests/benchmark/3_prometheus-test.yaml
```diff
@@ -243,7 +250,8 @@ spec:
   ports:
   #should be 1 port
   - name: prometheus
-    port: 9090
+    port: 80
```
krasi-georgiev: now `80` is exposed so should also change it in

this should be changed to `80`
https://github.com/prometheus/prombench/blob/e4dbeb29b9ed4c1609516ef4d8e9f1d5c37537a8/components/prow/manifests/nginx-controller.yaml#L355-L357

krasi-georgiev:
opened a new issue to look into using port names for all servicePort as well  
 https://github.com/prometheus/prombench/issues/108

Issue #95, opened 2018-08-03
Leave the tsdb blocks to the default values.


krasi-georgiev:
I think we should remove the custom settings for the blocks.  this setting is to trigger compaction more often, but I think we should be as close as possible to the defaults and just let the benchmarking tests run longer.

"--storage.tsdb.min-block-duration=15m",
"--storage.tsdb.max-block-duration=4h"

just waiting for  https://github.com/prometheus/prombench/pull/94 to be  merged  and will make this change as well.

PR #115, opened 2018-08-06
change deployment settings

krasi-georgiev:
untill we solve the problem with the scaling(didn't match node selector
error) will use 80 replicas for the fake server.

updated the prom builder probe. Looking at the logs this should work
well.

remove the custom settings for the pr buklder block retention
fixes: https://github.com/prometheus/prombench/issues/95

~~remove the settings to always pull the image for the release deployment
as these don't change so no need to download every time.~~

### DIFF
--- a/components/prombench/manifests/benchmark/2_loadgen.yaml
+++ b/components/prombench/manifests/benchmark/2_loadgen.yaml
@@ -7,8 +7,8 @@ data:
   config.yaml: |
     scaler:
       name: fake-webserver
-      high: 700
-      low: 100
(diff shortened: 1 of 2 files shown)


sipian:
We need `imagePullPolicy: Always` for release deployment because `{{.RELEASE}}` can be master too.

krasi-georgiev:
good catch , thanks, updated

sipian:
What's the reason behind changing periodSeconds to 2?

sipian:
lgtm

PR #118, opened 2018-08-07
change scaling policy and number of fake-webserver pods in loadgen

sipian:

### DIFF
--- a/components/prombench/apps/load-generator/main.py
+++ b/components/prombench/apps/load-generator/main.py
@@ -36,11 +36,11 @@ class Scaler(object):
 
     def run(self):
         while True:
-            print("scaling deployment %s down" % self.deployment)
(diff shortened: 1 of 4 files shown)


components/prombench/apps/load-generator/main.py
```diff
@@ -36,13 +36,13 @@ def __init__(self, cfg, url, req_kwargs):
 
     def run(self):
         while True:
-            time.sleep(self.interval)
-
-            print("scaling deployment %s" % self.deployment)
-
+            print("scaling deployment %s down" % self.deployment)
```
krasi-georgiev: that would read better as 
print("scaling deployment %s to min" % self.deployment)
print("scaling deployment %s to max" % self.deployment)

PR #121, opened 2018-08-08
refacotoring of the makefile to make it easyer to use locally

krasi-georgiev:
with this change I can use a custom AUTH_FILE location and use the
makefile locally.
When AUTH_FILE is not set it will use the default value.

Also broken the targets a bit to be able to remove/create nodes and
deployemts separately.

all changes are to make it easyer to use locally


sipian:
Lgtm

PR #129, opened 2018-08-12
overall bench settings update.

krasi-georgiev:
this is the current setup that works best so far.

nodecount is increased to 7 as a single k8s node is limited to 100 pods to need 7 when we want to test the system with 700 pods


sipian:
lgtm

PR #131, opened 2018-08-12
renamed nginx-controller port

krasi-georgiev:
This is a change you did when we were troubleshooting why the nginx backend fails.

Seems that everything is working even without it 

So would be nice to find out if we really need it and remove if not.

### DIFF
--- a/components/prow/manifests/nginx-controller.yaml
+++ b/components/prow/manifests/nginx-controller.yaml
@@ -35,7 +35,7 @@ spec:
           initialDelaySeconds: 30
           timeoutSeconds: 5
         ports:
-        - name: default-backend-port
+        - name: backend-port
           containerPort: 8080
         resources:
           limits:
@@ -57,7 +57,7 @@ metadata:
 spec:
   ports:
   - port: 80
-    targetPort: default-backend-port
+    targetPort: backend-port
   selector:
     app: default-http-backend
 ---


sipian:
I had made this change because when deploying nginx-server using named ports, I got a GKE error saying that named ports cannot have a name longer than 15 characters

sipian:
lgtm

krasi-georgiev:
my point was that even when the pod was crashing everything worked as expected so I am curious what is the `default-http-backend` used for?

sipian:
Acc to [this](https://github.com/kubernetes/ingress-nginx/blob/master/docs/user-guide/default-backend.md)

> The default backend is a service which handles all URL paths and hosts the nginx controller doesn't understand (i.e., all the requests that are not mapped with an Ingress).

sipian:
And default-http-backend was in the config yaml file provided by [official nginx docs](
 https://kubernetes.github.io/ingress-nginx/deploy/)

sipian:
Since this is passed as an arg  to the [nginx-server](https://github.com/prometheus/prombench/blob/master/components/prow/manifests/nginx-controller.yaml#L234) in the official [config]( https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/mandatory.yaml), I am wary of removing `default-http-backend`

krasi-georgiev:
best way to be sure is to look at the source of what it does.

I think this is the source and it is written in `go`  so a great exercise :)
https://github.com/kubernetes/ingress-nginx

krasi-georgiev:
Did a bit of research and I think it will be fine to remove it. 

https://github.com/kubernetes/ingress-nginx/blob/26ed627560e5d6e86392bdf394b8bc7efdba452c/cmd/nginx/flags.go#L49-L52

https://github.com/kubernetes/contrib/issues/1535

basically it forwards the 404 pages to that one for some compliance reasons so in our case that is not important. Lets remove it and for the flag just use  `--default-backend-service=none`

sipian:
I will test this once on a cluster.

PR #136, opened 2018-08-21
gke have updated the k8s version for the region we are using

krasi-georgiev:
it errors for 1.10.4-gke.2 as it is not available anymore for the region
we are using.


sipian:
:+1:

PR #139, opened 2018-08-26
Create k8s provider from inside the cluster and refactor loadgen-scaler to a cli tool

### LINKED ISSUE #124: Replace the current  scaler with the prombench cli to scale up and down the  replicas set.
The current python scaler is broken so instead of trying to fix it lets just use the prombench cli.

/prombench deployment apply -f fake-webserver.yaml -v REPLICAS=300
/prombench deployment apply -f fake-webserver.yaml -v REPLICAS=500

sipian:

[EDIT]
Closes #124


provider/gke/gke.go
```diff
@@ -174,6 +177,7 @@ func (c *GKE) clusterDeleted(req *containerpb.DeleteClusterRequest) (bool, error
 	return false, nil
 }
 
+// clusterDeleted checks whether a cluster has been created or not
 func (c *GKE) clusterRunning(zone, projectID, clusterID string) (bool, error) {
```
krasi-georgiev: // clusterRunning checks whether a cluster is in a running state.

provider/gke/gke.go
```diff
@@ -300,6 +304,7 @@ func (c *GKE) NodePoolDelete(*kingpin.ParseContext) error {
 	return nil
 }
 
+// nodePoolDeleted checks whether a nodepool has been deleted or not
```
krasi-georgiev: missing dots in many places.

provider/gke/gke.go
```diff
@@ -411,14 +417,28 @@ func (c *GKE) NewK8sProvider(*kingpin.ParseContext) error {
...
 	if err != nil {
 		log.Fatal("k8s provider error", err)
 	}
-	c.k8sProvider = k8s
+	c.k8sProvider = k8sProvider.New(c.ctx, k8sClientSet)
 	return nil
 }
 
+// newK8sClient returns a k8s client from GKE config that can apply and delete resources.
+func newK8sClient(config clientcmdapi.Config) (*kubernetes.Clientset, error) {
```
krasi-georgiev: it is a bit confusing that we have NewK8sClient and newK8sClient the same for K8sResourceApply and ResourceApply.

I will think how to simplify.

provider/gke/gke.go
```diff
@@ -441,6 +461,7 @@ func (c *GKE) ResourceDelete(*kingpin.ParseContext) error {
 	return nil
 }
 
+// applyTemplateVars applys golang templates to deployment files
```
krasi-georgiev: *applies

provider/k8s/k8s.go
```diff
@@ -31,23 +36,120 @@ type K8s struct {
...
+	return &K8sClient{
+		k8sProvider:    New(ctx, clientset),
+		DeploymentVars: make(map[string]string),
+		ctx:            ctx,
 	}, nil
 }
 
+// DeploymentsParse parses the deployment files and saves the result as bytes grouped by the filename.
+// Any variables passed to the cli will be replaced in the resources files following the golang text template format.
+func (c *K8sClient) DeploymentsParse(*kingpin.ParseContext) error {
```
krasi-georgiev: DeploymentsParse and applyTemplateVars is too much code to be duplicated in gke and k8s and eventually it will diverge, we need to put it somewhere we can reuse. Maybe provider/provider.go is a good place for it.

provider/gke/gke.go
krasi-georgiev: // nodePoolDeleted checks whether a nodepool has been deleted.
just like this is fine.

sipian:
@krasi-georgiev PTAL

provider/k8s/k8s.go
sipian: I have moved them to provider.go in https://github.com/prometheus/prombench/pull/139/commits/a7855f3d960e2db0d54589d29b727f50960b3e81

cmd/loadgen-scaler/scaler.go
```diff
@@ -0,0 +1,78 @@
...
+	app := kingpin.New(filepath.Base(os.Args[0]), "The Prombench-Scaler tool")
+	app.HelpFlag.Short('h')
+
+	k, err := k8s.NewK8sClient()
+	if err != nil {
+		fmt.Fprintln(os.Stderr, errors.Wrapf(err, "Error creating k8s client inside the k8s cluster"))
+		os.Exit(2)
+	}
+
+	s := scale{k8sClient: k}
```
krasi-georgiev: best to have an initialisation like `scaler.New()`. Can be in the same file just a bit more single purpose and keeping the main func as short as possible.

cmd/loadgen-scaler/scaler.go
```diff
@@ -0,0 +1,78 @@
...
+		IntVar(&s.intervalMinutes)
+
+	if _, err := app.Parse(os.Args[1:]); err != nil {
+		fmt.Fprintln(os.Stderr, errors.Wrapf(err, "Error parsing commandline arguments"))
+		app.Usage(os.Args[1:])
+		os.Exit(2)
+	}
+
+	log.Printf("Starting Prombench-Scaler")
+	for {
```
krasi-georgiev: instead of having this here can put it in  `func scaler.Scale(..)` and add it as 
```
k8sApp := app.Command("scale", `Scale a Kubernetes deployment object periodically up and down`).
		Action(s.k8sClient.DeploymentsParse).
		Action(scaler.Scale)
```

pkg/provider/k8s/k8s.go
```diff
@@ -31,23 +31,140 @@ type K8s struct {
...
+			resource, _, err := decode([]byte(text), nil, nil)
+			if err != nil {
+				return errors.Wrapf(err, "decoding the resource file:%v, section:%v...", deployment.Name, text[:100])
+			}
+			if resource == nil {
+				continue
+			}
+
+			switch kind := strings.ToLower(resource.GetObjectKind().GroupVersionKind().Kind); kind {
+			case "deployment":
```
krasi-georgiev: I would use an `if` statement here. for me a switch indicates that we might handle more cases in the future which is not the case with the scaling which can be done only for deployments.

pkg/provider/k8s/k8s.go
```diff
@@ -31,23 +31,140 @@ type K8s struct {
...
+	// DeploymentFile content after substituting the variables filename is used as the map key.
+	deploymentsContent []provider.ResourceFile
+
+	ctx context.Context
+}
+
 // New returns a k8s client that can apply and delete resources.
-func New(ctx context.Context, config clientcmdapi.Config) (*K8s, error) {
-	restConfig, err := clientcmd.NewDefaultClientConfig(config, &clientcmd.ConfigOverrides{}).ClientConfig()
+func New(ctx context.Context, clientset *kubernetes.Clientset) *K8s {
```
krasi-georgiev: Can't we  keep it as is and if `config ==nil' than initialise using the `InClusterConfig` ?

pkg/provider/k8s/k8s.go
```diff
@@ -31,23 +31,140 @@ type K8s struct {
...
+			if err != nil {
+				return errors.Wrapf(err, "decoding the resource file:%v, section:%v...", deployment.Name, text[:100])
+			}
+			if resource == nil {
+				continue
+			}
+
+			switch kind := strings.ToLower(resource.GetObjectKind().GroupVersionKind().Kind); kind {
+			case "deployment":
+				err = c.k8sProvider.deploymentScale(resource, replicas)
```
krasi-georgiev: Why did you  think we need a separate func here? 
As long as we keep the func a single purpose I think it is all good even if it is few extra lines than usual.

pkg/provider/k8s/k8s.go
```diff
@@ -31,23 +31,140 @@ type K8s struct {
...
+		return fmt.Errorf("unknown object version: %v kind:'%v', name:'%v'", v, kind, req.Name)
+	}
+	return nil
+}
+
+// K8sResourceApply iterates over all manifest files
+// and applies the resource definitions on the k8s cluster.
+//
+// Each file can contain more than one resource definition where `----` is used as separator.
+func (c *K8sClient) K8sResourceApply(*kingpin.ParseContext) error {
```
krasi-georgiev: why do we need K8sResourceApply and K8sResourceDelete ?

cmd/loadgen-scaler/scaler.go
```diff
@@ -0,0 +1,78 @@
...
+		time.Sleep(time.Duration(s.intervalMinutes) * time.Minute)
+
+		log.Printf("Scaling Deployment to %d", s.scaleDownReplicas)
+		if err := s.k8sClient.Scale(&s.scaleDownReplicas); err != nil {
```
krasi-georgiev: why not using `ResourceApply` directly? 
This will remove the need to have a separate `Scale` func which seems to overlap a lot with `ResourceApply`.

components/prombench/manifests/benchmark/1c_cluster-role-binding.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: loadgen-scaler
+  name: loadgen-scaler-{{ .PR_NUMBER }}
```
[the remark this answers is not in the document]
sipian: Since this is a clusterrolebinding, it is not namespaced.
If the {{ .PR_NUMBER }} is not used, then on deploying multiple prombench instances, previous loadgen-scaler rules will be modified to new ones.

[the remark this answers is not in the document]
krasi-georgiev: > previous loadgen-scaler rules will be modified to new ones.

Why is this a problem?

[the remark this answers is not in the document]
sipian: I had created loadgen-scaler as a cluster-role, because I used to deploy the cluster using gcloud. But now since everything uses the same service-account.json file, I can change it to namespace-role.
Then  {{ .PR_NUMBER }} won't be needed

components/prombench/manifests/benchmark/2_loadgen.yaml
```diff
@@ -73,11 +68,16 @@ spec:
       serviceAccountName: loadgen-scaler
       containers:
       - name: prom-load-generator
-        image: docker.io/prombench/load-generator:2.0.0
+        image: docker.io/prombench/loadgen-scaler:2.0.0
```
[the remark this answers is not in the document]
sipian: sgtm

components/prombench/manifests/benchmark/2_loadgen.yaml
```diff
@@ -73,11 +68,16 @@ spec:
...
         imagePullPolicy: Always
+        command:
+        - "make"
```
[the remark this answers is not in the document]
sipian: I used FROM prombench/prombench:2.0.0 in the Dockerfile so that I need not COPY fake-webser.yaml in the scaler Dockerfile.

components/prombench/manifests/benchmark/1c_cluster-role-binding.yaml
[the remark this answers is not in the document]
krasi-georgiev: nice , I like it when we can simplify :stuck_out_tongue_winking_eye:

cmd/loadgen-scaler/scaler.go
sipian: We need to set `req.Spec.Replicas = replicas` in the k8s yaml object.
I prefer setting it in the k8s object struct. Other options would be
- replace replicas field in yaml file
- Add the replicas as a template variable, take the template variabla as input & update it's value in the deploymentVars.
Both of the other options does'nt seem like a good design

krasi-georgiev: what about adding an extra parameter and using 
 ```
func (c *K8s) deploymentApply(resource runtime.Object, replicas int) error {...
```

Comment (sipian) [reply cmd/loadgen-scaler/scaler.go]: Yeah that's a good idea

File: pkg/provider/k8s/k8s.go (hunk 9 above)
Comment (sipian) [reply pkg/provider/k8s/k8s.go]: Yeah using default arguments is a good choice

File: cmd/loadgen-scaler/scaler.go (hunk 12 above)
Comment (krasi-georgiev) [reply cmd/loadgen-scaler/scaler.go]: or actually we don't even need the extra parameter just parse the file and before passing it to `deploymentApply` just modify  `resource.Spec.Replicas`

File: cmd/loadgen-scaler/Makefile [hunk 16]
```diff
@@ -0,0 +1,19 @@
+PROMBENCH_CMD        = ./scaler
+DOCKER_TAG = docker.io/prombench/loadgen-scaler:2.0.0
+
+scale:
+	$(PROMBENCH_CMD) scale -v PR_NUMBER:${PR_NUMBER} -u ${SCALE_UP_REPLICAS} -d ${SCALE_DOWN_REPLICAS} -i ${INTERVAL_MINUTES} \
+		-f components/prombench/manifests/benchmark/2_fake-webserver.yaml
+
+build:
+	@go version | grep go1.11 || exit  "Requires golang 1.11 with support for modules!"
+	@GO111MODULE=on go build scaler.go
```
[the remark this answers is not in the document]
sipian: I will modify this in a different PR

sipian:
@krasi-georgiev PTAL
I have handled all the asked changes (including the one we discussed offline).

pkg/provider/k8s/k8s.go
```diff
@@ -16,62 +17,104 @@ import (
...
+// K8s holds the fields used to generate API reqeust from within a cluster.
 type K8s struct {
 	clt *kubernetes.Clientset
+	// DeploymentFiles files provided from the cli.
+	DeploymentFiles []string
+	// Vaiables to subtitude in the DeploymentFiles.
+	// These are also used when the command requires some variables that are not provided by the deployment file.
+	DeploymentVars map[string]string
+	// DeploymentFile content after substituting the variables filename is used as the map key.
+	K8sResource []Resource
```
krasi-georgiev: update comment
and can be just as `Resources` to reduce the stutter.

and maybe this shouldn't be exported and use a getter as: K8s.Deployments()

pkg/provider/k8s/k8s.go
```diff
@@ -16,62 +17,104 @@ import (
...
+		clt:            clientset,
+		DeploymentVars: make(map[string]string),
 	}, nil
 }
 
-// ResourceApply applies manifest files.
-// The input map key is the filename and the bytes slice is the actual file content.
-// It expect files in the official k8s format.
-func (c *K8s) ResourceApply(deployments []provider.ResourceFile) error {
+// DeploymentsParse parses the k8s objects deployment files and saves the result as bytes grouped by the filename.
```
krasi-georgiev: comment should be updated. We save the k8s objects now.

pkg/provider/provider.go
```diff
@@ -1,20 +1,27 @@
...
 	GlobalRetryCount = 30
+	Separator        = "---"
 	globalRetryTime  = 10 * time.Second
 )
 
-//common struct used by k8s.go & gke.go to save name and content of input files
-type ResourceFile struct {
-	Name    string
-	Content []byte
+// Resource holds the file contents obtained after parsing deployment files.
```
krasi-georgiev: content*

... after parsing the template variables.

pkg/provider/k8s/k8s.go
```diff
@@ -16,62 +17,104 @@ import (
...
+		if len(k8sObjects) > 0 {
+			c.K8sResource = append(c.K8sResource, Resource{FileName: deployment.FileName, Objects: k8sObjects})
+		}
+	}
+	return nil
+}
 
+// ResourceApply applies manifest files.
+// The input map key is the filename and the bytes slice is the actual file content.
+// It expect files in the official k8s format.
```
krasi-georgiev: update the comment.

pkg/provider/k8s/k8s.go
```diff
@@ -113,26 +156,11 @@ func (c *K8s) ResourceApply(deployments []provider.ResourceFile) error {
 // ResourceDelete deletes all resources defined in the resource files.
 // The input map key is the filename and the bytes slice is the actual file content.
 // It expect files in the official k8s format.
```
krasi-georgiev: update the comment

pkg/provider/provider.go
```diff
@@ -32,3 +39,54 @@ func RetryUntilTrue(name string, retryCount int, fn func() (bool, error)) error
 	}
 	return fmt.Errorf("Request for '%v' hasn't completed after retrying %d times", name, retryCount)
 }
+
+// ApplyTemplateVars applies golang templates to deployment files.
```
krasi-georgiev: lower case

cmd/scaler/scaler.go
```diff
@@ -0,0 +1,114 @@
...
+			fmt.Fprintln(os.Stderr, errors.Wrapf(err, "Error scaling deployment"))
+			os.Exit(2)
+		}
+
+		time.Sleep(s.interval)
+
+		log.Printf("Scaling Deployment to %d", s.max)
+		if err := s.k8sClient.ResourceApply(maxResourceObjects); err != nil {
+			fmt.Fprintln(os.Stderr, errors.Wrapf(err, "Error scaling deployment"))
+			os.Exit(2)
```
krasi-georgiev: same

cmd/scaler/scaler.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func main() {
+
+	app := kingpin.New(filepath.Base(os.Args[0]), "The Prombench-Scaler tool")
+	app.HelpFlag.Short('h')
+
+	s := new()
+
+	k8sApp := app.Command("scale", `Scale a Kubernetes deployment object periodically up and down`).
```
krasi-georgiev: can we have a full example somewhere. don't forget to test locally that is shows as expected when typing a wrong command or scaler -h. I think prombench was done with some examples.

When we have full examples in the README.md can just have: run `scaler -h` for usage and examples. 
And no need to maintain the README.md when we update the usage and commands.

components/prombench/manifests/benchmark/1c_cluster-role-binding.yaml
```diff
@@ -1,11 +1,23 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: loadgen-scaler
+  namespace: prombench-{{ .PR_NUMBER }}
+rules:
+- apiGroups: ["apps"]
+  resources:
+  - deployments
+  verbs: ["get", "list", "update"]
+---
```
krasi-georgiev: did you try my idea to have a global role with full permissions instead of maintaining different roles for the different pieces?

components/prombench/manifests/benchmark/2_loadgen.yaml
```diff
@@ -73,11 +68,16 @@ spec:
...
-        image: docker.io/prombench/load-generator:2.0.0
+        image: docker.io/prombench/scaler:2.0.0
         imagePullPolicy: Always
+        command:
+        - "make"
         args:
-        - "scaler"
-        - "prombench-{{ .PR_NUMBER }}"
+        - "scale"
+        - "PR_NUMBER={{ .PR_NUMBER }}"
```
krasi-georgiev: a comment here to remind us why we need the PR_NUMBER

pkg/provider/gke/gke.go
```diff
@@ -70,48 +73,100 @@ func (c *GKE) NewGKEClient(*kingpin.ParseContext) error {
...
+		log.Fatalf("Couldn't parse deployment files: %v", err)
 	}
 
-	for _, name := range fileList {
-		content, err := c.applyTemplateVars(name)
+	c.gkeResource = deploymentResource
+	return nil
+}
+
+// K8SDeploymentsParse parses the k8s objects deployment files and saves the result as bytes grouped by the filename.
```
krasi-georgiev: saves as k8s objects.

pkg/provider/gke/gke.go
```diff
@@ -174,6 +230,7 @@ func (c *GKE) clusterDeleted(req *containerpb.DeleteClusterRequest) (bool, error
 	return false, nil
 }
 
+// clusterDeleted checks whether a cluster is in a running state.
```
krasi-georgiev: clusterRunning*

pkg/provider/gke/gke.go
```diff
@@ -70,48 +73,100 @@ func (c *GKE) NewGKEClient(*kingpin.ParseContext) error {
...
+		}
+		d := make(map[string]interface{})
+		if err := json.Unmarshal(content, &d); err != nil {
+			log.Fatalf("Couldn't parse auth file: %v", err)
+		}
+		v, ok := d["project_id"].(string)
+		if !ok {
+			log.Fatal("Couldn't get project id from the auth file")
+		}
+		c.DeploymentVars["PROJECT_ID"] = v
```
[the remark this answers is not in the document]
sipian: So PROJECT_ID shouldn't be read from AUTH_FILE if given ?

cmd/scaler/scaler.go
```diff
@@ -0,0 +1,114 @@
...
+	minResourceObjects := s.updateReplicas(&s.min)
+	maxResourceObjects := s.updateReplicas(&s.max)
+
+	for {
+		time.Sleep(s.interval)
+
+		log.Printf("Scaling Deployment to %d", s.min)
+		if err := s.k8sClient.ResourceApply(minResourceObjects); err != nil {
+			fmt.Fprintln(os.Stderr, errors.Wrapf(err, "Error scaling deployment"))
+			os.Exit(2)
```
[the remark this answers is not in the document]
sipian: In that case should we first scale up to max, then to min?

cmd/scaler/Dockerfile
```diff
@@ -1,6 +1,6 @@
...
+FROM alpine:3.8
 
-COPY scaler /prombench/scaler
-COPY Makefile /prombench/Makefile
+COPY scaler scaler
+COPY tmp/2_fake-webserver.yaml 2_fake-webserver.yaml
 
-WORKDIR /prombench
\ No newline at end of file
+ENTRYPOINT ["/scaler", "scale","-f", "/2_fake-webserver.yaml"]
```
krasi-georgiev: hm , this doesn't look right. The entrypoint should only include the binary name. any args should be passed at runtime.


We need to find a way to mount any config or manifest files at runtime instead of adding them to each image at build-time.

cmd/scaler/Makefile
```diff
@@ -0,0 +1,23 @@
+PROMBENCH_CMD        = ./scaler
+DOCKER_TAG = docker.io/prombench/scaler:2.0.0
+
+scale:
+	$(PROMBENCH_CMD) scale -v PR_NUMBER:${PR_NUMBER} ${SCALE_DOWN_REPLICAS} ${SCALE_UP_REPLICAS} ${INTERVAL} \
+		-f components/prombench/manifests/benchmark/2_fake-webserver.yaml
+
+build:
+	@go version | grep go1.11 || exit  "Requires golang 1.11 with support for modules!"
+	@GO111MODULE=on GOARCH=amd64 CGO_ENABLED=0 GOOS=linux go build scaler.go
```
krasi-georgiev: how about using a small golang 1.11 image from  https://hub.docker.com/_/golang/ to build the binary? 
We wouldn't need to check for modules support and devs wouldn't need to have golang installed at all.

cmd/scaler/Makefile
```diff
@@ -0,0 +1,23 @@
+PROMBENCH_CMD        = ./scaler
```
krasi-georgiev: do we really need this?

cmd/scaler/scaler.go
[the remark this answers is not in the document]
krasi-georgiev: I don't think it matters.

cmd/scaler/scaler.go
sipian: One issue is when an in-cluster k8s clientset is created https://github.com/sipian/prombench/blob/refactor-loadgen-scaler/cmd/scaler/scaler.go#L28.

If the working environment is not inside a cluster, we get the error
 `Error creating k8s client inside the k8s cluster: k8s config error: unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined
`

This error even comes for `./scaler -h` .
Even if we separate this into a separate Action, this error message will still be there.

cmd/scaler/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM alpine:3.8
+
+COPY scaler scaler
+COPY tmp/2_fake-webserver.yaml 2_fake-webserver.yaml
+
+ENTRYPOINT ["/scaler", "scale","-f", "/2_fake-webserver.yaml"]
```
krasi-georgiev: We shouldn't be  copying 2_fake-webserver.yaml inside the image as this requires a new image on every config change. Lets research what are the options to mount these at runtime.
At the end the entrypoint should be only the binary and anything else should be passed at runtime.

krasi-georgiev: it will probably be for a separate PR, but at least lets research now and open an issue.

I am thinking maybe create a  config map directly from the file on github when running each test?
This is a bit prone to error when we update the paths, but would be easy to catch and fix.

krasi-georgiev: maybe also talk to the Prow team , they might give some other ideas.

sipian: We can pass "scale","-f", "/2_fake-webserver.yaml" as k8s args.
Do you want the config map to be created automatically?
How about creating a config map with same contents as fake-webserver deployment in fake-webserver.yaml file

1 drawback will be having to change the config in 2 places.
But since they will be in the same file, maybe it won't be much of a problem

cmd/scaler/Makefile
sipian: I will fix this in #145

components/prombench/manifests/benchmark/1c_cluster-role-binding.yaml
sipian: We can use the [Prometheus cluster-role](https://github.com/prometheus/prombench/blob/master/components/prombench/manifests/results/authenticate_rbac.yaml#L8) for loadgen-scaler (after add list/update actions to it).
But often granting an object more permission than it needs is not a good idea.

cmd/scaler/Dockerfile
sipian: https://gist.github.com/sipian/7f871178c186b9c966ecebcbb6685c05
An example config using this technique.

cmd/scaler/README.md
```diff
@@ -0,0 +1,36 @@
...
+The project uses [go modules](https://github.com/golang/go/wiki/Modules) so it requires go with support for modules.
+
+```
+go build scaler.go
+// reads go.mod from the project root and downloads all dependancies.
+```
+
+## Usage
+```
+// (Note: These commands should be executed inside a k8s container)
```
krasi-georgiev: should also mention an example of how to set the corret rbac permission

kind: Role
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: loadgen-scaler
  namespace: prombench-{{ .PR_NUMBER }}
rules:
- apiGroups: ["apps"]
  resources:
  - deployments
  verbs: ["get", "list", "update"]
---

cmd/scaler/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+This tool uses [k8s provider](../../pkg/provider/k8s) to scale a deployment up and down periodically, from within a k8s cluster.
+
+## Build
+The project uses [go modules](https://github.com/golang/go/wiki/Modules) so it requires go with support for modules.
+
+```
+go build scaler.go
+// reads go.mod from the project root and downloads all dependancies.
+```
```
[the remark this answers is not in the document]
sipian: Well they do have the same dependencies

Issue #132, opened 2018-08-12
Update the Prom config to have 50 jobs


krasi-georgiev:
@grobie mentioned at Promcon that we should be able to replicate https://github.com/prometheus/prometheus/issues/4124 by increasing the number of jobs so lets try to increase to 50 jobs but still same targets count 300 pods and 1500 targets, 500 pods and 3500 targets

PR #141, opened 2018-08-28
config-update to use many jobs with small number of pods

krasi-georgiev:
closes: #132

last SD bug was trigered with a small number of pods and a large number
of jobs.

this also allows us to use a single node for the fake-webserver.

This config is tested to utulize the fake-webserver node at 80% at 20
pods and the Prometheus nodes at around 60% avarage(we need some room
for the compaction processes).

Updated the scrape_interval: to 5s as this is closer to a real
production environment.

### DIFF
--- a/components/prombench/manifests/benchmark/2_loadgen.yaml
+++ b/components/prombench/manifests/benchmark/2_loadgen.yaml
@@ -7,8 +7,8 @@ data:
   config.yaml: |
     scaler:
       name: fake-webserver
-      high: 700
-      low: 300
+      high: 20
+      low: 1
       intervalMinutes: 15
     querier:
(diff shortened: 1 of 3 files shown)


krasi-georgiev:
btw I have already pushed the new prombench image that includes these changes.

sipian:
Since the aim of the project is benchmarking, I think we should have both many jobs and high load.

krasi-georgiev:
with this change the total number of targets is  5000 which is even more than what we had before (3500)

krasi-georgiev:
20 pods * 5 ports per pod * 50 jobs

sipian:
:+1:

PR #143, opened 2018-08-29
moved the apps under cmd and the provider packages under pkg

krasi-georgiev:


sipian:
:+1:

PR #144, opened 2018-09-06
Don't wait for namespace to be deleted

### LINKED ISSUE #119: When deleting the namepsace do we need to wait until it is confirmed to be deleted?
To avoid the long delay why don't we just send the delete request and continue with deleting the nodepools?

I think it will be deleted in the background and even if fails there is nothing we can do in the automation so maybe there is no reason to check the status to make sure it is deleted before continuing

sipian:
Closes #119 

### DIFF
--- a/pkg/provider/k8s/k8s.go
+++ b/pkg/provider/k8s/k8s.go
@@ -852,7 +852,7 @@ func (c *K8s) namespaceDelete(resource runtime.Object) error {
 		if err := client.Delete(req.Name, &apiMetaV1.DeleteOptions{PropagationPolicy: &delPolicy}); err != nil {
 			return errors.Wrapf(err, "resource delete failed - kind: %v, name: %v", kind, req.Name)
 		}
-		log.Printf("resource deleting - kind: %v , name: %v", kind, req.Name)
+		log.Printf("resource deleted - kind: %v , name: %v", kind, req.Name)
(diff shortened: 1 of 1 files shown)


krasi-georgiev:
did you test if this works well?

sipian:
Since we are not waiting for namespace be deleted, nodepool_delete is taking a lot of time to delete a node pool and timing out.
Due to this, I think for now we can leave this condition as it is.
Later we can come up with a more elegant solution

krasi-georgiev:
ok lets keep this open as a reminder and will revisit.

geekodour:
@krasi-georgiev I think we can close this without merging as we need to make sure that the namespace got deleted with its resources before starting to delete the nodepool.

krasi-georgiev:
@geekodour 	can you remind me why do we need to wait?

geekodour:
deleting the namespace also deletes the related k8s resources to that namespace. so we're also waiting for all the resources to be deleted when we're waiting namespace to be deleted. because otherwise `nodepool_delete` times out sometimes if the resources in the namespace are not cleared or are getting modified in some manner.

krasi-georgiev:
ok so it seems that even if the namespace deletion doesn't wait this doesn't save any time in the overall node pool deletions so no point in this change.

PR #149, opened 2018-09-14
update diagram

### LINKED ISSUE #140: update the diagram
from 
https://drive.google.com/open?id=11XQeqAHypRTqJTeck0fUpEAC93kUy9Rv

sipian:
Closes #140 
I have changed the color of the namespace.
People might get confused with the green color. They might think it is a prowjob.


sipian:
Should we also push the draw.io XML file to GitHub?

It might be hard to find in drive after a few months.

krasi-georgiev:
yep good idea. name it design.xml to be the same as the others.

PR #150, opened 2018-09-17
Update use of prometheus.NewProcessCollector

beorn7:
Also, fix the `go vet` errors.

### DIFF
--- a/cmd/fake-webserver/main.go
+++ b/cmd/fake-webserver/main.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/pprof"
-	"os"
 	"time"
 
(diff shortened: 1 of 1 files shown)


krasi-georgiev:
Thanks!
Would you mind just adding the `prometheus.NewProcessCollector` changes and will address the go vet warning in a different PR as we have some refactoring pending in https://github.com/prometheus/prombench/pull/139 and would be best to fix the vet warnings after that one is merged to avoid merge confilcts.

beorn7:
I reduced this to just the ProcessCollector commit.

I guess #152 will catch those minor go vet errors anyway. (I only fixed them because I wanted to run go vet to make sure I'm using client_golang correctly, and then I thought I pick those up on the way.)

krasi-georgiev:
Thanks

Issue #35, opened 2018-07-13
Implement a better logger - go kit log


krasi-georgiev:
at the moment we are using the "log" package , but this is missing some nice features like showing the file name and line position where the log came from.

dimitrovvlado:
@krasi-georgiev 
I know it doesn't exactly follow the description of the issue (introducing a new dep), but it is possible to display the source file and line number with the standard logger as proposed here: https://github.com/prometheus/prombench/pull/153

krasi-georgiev:
closed in https://github.com/prometheus/prombench/pull/153

PR #153, opened 2018-09-18
Setting Lshortfile flag to logger

dimitrovvlado:
Setting Lshortfile flag allows final file name and line number to be displayed when using the go's standard log. A sample output:
```
2018/09/18 17:00:25 main.go:9: Testing
```
Addresses: https://github.com/prometheus/prombench/issues/35

### DIFF
--- a/cmd/prombench/prombench.go
+++ b/cmd/prombench/prombench.go
@@ -3,6 +3,7 @@ package main // import "github.com/prometheus/prombench/cmd/prombench"
 import (
 	"fmt"
 	"os"
+	"log"
 	"path/filepath"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


krasi-georgiev:
Thanks a lot for this. @dimitrovvlado 

one of the few occasions where the solution is more simple than expected :stuck_out_tongue_winking_eye: 

can we have `Llongfile` instead of `Lshortfile`. This way it is clickable in vscode and takes me to the right spot in the file for this log.

krasi-georgiev:
btw in the issue description best to use closes: #35 so that when I merge it also auto closes the issue.

thanks again!

dimitrovvlado:
> one of the few occasions where the solution is more simple than expected 😜
I don't mind switching to another logger if more requirements emerge :)

Issue #39, opened 2018-07-23
Remove design.md and add the information to the main README.md


krasi-georgiev:

sipian:
The schema diagram in the readme pretty much sums up design.md. 
Do we really need to add design.md information to readme.md?
The information that I think can be worth mentioning is the link to the [proposal](https://docs.google.com/document/d/1aCGHS0hOrh3LiQLuOa1EWA6knF7HmqWbhp3ev66hB7Y/edit?ouid=118160464041419930165&usp=docs_home&ths=true)

krasi-georgiev:
I want to add something like a - The life of a prombench job.
It  will describe step by step what happens from /bench .... trough GKE nodepool create , prometheus-meta , grafana , alertmanager , alertmanager-github-bridge etc....

PR #155, opened 2018-09-18
Delete design.md

sipian:
Since design.md is outdated, we should delete this.
It can lead to confusion. 
I am working on a blog-post for prombench. (We can use that to resolve https://github.com/prometheus/prombench/issues/39)


krasi-georgiev:
ok 

I think we should stil keep a tl;dr in a design.md document here which we can update when needed.

PR #156, opened 2018-09-18
Fix link to deployment config in prometheus-builder readme

sipian:

### DIFF
--- a/temp/prometheus-builder/README.md
+++ b/temp/prometheus-builder/README.md
@@ -7,7 +7,7 @@ Prombench uses this to build binaries for the Pull Request being benchmarked.
 
 #### Kubernetes Pod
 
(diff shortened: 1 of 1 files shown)


temp/prometheus-builder/README.md
```diff
@@ -7,7 +7,7 @@ Prombench uses this to build binaries for the Pull Request being benchmarked.
 
 #### Kubernetes Pod
 
-A sample deployment config can be found [here](components/prombench/manifests/benchmark/3_prometheus-test.yaml#L176)
+A sample deployment config can be found [here](../../components/prombench/manifests/benchmark/3_prometheus-test.yaml#L736)
```
krasi-georgiev: Specifying links with line number would be very easy to break.

there isn't anything special about the deployment so I would say can even remove the example link.
The docker example  is  good enough I think.

PR #161, opened 2018-09-26
Add go tool tests to Makefile

### LINKED ISSUE #152: Fix and add test for go vet errors.
This would be a good test to add as a first with travis.

first suggested in: https://github.com/prometheus/prombench/pull/150

this might be helpful for other tests.
https://github.com/diazjf/fakeclient

sipian:
Closes #152 

I have kept the go tool tests similar to the one used by the Prometheus repo.
As suggested by brian, promu is used to build binaries.

Once these changes are approved, I can add travis CI to the repo.

I have added .gitignore because binaries will be built using `make build` now.


.gitignore
```diff
@@ -0,0 +1,3 @@
+/prombench
+/scaler
+/fake-webserver
```
krasi-georgiev: lets try without a .gitignore file as everyone has a different workflows.
use .git/info/excludes instead.

btw you committed some binary files.

krasi-georgiev (commented):
lets also setup the CI with this so we can see it passing.

I am not sure between travis or circle, but @simonpasquier did some research recently so can give an advice.

.gitignore
sipian: I deleted fake-webserver binary file.
It had been pushed some time back.

sipian: I added .gitignore because if we make changes to the code and want to build it locally using `make`.
promu will always build the 3 binaries.
This will remain same for every workflow.
prometheus repo also has .gitignore for binaries built during `make`

simonpasquier:
As I understand, the CI will eventually build container images and push them to some registry. If yes, almost all repositories use CircleCI for this and I'd advocate for it. @sipian feel free to ping me if you have questions on how to set it up.

.gitignore
krasi-georgiev: > I deleted fake-webserver binary file.
It had been pushed some time back.
 
yes sorry , my bad.

> I added .gitignore because if we make changes to the code and want to build it locally using make.
promu will always build the 3 binaries.
This will remain same for every workflow.
prometheus repo also has .gitignore for binaries built during make

still I prefer if we don't add a .gitignore if this can be avoided by simply adding this to  .git/info/excludes.

krasi-georgiev:
@sipian actually why not use the Makefile.common?

sipian:
Yes we can use Makefile.commmon

Makefile
```diff
@@ -1,10 +1,26 @@
...
-DOCKER_TAG = docker.io/prombench/prombench:2.0.0
+PROMBENCH_CMD := ./prombench
+DOCKER_REPO := prombench
+DOCKER_IMAGE_NAME := prombench
+DOCKER_IMAGE_TAG := 2.0.0
 
+include Makefile.common
+
+# This is to prevent go get promu, staticcheck & govendor from updating go.mod 
+export GO111MODULE = off
```
simonpasquier: wouldn't it be better to modify the original `Makefile.common` file? IIUC you'd never want for any project that `go get {promu,statitccheck,govendor}` modifies the local `go.mod` so the corresponding `go get` commands in `Makefile.common` should be prefixed with `GO111MODULE=off`.

Makefile
```diff
@@ -1,10 +1,26 @@
...
+
+.PHONY: build
+build: promu
+	@echo ">> building binaries"
+	@go version | grep go1.11 || exit  "Requires golang 1.11 with support for modules!"
+	@GO111MODULE=on $(PROMU) build
```
simonpasquier: Same here, it would be better to have `Makefile.common` being able to deal with Go modules.

sipian:
@simonpasquier
Thanks

simonpasquier:
@sipian you can use `Makefile.common` from the Prometheus master branch which now supports Go modules.

sipian:
Sure Will do. Thanks.

krasi-georgiev:
anything blocking to continue with this?

sipian:
AFAIR Nothing is blocking.
I plan to return to it in a day or two.

CrazyByDefault:
Is this PR still active? This would close #152 and looks good to go

krasi-georgiev:
I guess the [Makefile.common](https://github.com/prometheus/prometheus/blob/master/Makefile.common) has been updated few times since than so worth updating it again.

If this PR is to close  #152 I think should also add the first CircleCI config to the same PR to make sure the tests passes. Otherwise can do it in a separate PR but this one shouldn't close #152 in that case.

I think @sipian 	wouldn't mind if you want to take over @CrazyByDefault 	(nice nickname btw :smile: )

sipian:
@CrazyByDefault 
Yeah, you can go ahead with the PR

krasi-georgiev:
will continue the work in 
https://github.com/prometheus/prombench/pull/177

PR #172, opened 2018-12-06
Don't add config.file arg for releases >= 2.6.0

sipian:
A simple hack for the changed Dockerfile in 2.6.0 (Ref https://github.com/prometheus/prometheus/pull/4796)

Signed-off-by: sipian <cs15btech11019@iith.ac.in>


krasi-georgiev:
can you just fix it so that it works for 2.6 and onward. It is very unlikely to need to test against anything older than 2.6 so don't think it is worth it to pollute the code with something like this.

will just manually fix it for the current bench and it wont be needed for the next one when running tests against the 2.6 release. If it comes up again will revisit.

`--log.level=debug` is nice , we should keep that.

sipian:
Have removed the release >= 2.6.0 hack.
Have added a custom command now, instead of arg, to prevent any Dockerfile changes giving problems to prombench.
Have tested it successfully with 2.5.0 and master.

krasi-georgiev (approved):

Issue #154, opened 2018-09-18
Add random restarts to check how it recovers


krasi-georgiev:
To catch bugs that happen only after a restart. Sometimes after running Prometheus for long time followed by a restart doesn't recover.

Like https://github.com/prometheus/prometheus/issues/4603

krasi-georgiev:
Oh maybe that one was on upgrade.

geekodour:
Will this not affect the running benchmarks run by the querier and scaler? or it's about affecting the benchmarks? I mean, In case if the prometheus server does not recover the benchmarks will not be what would be expected.

krasi-georgiev:
if both running Prometheus are restarted at the same time it will not affect the bench tests.

The idea os that killing Prometheus even when it is in the middle of compaction or some other operation shouldn't prevent the next normal startup. 

This happened on few occasions.

The graphs would be reset with these restarts, but that is not important as we just compare both versions so if these should look  similar when both are restarted at the same time.

geekodour:
oh yes, I see. So, for this I think deleting the pods for Prometheus Master and Prometheus PR should work ?

geekodour:
I've not tested this yet but this is a sleep timer using `$RANDOM` . I've explored other ways like k8s cronjobs and liveness but I think this should do.

modifying `components/prombench/manifests/benchmark/3_prometheus-test.yaml`

```yaml
      containers:
      - name: prometheus
        image: quay.io/prometheus/prometheus:{{ .RELEASE }}
        imagePullPolicy: Always
        command: [ "/bin/sh", "-c" ]
        args:
        - runwithrestarts(){
            /bin/prometheus \
            --web.external-url=http://prombench.prometheus.io/{{ .PR_NUMBER }}/prometheus-release \
            --storage.tsdb.path=/prometheus \
            --web.console.libraries=/etc/prometheus/console_libraries \
            --web.console.templates=/etc/prometheus/consoles \
            --config.file=/etc/prometheus/prometheus.yml \
            --log.level=debug &;
            prompid=$!;
            sleep $[ ($RANDOM % 100) + 45 ]m;
            kill -9 $prompid;
            runwithrestarts;
          }
          runwithrestarts;
        #args: [
        #  "--web.external-url=http://prombench.prometheus.io/{{ .PR_NUMBER }}/prometheus-release",
        #  "--storage.tsdb.path=/prometheus",
        #  "--web.console.libraries=/etc/prometheus/console_libraries",
[...]

PR #180, opened 2019-03-27
[WIP] Added restarter

geekodour:
See #154 

Updating a metadata in the manifest and applying update restarts the prometheus instances.


krasi-georgiev:
Thanks
haven't looked at the code yet, but just to be clear it needs to restart both instances at the same time and should kill the prometheus binary and not a graceful restart

krasi-georgiev:
about the DCO 
https://github.com/prometheus/prometheus/wiki/DCO-signing

geekodour:
I think k8s sends a `SIGTERM` so might be a graceful restart. I am not sure, I'll look into it and make changes.

krasi-georgiev:
yes it needs to be a kill signal and it should be at the same time on both instance and that both should be started again at the same time so that both graph align properly after the restart.

krasi-georgiev:
Is there a blocker here?

geekodour:
yes #184 .

If #184 gets merged the same restart script will work for both the instances as they are run in the same way. this PR is not yet review ready I think.

geekodour:
Still have to add restart on the tsdb metrics, otherwise this is working fine. 

Alternative way to do the same will use `terminationGracePeriodSeconds:0` and then applying a new deployment. Let me know if that'll be more preferable than this.

also, should I squash the commits to one?

components/prombench/manifests/benchmark/3b_prometheus-test.yaml
```diff
@@ -0,0 +1,306 @@
...
+data:
+  restarter.sh: |
+    #!/bin/sh
+    echo "Restarting prometheus"
+    kill -SIGHUP $(pidof start.sh)
+  killer.sh: |
+    #!/bin/sh
+    echo "Killing prometheus"
+    kill -9 $(pidof prometheus)
+  fetchtini.sh: |
```
geekodour: we can move the tini download part to the prometheus-builder image
then run prometheus-builder for both deployments

- the release deployment should only copy tini to the required directory
- the pr deployment will work same + copy tini to the required directory

this way we won't need to run this script from k8s and reduce the number of
initContainers by 1 for pr deployment

components/prombench/manifests/benchmark/3_prometheus-test.yaml
```diff
@@ -698,6 +698,68 @@ data:
...
+    echo "Downloading tini init system"
+
+    if [ -z $VOLUME_DIR ]; then
+        echo "environment variables not set correctly, exiting."
+        exit 1;
+    fi
+
+    wget -O $VOLUME_DIR/tini \
+    https://github.com/krallin/tini/releases/download/v0.18.0/tini-static
+    chmod +x $VOLUME_DIR/tini
```
geekodour, line 727: we can move the tini download part to the prometheus-builder image
then run prometheus-builder for both deployments

- the release deployment should only copy tini to the required directory
- the pr deployment will work same + copy tini to the required directory

this way we won't need to run this script from k8s and reduce the number of
initContainers by 1 for pr deployment

krasi-georgiev:
> Still have to add restart on the tsdb metrics, otherwise this is working fine.

Remind me why do we need these metrics?

> Alternative way to do the same will use terminationGracePeriodSeconds:0 and then applying a new deployment. Let me know if that'll be more preferable than this.

Do you mean to do something like - delete deployment without waiting for a grace shutdown and recreate the deployment?
if this send a `kill -9` signal than yes It looks like the easier approach.

> also, should I squash the commits to one?

no need I will do it when merging.

geekodour:
though this approach gives much more granularity using the approach in #187 is simpler once `preStop` hook is functional. can this be closed then?

krasi-georgiev:
can you sum up the main difference, as I haven't looked at the code yet.

geekodour:
#180 (this) uses WebSockets to exec commands inside the containers. In this case, the restarter program has complete control over when the `SIGKILL` is sent and when Prometheus is restarted. 

#187 uses the fact that Deployments will restore the pod to its desired state. So the restarter program only updates the Deployment spec and uses  `strategy.type: Recreate` and [`preStop`](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#container-hooks) hook to send `SIGKILL` to existing Pod when the deployment is applied and once the container is killed. Kubernetes brings it back up so we don't have to worry about restarting the prometheus here. 

None of #180 or #187 change `terminationGracePeriodSeconds` so it remains the same in the manifest files.

Both of these use [tini init](https://github.com/krallin/tini/) so that we can run Prometheus as non PID1 process so that we can send the `SIGKILL`

geekodour:
Closing for the same reason as #187

Issue #179, opened 2019-03-25
Improving CLI Logging


geekodour:
I think there is a possibility to improve the CLI logging for prombench.

Current logging:
```
2019/03/25 08:05:45 /home/geekodour/go/src/github.com/prometheus/prombench/pkg/provider/provider.go:33: Request for 'applying deployment:prometheus-test-master' is in progress. Checking in 10s
2019/03/25 08:05:55 /home/geekodour/go/src/github.com/prometheus/prombench/pkg/provider/provider.go:33: Request for 'applying deployment:prometheus-test-master' is in progress. Checking in 10s
2019/03/25 08:06:05 /home/geekodour/go/src/github.com/prometheus/prombench/pkg/provider/provider.go:33: Request for 'applying deployment:prometheus-test-master' is in progress. Checking in 10s
2019/03/25 08:06:16 /home/geekodour/go/src/github.com/prometheus/prombench/pkg/provider/provider.go:33: Request for 'applying deployment:prometheus-test-master' is in progress. Checking in 10s
2019/03/25 08:06:26 /home/geekodour/go/src/github.com/prometheus/prombench/pkg/provider/provider.go:33: Request for 'applying deployment:prometheus-test-master' is in progress. Checking in 10s
2019/03/25 08:06:36 /home/geekodour/go/src/github.com/prometheus/prombench/pkg/provider/provider.go:33: Request for 'applying deployment:prometheus-test-master' is in progress. Checking in 10s
2019/03/25 08:06:47 /home/geekodour/go/src/github.com/prometheus/prombench/pkg/provider/provider.go:33: Request for 'applying deployment:prometheus-test-master' is in progress. Checking in 10s
2019/03/25 08:06:57 /home/geekodour/go/src/github.com/prometheus/prombench/pkg/provider/provider.go:33: Request for 'applying deployment:prometheus-test-master' is in progress. Checking in 10s
```

krasi-georgiev:
don't think we need an issue for that, but PR to improve it is welcome.
Thanks.

Issue #178, opened 2019-03-24
Usage of python scalar in loadgen unclear


geekodour:
So I was looking at [6_loadgen.yaml](https://github.com/prometheus/prombench/blob/master/components/prombench/manifests/benchmark/6_loadgen.yaml) and I am unclear about the usage of the python scalar.

There's a [python scalar](https://github.com/prometheus/prombench/blob/master/cmd/load-generator/main.py#L25) and there's the [go scalar](https://github.com/prometheus/prombench/tree/master/cmd/scaler). The python scalar is not used in `6_loadgen.yaml` but the python querier is used. So the python scalar is redundant and should be removed?

Secondly I think there's no Makefile for pushing [`cmd/load-generator/Dockerfile`](https://github.com/prometheus/prombench/blob/master/cmd/load-generator/Dockerfile) , the Docker file for: https://hub.docker.com/r/prombench/load-generator .

cc @krasi-georgiev  @sipian

sipian:
The python scaler was replaced with the client scaler in https://github.com/prometheus/prombench/pull/139.

At that time it was discussed to leave python scaler as it is. Don't remember the reason though 😛.
I think we can remove it now as it causes unnecessary confusion as stated above.

For the loadgen Makefile , since we didn't have a CI to push docker images, I guess we had left it like that.

We'll have a CI soon after
https://github.com/prometheus/prombench/pull/177  is merged.

geekodour:
great! I'll add the changes then.

Issue #125, opened 2018-08-09
lets use the promtool to replace the python querier.


krasi-georgiev:
lets build a docker image that uses the promtool to send queries to the tested Prometheus servers

krasi-georgiev:
[This is the Python querier](https://github.com/prometheus/prombench/tree/master/cmd/load-generator) that we want to replace with the [Promtool](https://github.com/prometheus/prometheus/tree/master/cmd/promtool)

and [here is the manifest](https://github.com/prometheus/prombench/blob/master/components/prombench/manifests/benchmark/6_loadgen.yaml) how it is deployed

gouthamve:
Hmm, just curious. The Python querier also exposes metrics about the query
latencies. How do you propose to do that using just promtool?

Further if we're going to build a wrapper around promtool for it, we're
better off with the Python querier no?

Thanks
Goutham

geekodour:
metrics in the python querier:
```
1 - loadgen_query_duration_seconds
2 - loadgen_queries_total
3 - loadgen_failed_queries_total
```
I am not very happy with what I thought of but was thinking about starting with replacing L116 with async subprocess call to promtool. 

- `loadgen_queries_total` it's just a increment every time promtool is called
-  `loadgen_failed_queries_total` can be set using return codes, if we need more granularity on return codes we can change how promtool behaves.
- `loadgen_query_duration_seconds` just calculate this once the async process finishes.

https://github.com/prometheus/prombench/blob/93113c77da24f189ff58139e4df01f1cfebd4b9e/cmd/load-generator/main.py#L116-L129

If we don't want to use subprocess, I assume we can write a go implementation and import the promtool package but that will be using the HTTP API anyway.

What do you suggest?

cc @gouthamve

geekodour:
Or we could extract `querier` out as a separate tool like scalar and also expose the /metrics endpoint from there.

Where are the `loadgen_*` metrics used? grepping the metrics just shows the declarations and the grafana dashboard provisioning does not seem to be using them either.

krasi-georgiev:
I looked at the promtool code quickly and @gouthamve 	 is right using the promtool wouldn't make any sense here. 
It seems that re-implementing the python querier in golang is the way to go.

The main benefit for this is that most Prom devs are already familiar with golang so can contribute easily.

geekodour:
Yes, that'd be a better approach. I'll follow that.

PR #182, opened 2019-03-30
Removed Python Scaler and Improved Logging

geekodour:
See #178 #179 

- tested this with the new docker image on gke and is working as expected.
- changed logging to just log the filename and time instead of the full path and the datetime.

related #125

### DIFF
--- a/cmd/load-generator/main.py
+++ b/cmd/load-generator/main.py
@@ -108,27 +108,26 @@ def duration_seconds(s):
     raise "unknown duration %s" % s
 
 def main():
-    if len(sys.argv) < 4 or sys.argv[1] not in ["scaler", "querier"]:
+    if len(sys.argv) < 3:
         print("unexpected arguments")
-        print("usage: <load_generator> <scaler|querier> <namespace> <pr_number>")
+        print("usage: <load_generator> <namespace> <pr_number>")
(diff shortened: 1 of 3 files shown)


cmd/load-generator/main.py
```diff
@@ -158,25 +117,10 @@ def main():
...
-
     config = yaml.load(open("/etc/loadgen/config.yaml", 'r').read())
 
     print("loaded configuration")
 
-    if sys.argv[1] == "scaler":
-        Scaler(config["scaler"], url, req_kwargs).run()
-        return
-
     if sys.argv[1] == "querier":
```
sipian: You can remove this check. We need not check for presence of `"scalar"` in sys.argv.
The "querier" argument can be removed from the config file since only that is there now. https://github.com/prometheus/prombench/blob/02b50cc659cca89db4a1392ddd510c5fdf1bd674/components/prombench/manifests/benchmark/6_loadgen.yaml#L114

sipian:
lgtm
I have deployed and tested the PR. All working well and good :smile:
I have also pushed the docker images to `docker.io/prombench/prombench:2.0.0` & `docker.io/prombench/load-generator:2.0.0`.
Thanks @geekodour

Issue #183, opened 2019-04-04
Running prometheus-builder as an init container


geekodour:
I have a suggestion related to `prometheus-builder`, 

`prometheus-builder` would be more fit for an [init Container](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/).

That way prometheus-builder would work as follows:
- build the PR version
- copy the binary and supporting files to a mounted volume (pref. [emptyDir](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir))
- end of prometheus-builder init container

Then in the actual container for the pod, we can use the base image that the release is using and mount the volume where we copied over the binary and supporting files and execute the prometheus binary from there using kubernetes `commands` and `args` much like what happens with the release image.

as mentioned in #173

krasi-georgiev:
yep that sound like a good idea, just go for it.

Don't think we need to have an issue for that. just open the PR and will discuss there.

PR #146, opened 2018-09-12
[WIP]Use golang docker to build prombench

sipian:
Closes #145 
Signed-off-by: Harsh Agarwal <cs15btech11019@iith.ac.in>

PR #184, opened 2019-04-05
Made prometheus-builder run as an initContainer

geekodour:
See #183 

If this gets merged, work on #180 will become easier as the same commands will work on both pods for restarting, as the prometheus instance in the `pr` version will no longer be running from `build.sh` but directly like the `release` version.

Signed-off-by: geekodour <hrishikeshbman@gmail.com>


components/prombench/manifests/benchmark/3_prometheus-test.yaml
```diff
@@ -733,26 +733,39 @@ spec:
...
+          mountPath: /usrbin
       containers:
       - name: prometheus
-        image: docker.io/prombench/prometheus-builder:2.0.0
+        image: docker.io/prom/busybox:latest
         imagePullPolicy: Always
-        # The prometheus-builder takes a while to build and run
-        # so make sure we start it before the release deployment.
-        # Mark it ready only when prometheus is started.
-        # This way we have the least time difference in the scraped metrics.
```
krasi-georgiev: This comment is good to explain why we need the readiness probe. Can change the wording if needed, but we need some explanation on why do we need a `readinessProbe`

components/prombench/manifests/benchmark/3_prometheus-test.yaml
```diff
@@ -733,26 +733,39 @@ spec:
                 - prometheus
       securityContext:
         runAsUser: 0
+      initContainers:
+      - name: prometheus-builder
+        image: docker.io/geekodour/prometheus-builder:2.2.0
```
krasi-georgiev: @sipian could you push a new image with the changes from this PR to the prombench docker repo.
If the changes are only with some refactoring the new tag could be `docker.io/prombench/prometheus-builder:2.0.1`

Than this can be changed to

```suggestion
        image: docker.io/prombench/prometheus-builder:2.0.1
```

At some point to make test changes easier maybe can even give access to the docker repo to @geekodour if he promises to not break anything :smile: 
The important thing would be to push new images under new tags and NOT overwrite anything that is currently in use in production.

components/prombench/manifests/benchmark/3_prometheus-test.yaml
```diff
@@ -733,26 +733,39 @@ spec:
...
+        args: ["{{ .PR_NUMBER }}"]
+        volumeMounts:
+        - name: usrbin-volume
```
krasi-georgiev: How about 'prometheus-builder' to make it more descriptive?

components/prombench/manifests/benchmark/3_prometheus-test.yaml
```diff
@@ -733,26 +733,39 @@ spec:
...
+          mountPath: /usrbin
```
krasi-georgiev: than this can be `/usr/bin/prometheus-builder`

krasi-georgiev (commented):
Very good PR. With some small changes this is a good improvement.

geekodour:
~I'll add the changes today.~
 added the changes

cc @krasi-georgiev

components/prombench/manifests/benchmark/3_prometheus-test.yaml
krasi-georgiev: lets leave it unresolved until @sipian confirms new image has been pushed and tested.

temp/prometheus-builder/build.sh
```diff
@@ -2,30 +2,38 @@
 
 PR_NUMBER=$1
 FLAGS="${@:2}"
-
-if [ -z "$PR_NUMBER" ]; then echo "ERROR::PR NUMBER is missing in argument" && exit 1; fi
-
 DIR="/go/src/github.com/prometheus/prometheus"
+VOLUME_DIR="/prometheus-builder" # from 3_prometheus-test.yaml
```
krasi-georgiev: can't we pass this as an env variable so it is all in `3_prometheus-test.yaml` this we we need to remember to change this only in a single file.

temp/prometheus-builder/build.sh
```diff
@@ -2,30 +2,38 @@
...
-             --log.level=debug
+echo ">> Creating prometheus binaries"
+if ! make build; then
+    echo "ERROR:: Building of binaries failed"
+    exit 1;
+fi
+
+echo ">> Copy files to volume"
+cp prometheus               $VOLUME_DIR/prometheus
+cp promtool                 $VOLUME_DIR/promtool
```
krasi-georgiev: why do we need the promtool?

krasi-georgiev:
maybe next can continue the work from https://github.com/prometheus/prombench/pull/146

temp/prometheus-builder/build.sh
geekodour: yes! we can do this. I'll add it. this will also avoid rebuilding the image if we change the `mountPath`

temp/prometheus-builder/build.sh
geekodour: we don't need, I copied over the docker `COPY` commands from the official prometheus docker image. I'll remove the promtool copy as we're not making use of it.

components/prombench/manifests/benchmark/3_prometheus-test.yaml
```diff
@@ -736,7 +736,10 @@ spec:
       initContainers:
       - name: prometheus-builder
         image: docker.io/prombench/prometheus-builder:2.0.1
-        args: ["{{ .PR_NUMBER }}"]
+        args: [
+          "{{ .PR_NUMBER }}",
+          "/prometheus-builder", # should be same as mountPath
```
krasi-georgiev: I would prefer named variables so that changing the order won't make it break.

PR_NUMBER="{{ .PR_NUMBER }}"
VOLUME_DIR="/prometheus-builder", # should be same as mountPath

krasi-georgiev:
@geekodour can you also bump the golang version of the builder to 1.12 as part of this PR in
https://github.com/prometheus/prombench/blob/77fe5618c07e693f3ed5922252cbd50570d67b62/temp/prometheus-builder/Dockerfile#L1

components/prombench/manifests/benchmark/3_prometheus-test.yaml
```diff
@@ -747,12 +758,22 @@ spec:
...
           periodSeconds: 2
           failureThreshold: 30
-        args: ["{{ .PR_NUMBER }}"]
+        command: ["/usr/bin/prometheus"]
+        args: [
+          "--web.external-url=http://prombench.prometheus.io/{{ .PR_NUMBER }}/prometheus-pr",
+          "--storage.tsdb.path=/prometheus",
+          "--web.console.libraries=/usr/bin/console_libraries",
+          "--web.console.templates=/usr/bin/consoles",
+          "--config.file=/etc/prometheus/prometheus.yml",
```
krasi-georgiev, line 767: btw why do we need all these additional flags? Aren't some of these added by default? 

`--log.level=debug` is good to keep

geekodour: we can get rid of the `--web.console.*` flags as the point to the executable's running directory. I'll add it

Also can we remove the Makefiles inside `/temp/prometheus-builder/` ? I don't see them being used.

krasi-georgiev: > we can get rid of the `--web.console.*` flags as the point to the executable's running directory. I'll add it

Thanks, if we can get rid of anything else please remove it. 




> Also can we remove the Makefiles inside `/temp/prometheus-builder/` ? I don't see them being used.

yes I  noticed this today as well.

krasi-georgiev:
I have already tested and pushed the new images for these changes.

components/prombench/manifests/benchmark/3_prometheus-test.yaml
krasi-georgiev: > we can get rid of the --web.console.* flags as the point to the executable's running directory.

I can see that for the other deployment it is still there. Do you know why are these needed?

geekodour: from [k8s docs](https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/):
> If you supply only args for a Container, the default Entrypoint defined in the Docker image is run with the args that you supplied.

While the release docker image does put most of the flags at the same [places](https://github.com/prometheus/prometheus/blob/master/Dockerfile), it does not have have the `--web.external.url` flag so have to have to specify the other flags that are different from default with it.

the [Dockerfile](https://github.com/prometheus/prometheus/blob/master/Dockerfile) is `COPY`ing console files to `/usr/share/prometheus/console_libraries/` and then softlinking it to `/etc/prometheus` so our args in release deployment object are working and required.

krasi-georgiev: in that case lets try to keep the one we build and the one we run from the release images as identical as possible for easy maintenance.

krasi-georgiev:
LGTM

ping @sipian

krasi-georgiev:
looks like @sipian is busy so mergins as is so you can continue and will open a PR if there are any other changes.

PR #185, opened 2019-04-09
used docker golang to build prombench

### LINKED ISSUE #145: Let's use Docker to build the binaries.
instead or relying that the local dev env has golang with modules support installed let's just use the minimal [golang docker image](https://hub.docker.com/_/golang/)

https://github.com/prometheus/prombench/blob/f13a14db5a1d3bee58999ff789cc58ff593ff22a/Makefile#L31-L33

geekodour:
Cont. #146 
Closes #145 

Signed-off-by: geekodour <hrishikeshbman@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 PROMBENCH_CMD        = ./prombench
 DOCKER_TAG = docker.io/prombench/prombench:2.0.0
-GOLANG_IMG = golang:1.11
+GOLANG_IMG = golang:1.12
 PROMBENCH_DIR = /go/src/github.com/prometheus/prombench
 USERID = $(shell id -u ${USER})
 USERGROUP = $(shell id -g ${USER})
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ GOLANG_IMG = golang:1.11
 PROMBENCH_DIR = /go/src/github.com/prometheus/prombench
 USERID = $(shell id -u ${USER})
 USERGROUP = $(shell id -g ${USER})
-DOCKER_CMD = docker run --rm \
+DOCKER_CMD = docker run --rm\
 			  -e GOPATH='/go' \
 			  -e GO111MODULE='on' \
 			  -v ${GOPATH}:/go \
@@ -42,7 +42,7 @@ nodepool_delete:
 
 build:
 	@$(DOCKER_CMD) go build ./cmd/prombench/
-	@sudo chown ${USERID}:${USERGROUP} prombench go.sum
+	@$(DOCKER_CMD) chown ${USERID}:${USERGROUP} ./prombench ./go.sum
 
 docker: build
 	@docker build -t $(DOCKER_TAG) .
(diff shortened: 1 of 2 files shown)


krasi-georgiev:
actually can you check if we can use promu for this?

krasi-georgiev:
hm doesn't look like promu has a docker image.

geekodour:
@krasi-georgiev I am not feeling good about having to use `sudo` to change ownership of the executable but if I run the golang docker image as the running user then it shows something like `cannot use internal cache, required for gomodules` which is reasonable as internal cache only the container root user has access to.

Is having the `sudo` command there fine?

krasi-georgiev:
Can you first check if promu has a docker image and if yes I guess the problem is already solved there. Ask in the IRC -dev channel. I think @SuperQ or @simonpasquier can answer that question.

krasi-georgiev:
after seeing Ben's comment on IRC it sounds it would be easiest to  stick to this simple approach of using the docker golang image directly.


>  I am not feeling good about having to use sudo to change ownership of the executable but if I run the golang docker image as the running user then it shows something like cannot use internal cache, required for gomodules which is reasonable as internal cache only the container root user has access to.

when running docker you can map users/groups from the host so this should avoid the need to use sudo. Sudo is another dependency and I like to minimise dependencies.

Makefile
```diff
@@ -4,7 +4,7 @@ GOLANG_IMG = golang:1.11
 PROMBENCH_DIR = /go/src/github.com/prometheus/prombench
 USERID = $(shell id -u ${USER})
 USERGROUP = $(shell id -g ${USER})
-DOCKER_CMD = docker run --rm \
+DOCKER_CMD = docker run --rm\
```
krasi-georgiev: instead of `@$(DOCKER_CMD) chown ${USERID}:${USERGROUP} `
you should be able to do `@$(DOCKER_CMD) = docker run -u $USERID:$USERGROUP`

geekodour: I tried doing that earlier, but it gives me the following error:
```
go: disabling cache (/.cache/go-build) due to initialization failure: mkdir /.cache: permission denied
go: cannot use modules with build cache disabled
```

krasi-georgiev: ok go it , lets leave as is.

krasi-georgiev: maybe try setting GOCACHE or HOME envs

https://github.com/golang/go/issues/26280

krasi-georgiev: https://golang.org/pkg/cmd/go/internal/help/

krasi-georgiev: you can set GOCACHE to /tmp which is writable by everyone

geekodour: Yes i just did that, it's working now.

krasi-georgiev:
LGTM
ping @sipian

sipian (approved):

Makefile
```diff
@@ -1,39 +1,52 @@
 PROMBENCH_CMD        = ./prombench
 DOCKER_TAG = docker.io/prombench/prombench:2.0.0
+GOLANG_IMG = golang:1.11
```
krasi-georgiev: why not 1.12? I just tested it and it seems to be working fine.

krasi-georgiev:
also I just remembered that it would be good to mention in https://github.com/prometheus/prombench/blob/master/cmd/prombench/README.md that can use `make build` to build it.

krasi-georgiev:
Thanks!

PR #187, opened 2019-04-26
[WIP] Added Restarter - Uses preStop Hook

geekodour:
After setting `terminationGracePeriodSeconds` to `0` in the container `PodSpec` for the deployments, Kubernetes still seems to be sending `SIGTERM`

So used a [preStop hook](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#container-hooks) instead. ~But the preStop hook is not getting triggered for some reason, investigating it now.~

As mentioned previously in #180 triggering restarts based on tsdb metrics(whenever a compaction is taking place) is still due.

Also added Job resource, didn't add the `JobDelete` method yet. Now both of the prometheus instances start ~10s apart(when starting) and almost at the same time in subsequent restarts.

Having the pod build prometheus on each pod deployment takes a lot of time and both the instances are not restarted at the same time, Jobs combined with persistent volumes fixes it.

ping @krasi-georgiev

### DIFF
--- a/cmd/restarter/Dockerfile
+++ b/cmd/restarter/Dockerfile
@@ -1,0 +1,5 @@
+FROM alpine:3.8
+
+COPY restarter /restarter
+
+ENTRYPOINT ["/restarter"]
--- a/cmd/restarter/Makefile
+++ b/cmd/restarter/Makefile
@@ -1,7 +1,7 @@
-DOCKER_TAG = docker.io/geekodour/restarter:1.0.0
+DOCKER_TAG = docker.io/geekodour/restarter:graceperiod
 
 build:
-	@go version | grep go1.11 || exit  "Requires golang 1.11 with support for modules!"
+	@go version | grep go1.12 || exit  "Requires golang 1.12 with support for modules!"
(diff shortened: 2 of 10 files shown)


krasi-georgiev:
> After setting `terminationGracePeriodSeconds` to `0` in the container `PodSpec` for the deployments, Kubernetes still seems to be sending `SIGTERM`
> 
I found this: "In kubectl version >= 1.5, you must specify an additional flag --force along with --grace-period=0 in order to perform force deletions."
https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods



> As mentioned previously in #180 triggering restarts based on tsdb metrics(whenever a compaction is taking place) is still due.
> 

don't worry about this, we can add this in a separate PR to keep this one as small as possible.

krasi-georgiev:
did you try that suggestion - --force along with --grace-period=0 

there must be some flag in the k8s client.

geekodour:
closing this as no longer working on this, will open a new PR addressing this problem; new kubernetes features should make this easier to implement since we upgraded versions since the time this was worked on.

PR #188, opened 2019-05-13
added CRD apply for prombench

geekodour:


pkg/provider/k8s/k8s.go
```diff
@@ -24,10 +24,16 @@ import (
 
 	"github.com/prometheus/prombench/pkg/provider"
 
+	apiserverextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
```
krasi-georgiev: lets use camel case everywhere to make it easyer to read:
`apiserverextensionsv1beta1 -> apiServerExtensionsV1Beta1` or something similar which is easy to read.
The same for the other places.

go.mod
```diff
@@ -1,62 +1,45 @@
...
+	k8s.io/client-go v11.0.0+incompatible
+	k8s.io/klog v0.3.0 // indirect
+	sigs.k8s.io/yaml v1.1.0 // indirect
+)
+
+replace (
+	k8s.io/api => k8s.io/api v0.0.0-20190222213804-5cb15d344471
+	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.0.0-20190221221350-bfb440be4b87
+	k8s.io/apimachinery => k8s.io/apimachinery v0.0.0-20190221213512-86fb29eff628
+	k8s.io/client-go => k8s.io/client-go v10.0.0+incompatible
```
krasi-georgiev: Why are these replaces needed?

pkg/provider/k8s/k8s.go
```diff
@@ -416,7 +433,49 @@ func (c *K8s) deploymentApply(resource runtime.Object) error {
 		func() (bool, error) { return c.deploymentReady(resource) })
 }
 
+func (c *K8s) customResourceDefinitionApply(resource runtime.Object) error {
```
krasi-georgiev:
```suggestion
func (c *K8s) customResourceApply(resource runtime.Object) error {
```

Lets keep the names as short as possible. the same for Delete - `customResourceDelete`

pkg/provider/k8s/k8s.go
```diff
@@ -416,7 +433,49 @@ func (c *K8s) deploymentApply(resource runtime.Object) error {
...
+		log.Printf("resource created - kind: %v, name: %v", kind, req.Name)
+	default:
+		return fmt.Errorf("unknown object version: %v kind:'%v', name:'%v'", v, kind, req.Name)
+	}
+
+	return nil
+}
+
 func (c *K8s) ingressApply(resource runtime.Object) error {
+	// TODO(geekodour): maybe use Ingress from networking.k8s.io/v1beta1
```
krasi-georgiev: Instead if this comment wouldn't be better to open a new issue and adding more details?

krasi-georgiev (commented):
Very good PR, thanks!

I assume you have tested it?


Do you think we can also find some way to remove the need for kybectl completely?

```
kubectl get ingress ing -o go-template='{{ range .status.loadBalancer.ingress}}{{.ip}}{{ end }}'

export INGRESS_IP=$(kubectl get ingress ing -o go-template='{{ range .status.loadBalancer.ingress}}{{.ip}}{{ end }}')
```
This is the only reason why we still need it I think.

pkg/provider/k8s/k8s.go
geekodour: I looked into a bit more, this will be required when we upgrade k8s version to 1.14.x so we can leave it as is now. I'll remove the comment.

geekodour:
yes I've tested it, about removing `kubectl` I'll try to implement the approach i mentioned in #77 in a seperate PR.

components/prow/manifests/nginx-controller.yaml
```diff
@@ -37,13 +37,6 @@ spec:
         ports:
         - name: backend-port
           containerPort: 8080
-        resources:
```
krasi-georgiev, line 40: any idea why was this here?

pkg/provider/k8s/k8s.go
```diff
@@ -36,7 +42,8 @@ type Resource struct {
 
 // K8s holds the fields used to generate API reqeust from within a cluster.
 type K8s struct {
-	clt *kubernetes.Clientset
+	clt       *kubernetes.Clientset
+	apiextclt *apiServerExtensionsClient.Clientset
```
krasi-georgiev: Please use camel case to make it easyer to read.
```suggestion
	ApiExtClient *apiServerExtensionsClient.Clientset
```

components/prow/manifests/nginx-controller.yaml
geekodour: this is from the official helm charts for `nginx-ingress` 
https://github.com/helm/charts/blob/a5f93f44c34fdb188919ac25b41423e9af718661/stable/nginx-ingress/values.yaml#L352-L358

but having those resource limitations was causing the pod to fail, I am not sure how the current `nginx-ingress` is running but It was failing for me every time the resource limitations were set.

I couldn't find why they were added there in the first place.

krasi-georgiev: ok , thanks for the extra details. I don't think we need these limitation as well

geekodour:
about changing to `CLIENT_EMAIL` :
[1] https://cloud.google.com/kubernetes-engine/docs/how-to/role-based-access-control#iam-rolebinding-bootstrap

additional logs:
`master`:
http://dpaste.com/1RTCJVJ.txt (passes, clientid)
http://dpaste.com/2TAN2PS.txt (fails, clientemail)
`this branch/pr`:
http://dpaste.com/2X6CQCN.txt (fails on clientid and passes on clientemail)


@sipian 's comment : https://github.com/coreos/prometheus-operator/issues/357#issuecomment-397078995

krasi-georgiev:
Thanks for the clarification. 

From the link it seems that they simplified some role restrictions so if we can also simplify some role/bindings feel free to do it.
Do whichever is the simplest solution and still covers some common security level.

krasi-georgiev:
let me know when you have tested it and it is ready for merge.

geekodour:
I've tested this, I pasted the test logs in my previous comment. I think we can merge this but we also need to update or add a new tag for the docker images I think as this PR also updates the dependencies.

PR #191, opened 2019-05-21
[RDY] Uses custom domain instead of ingress for grafana and prometheus

geekodour:
See #77 

- This should allow the domain to be set using the deployment variable `DOMAIN_NAME`
- This also eliminates the use of `INGRESS_IP` in the manifest files and in prombench commands

I've tested this once on whether all the manifest files get applied accordingly. however, I didn't check if there is any change in grafanas behavior. I'll test it soon.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ nodepool_create:
 resource_apply:
 	$(PROMBENCH_CMD) gke resource apply -a ${AUTH_FILE} \
 		-v ZONE:${ZONE} -v PROJECT_ID:${PROJECT_ID} -v CLUSTER_NAME:${CLUSTER_NAME} \
-		-v PR_NUMBER:${PR_NUMBER} -v RELEASE:${RELEASE} -v DOMAIN_NAME:prombench.prometheus.io \
+		-v PR_NUMBER:${PR_NUMBER} -v RELEASE:${RELEASE} -v DOMAIN_NAME:${DOMAIN_NAME} \
 		-f components/prombench/manifests/benchmark
 
 resource_delete:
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -28,7 +28,7 @@ nodepool_create:
 resource_apply:
 	$(PROMBENCH_CMD) gke resource apply -a ${AUTH_FILE} \
 		-v ZONE:${ZONE} -v PROJECT_ID:${PROJECT_ID} -v CLUSTER_NAME:${CLUSTER_NAME} \
-		-v PR_NUMBER:${PR_NUMBER} -v RELEASE:${RELEASE} \
+		-v PR_NUMBER:${PR_NUMBER} -v RELEASE:${RELEASE} -v DOMAIN_NAME:prombench.prometheus.io \
```
krasi-georgiev: should be an env var. Should try to avoid hard codding string as much as possible. 
It will allow to quickly switch between test and prod envs.

README.md
```diff
@@ -42,26 +42,22 @@ export GRAFANA_ADMIN_PASSWORD=password
     -f components/prow/manifests/rbac.yaml -f components/prow/manifests/nginx-controller.yaml
 ```
 
-- Export the nginx-ingress-controller IP address.
+- Export the domain name to an environment variable
```
krasi-georgiev: maybe can move it after 
https://github.com/prometheus/prombench/pull/191/files#diff-04c6e90faac2675aa89e2176d2eec7d8R31

krasi-georgiev:
is this still WIP?

geekodour:
I tested this e2e, working as expected.

geekodour:
This also will need docker image update for prombench, I think we can push it to 2.0.1 itself(the last one)?

krasi-georgiev:
Thanks.
yes feel free to push to 2.0.1

at some point we need to push these new changes to the production Prow server.

PR #194, opened 2019-05-30
gce service account using the gcloud cli

geekodour:
this should allow creation of service account with required permissions without using web GKE UI

I've tested this, seems to work for me. @nikita-mk can you test this if this works for you as well?

cc @krasi-georgiev @nikita-mk 

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,8 +9,35 @@ It is designed to support adding more k8s providers.
 ### Create a k8s cluster
 ---
 - Create a new project on Google Cloud.
-
 - Create a Service Account on GKE with role `Kubernetes Engine Service Agent` & `Kubernetes Engine Admin` and download the json file.
+
+Alternatively you can use the gcloud cli to create a service account:
+```
+export PROJECT_ID=<google-cloud project-id>
+export SERVICE_ACCOUNT_NAME=<gcp-service-account-name>
+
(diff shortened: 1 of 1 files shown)


krasi-georgiev:
Thanks, Although this is useful I would prefer to not include so specific details as there get outdated pretty quickly. Maybe instead just add a link to the actual document where this is explained.

Otherwise I think it is kind of out of scope for the Prombench readme.

nikita-mk:
It worked for me as well

krasi-georgiev:
wouldn't be best to have it as something like:
```
- Create a [Service Account](https://cloud.google.com/iam/docs/creating-managing-service-accounts) on GKE with role `Kubernetes Engine Service Agent` & `Kubernetes Engine Admin`. If using gcloud cli add the [`roles/container.admin`](https://cloud.google.com/kubernetes-engine/docs/how-to/iam#kubernetes-engine-roles) and [`roles/iam.serviceAccountUser`](https://cloud.google.com/kubernetes-engine/docs/how-to/iam#service_account_user) roles to the GCP serviceAccount and download the json file.
```

or something similar summed in a single paragraph.

krasi-georgiev:
ping

krasi-georgiev:
Thanks!

PR #177, opened 2019-03-21
Add go tool tests and CircleCI config to vet

CrazyByDefault:
Closes #152 , continuation of PR #161 
closes #161
closes #100

I've added a simple CircleCI config that runs make vet, and have updated the Makefile.common to the latest one used on prometheus master

Requesting review @krasi-georgiev :)

PR #195, opened 2019-05-31
remove fake-webserver binary

mfrw:
It seems `fake-webserver` was accidentally checked in.


krasi-georgiev:
Thanks

geekodour:
@krasi-georgiev we also lack a `.gitignore` I'll add it, also do you think `make build` should also build other go binaries in [`/cmd`](https://github.com/prometheus/prombench/tree/master/cmd)

ref. https://github.com/prometheus/prombench/pull/177

krasi-georgiev:
@geekodour 	yep good idea.

PR #196, opened 2019-06-02
Added Makefile.common and .gitignore

### LINKED ISSUE #61: Use multistage builds for the docker images.
first stage -  build inside a golang container
second stage - just copy the binary in an alpine container.

https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds

geekodour:
Cont. #177 
Closes #61 

This PR uses promu to build all the go binaries in prombench and uses the docker targets from Makefile.common , also modified Makefile.common to use `DOCKERFILE_PATH`

Todo:
- [x] update readme
- [x] ~add circleCI (should we do this is a different PR?)~

cc @krasi-georgiev  @simonpasquier


krasi-georgiev:
> add circleCI (should we do this is a different PR?)

yes lets do in a separate PR

.gitignore
```diff
@@ -0,0 +1,9 @@
+*#
+/*.yaml
+*.exe
+tags
+
+prombench
```
krasi-georgiev: I would say lets just leave

```
prombench
cmd/fake-webserver/fake-webserver
cmd/scaler/scaler
```

any other ones that are local env specific can be added to `.git/info/exclude`

Makefile.common
```diff
@@ -0,0 +1,252 @@
+# Copyright 2018 The Prometheus Authors
```
simonpasquier, line 1: It doesn't seem to be the same version than the [original Makefile.common](https://github.com/prometheus/prometheus/blob/master/Makefile.common). The rule is that any change to Makefile.common needs to be opened against the github.com/prometheus/prometheus repository. Then we have automation in place that automatically pushes changes to the other repositories when the copies don't match with the source anymore.

geekodour: this change adds `DOCKERFILE_PATH` makefile variable, so if I add the PR to prometheus/prometheus will also have to update the corresponding Makefiles of each repository using Makefile.common.  right?

Also, can we add prombench to the automation where the Makefile.common changes are pushed?

simonpasquier: > this change adds DOCKERFILE_PATH makefile variable, so if I add the PR to prometheus/prometheus will also have to update the corresponding Makefiles of each repository using Makefile.common. right?

Yes this change needs to go first to prometheus/prometheus.

> Also, can we add prombench to the automation where the Makefile.common changes are pushed?

This will be done automatically.

geekodour:
the new `Makefile.common` will generate docker images in this format: `prombench/prombench-linux-amd64:<TAG>`

.promu.yml
```diff
@@ -0,0 +1,15 @@
+go:
+    # Whenever the Go version is updated here, .travis.yml and
+    # .circle/config.yml should also be updated if in use.
+    version: 1.12
+repository:
+    path: github.com/prometheus/prombench
+build:
+    binaries:
+        - name: prombench
```
krasi-georgiev, line 9: I would say to simplify and make it consistent lets build all tools in the root dir
```
		- name: prombench
		  path: ./cmd/prombench
        - name: scaler
          path: ./cmd/scaler
        - name: fake-webserver
          path: ./cmd/fake-webserver
```
or if tha makes things more complicated than put all in their own folders in ./cmd/name/name

krasi-georgiev:
>the new Makefile.common will generate docker images in this format: prombench/prombench-linux-amd64:

hm , that means we would need to change some deployment files than right?

.promu.yml
geekodour: This required some changes to how docker builds are run, I added the changes here as well opened a PR for the original Makefile.common here: https://github.com/prometheus/prometheus/pull/5669

krasi-georgiev (commented):
Is this ready for merge?

need to sign the DCO

geekodour:
I think it will be good to wait till the change is accepted in the original Makefile.common https://github.com/prometheus/prometheus/pull/5669

if we need to make any other change we can add here then.

> hm , that means we would need to change some deployment files than right?

Yes, after the docker images are pushed. I'll try to add the CI next so that we can have automatic image pushing.

krasi-georgiev:
ok ping me when ready for merge.

simonpasquier:
I'm not sure we need to change `Makefile.common` to add yet another variable. In general, the (Circle) CI generates the container images using the binaries from the `build` job which runs the [`promu crossbuild`](https://github.com/prometheus/consul_exporter/blob/36ccaa30de2017669da4c9d7952045ed59780aa6/.circleci/config.yml#L25-L38) command which puts things in the `.build` directory. This setup enables out-of-the-box support for multi-arch images (here is an [example](https://github.com/prometheus/prometheus/blob/91d7175eaac18b00e370965f3a8186cc40bf9f55/Dockerfile#L1-L12)) which could be quite interesting for prombench IMO. Not all repositories have multi-arch images (yet) in which case the CI symlinks the binary from the `.build/linux-amd64` directory to the project's root directory. Related to this, we've got a discussion happening in https://github.com/prometheus/prometheus/pull/5609 to see how we can fix `make build docker` so it works again on local machines.

Back to your PR and taking `fake-webserver` as an example, I'd modify the Dockerfile like this:

```
ARG ARCH="amd64"
ARG OS="linux"
FROM quay.io/prometheus/busybox-${OS}-${ARCH}:latest
ARG ARCH="amd64"
ARG OS="linux"
COPY .build/${OS}-${ARCH}/fake-webserver        /bin/fake-webserver
...
```

Then running this in the CI should work:
```
make docker DOCKERFILE_PATH=./cmd/fake-webserver
```

simonpasquier:
Regarding

> the new Makefile.common will generate docker images in this format: prombench/prombench-linux-amd64:<TAG>

It should be supported correctly by the Docker engine: if you say `docker run prombench/prombench ...`, it will pick up the correct image based on the machine's architecture/OS.

geekodour:
@simonpasquier thanks a lot for the info! I'll test this today. Also, I am not sure if `make docker DOCKERFILE_PATH=./cmd/fake-webserver` will work because `.build` directory will not be found as it will be in the root directory and the docker build context is inside the `./cmd/fake-webserver` directory.

Adding the changes in https://github.com/prometheus/prometheus/pull/5669 should make this work. I'll test this and get back here.

simonpasquier:
@geekodour you're right, I'll +1 your Makefile.common's PR.

geekodour:
Thanks @simonpasquier https://github.com/prometheus/prometheus/pull/5669 got merged, I added the new `Makefile.common` to this PR. 

@krasi-georgiev if everything's okay with the changes, I think this can be merged.

krasi-georgiev:
Nice work , Thanks!

geekodour:
EDIT2: it worked, it was a network issue.

--- 
@simonpasquier @krasi-georgiev @SuperQ  I was working on the dockerimage for new tool for prombench and was trying to get the manifests working, but I am facing the following issue:

**runs successfully**
```shell
make docker DOCKERFILE_PATH=tools/amgh/Dockerfile DOCKERBUILD_CONTEXT=./ DOCKER_IMAGE_NAME=amgh DOCKER_IMAGE_TAG=0.0.1
```
**runs successfully**
```shell
make docker-publish DOCKERFILE_PATH=tools/amgh/Dockerfile DOCKERBUILD_CONTEXT=./ DOCKER_IMAGE_NAME=amgh DOCKER_IMAGE_TAG=0.0.1 
```
**fails**
```shell
make docker-manifest DOCKERFILE_PATH=tools/amgh/Dockerfile DOCKERBUILD_CONTEXT=./ DOCKER_IMAGE_NAME=amgh DOCKER_IMAGE_TAG=0.0.1 
```
**fail output**
```shell
DOCKER_CLI_EXPERIMENTAL=enabled docker manifest create -a "prombenchgeekodour/amgh:0.0.1" prombenchgeekodour/amgh-linux-amd64:0.0.1
no such manifest: docker.io/prombenchgeekodour/amgh-linux-amd64:0.0.1
make: *** [Makefile.common:232: common-docker-manifest] Error 1
```
the test image is published here: https://hub.docker.com/r/prombenchgeekodour/amgh-linux-amd64/tags

the Dockerfile:
```Dockerfile
ARG ARCH="amd64"
ARG OS="linux"
FROM quay.io/prometheus/busybox-${OS}-${ARCH}:glibc
LABEL maintainer="The Prometheus Authors <prometheus-developers@googlegroups.com>"

COPY amgh /bin/amgh

CMD /bin/amgh

```

what am I missing?

EDIT: same with the hello-world image. seems more like a docker cli issue?
```
λ docker manifest inspect --verbose hello-world
no such manifest: docker.io/library/hello-world:latest
```

PR #200, opened 2019-06-13
[WIP] Simpleargs cli utility for github actions

geekodour:
This enables comment validation and matching for github comments.

Please see `README.md` for more info.

Currently I just used the binary in the github actions demo, we can create a docker image for this if required.

If tool is accepted I think it should go inside `/third_party` according to this https://github.com/golang-standards/project-layout

ps: very bad at naming tools, please suggest a better name :smile:

### DIFF
--- a/cmd/simpleargs/main.go
+++ b/cmd/simpleargs/main.go
@@ -18,7 +18,7 @@ var writepath string
 func writeArgs(groups []string) {
 	for i, group := range groups[1:] {
 		data := []byte(group)
-		filename := fmt.Sprintf("ARG%d", i)
+		filename := fmt.Sprintf("ARG_%d", i)
 		err := ioutil.WriteFile(filepath.Join(writepath, filename), data, 0644)
 		if err != nil {
 			log.Fatalln(err)
--- a/cmd/simpleargs/main.go
+++ b/cmd/simpleargs/main.go
@@ -1,0 +1,61 @@
+package main
+
+import (
+	"fmt"
+	"github.com/google/go-github/v26/github"
+	"gopkg.in/alecthomas/kingpin.v2"
+	"io/ioutil"
+	"log"
+	"os"
+	"path/filepath"
+	"regexp"
+)
+
+var regex string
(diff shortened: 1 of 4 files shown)


geekodour:
cc @nikita-mk @krasi-georgiev

cmd/simpleargs/main.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+var regex string
+var eventfile string
+var writepath string
+
+func writeArgs(groups []string) {
+	for i, group := range groups[1:] {
+		data := []byte(group)
+		filename := fmt.Sprintf("ARG%d", i)
```
krasi-georgiev: maybe to make it a bit more readable
```suggestion
		filename := fmt.Sprintf("ARG_%d", i)
```

krasi-georgiev:
Thanks @geekodour 	that was really fast and the implementation looks quite nice. I will wait for @nikita-mk 	to test is and suggest changes and will do a final review once she confirms that it works as expected.

> If tool is accepted I think it should go inside /third_party according to this golang-standards/project-layout

I kind of prefer the `/tools` dir and maybe move the other ones from cmd there as well 
https://github.com/golang-standards/project-layout#tools
Supporting tools for this project. 


> ps: very bad at naming tools, please suggest a better name smile

how about : text2args 
It currently supports only github comments, but at some point if needed can extend it to support other input formats so it can become a more generic tool to parse some input text and output the results into the args file.

`text2args --input=./event.json --input-format=github-comment --output=./ --regex="^myregex$"`

> Currently I just used the binary in the github actions demo, we can create a docker image for this if required.

at some point we would need to add an automated job to push all these tools to the prombench repo in docker hub.

geekodour:
@krasi-georgiev okay I'll add the changes.

@nikita-mk do you think we still need to add json output support for this?

nikita-mk:
No, we don't

krasi-georgiev:
superseded by https://github.com/prometheus/prombench/pull/207

PR #207, opened 2019-06-24
Golang cli tool for github actions

nikita-mk:
See #200 
Combines decoding service account auth file, parsing github comments, saving version, pr number to files and author authentication.
cc @krasi-georgiev @geekodour 


cmd/simpleargs/main.go
```diff
@@ -0,0 +1,141 @@
...
+		} else {
+			log.Printf("author is member or collaborator")
+
+			owner := *e.GetRepo().Owner.Login
+			repo := *e.GetRepo().Name
+			number := *e.GetIssue().Number
+			log.Printf(owner)
+			log.Printf(repo)
+			log.Printf("%d", number)
+			writeArg(strconv.Itoa(number), "pr") //writing PR number to file
```
geekodour: instead of saving the pr number which will be available in the `event.json` file anyway. think we can use `jq` directly when adding the arg.

cmd/simpleargs/main.go
```diff
@@ -0,0 +1,141 @@
...
+
+			argRe := regexp.MustCompile(regex)
```
geekodour: I think `MustCompile` uses panic and this will not error out with `78` for a invalid regex. we'll be using proper tested regexes anyway so think we still can keep this the same for sometime if we don't face issues.

cmd/simpleargs/main.go
```diff
@@ -0,0 +1,141 @@
...
+	//Parsing event.json
+	event, err := github.ParseWebHook("issue_comment", data)
+	if err != nil {
+		log.Fatalln("could not parse = %v\n", err)
+	}
+	log.Printf("Parsing event.json")
+
+	//Checking author association and saving args to file
+	switch e := event.(type) {
+	case *github.IssueCommentEvent:
```
geekodour: I think keeping the switch case in the `main` and moving other parts into their own functions and calling then inside the `case` block will make this cleaner.

also if we use a generic comment function we'll be able to use it in the following functions whenever needed.

* member validation function
* regex validation and writing args to disk
* commenting that all validations have passed, starting benchmark.

eg.
```go
	switch e := event.(type) {
	case *github.IssueCommentEvent:
        // call member validation, comment not allowed to run if unauth. (benchmark.go has this feature)
        // call regex validation and writing args to disk
        // call the generic commenting function to signal benchmark is starting
	default:
		log.Fatalln("simpleargs only supports issue_comment event")
	}

```

geekodour (commented):
@krasi-georgiev @nikita-mk added some review comments.

cmd/simpleargs/main.go
```diff
@@ -0,0 +1,141 @@
...
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strconv"
+
+	"github.com/google/go-github/github"
```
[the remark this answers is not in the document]
nikita-mk: I will change that

cmd/simpleargs/main.go
```diff
@@ -0,0 +1,141 @@
...
+	app.Flag("eventfile", "path to event.json").Default("/github/workflow/event.json").StringVar(&eventfile)
+	app.Flag("writepath", "path to write args to").Default("/github/home").StringVar(&writepath)
+	app.Arg("regex", "Regex pattern to match").Required().StringVar(&regex)
+	kingpin.MustParse(app.Parse(os.Args[1:]))
+	log.Printf("After flag declaration")
+	var err error
+
+	//Github client for posting comments
+	token := os.Getenv("GITHUB_TOKEN")
+	http.DefaultClient.Transport = roundTripper{token}
```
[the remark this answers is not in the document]
nikita-mk: That too worked. It was taking more time to build than roundTripper, so I used this.

cmd/simpleargs/main.go
```diff
@@ -0,0 +1,141 @@
...
+	client := github.NewClient(http.DefaultClient)
```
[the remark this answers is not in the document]
nikita-mk: Sure

cmd/simpleargs/main.go
```diff
@@ -0,0 +1,141 @@
...
+
+	//Decoding and saving service account authentication file
```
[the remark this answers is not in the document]
nikita-mk: In Secrets, we store the base64 encoded auth file. I decoded and saved it here instead of having a separate action for doing so.

cmd/simpleargs/main.go
```diff
@@ -0,0 +1,141 @@
...
+	case *github.IssueCommentEvent:
+		if (*e.GetComment().AuthorAssociation != "COLLABORATOR") && (*e.GetComment().AuthorAssociation != "MEMBER") {
+			log.Printf("author is not a member or collaborator")
+			os.Exit(78)
+		} else {
+			log.Printf("author is member or collaborator")
+
+			owner := *e.GetRepo().Owner.Login
+			repo := *e.GetRepo().Name
+			number := *e.GetIssue().Number
```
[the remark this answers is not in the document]
nikita-mk: Sure.

cmd/simpleargs/main.go
```diff
@@ -0,0 +1,141 @@
...
+			log.Printf(owner)
+			log.Printf(repo)
+			log.Printf("%d", number)
```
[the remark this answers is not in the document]
nikita-mk: Will remove those

cmd/simpleargs/main.go
```diff
@@ -0,0 +1,141 @@
...
+		log.Printf("In write args")
+		err := ioutil.WriteFile(filepath.Join(writepath, filename), data, 0644)
+		log.Printf(filepath.Join(writepath, filename))
+		if err != nil {
+			log.Fatalln(err)
+		}
+	}
+}
+
+func writeArg(groups string, filetype string) {
```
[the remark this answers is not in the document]
nikita-mk: Okay

cmd/simpleargs/main.go
nikita-mk: Installing `jq` increases the time.

cmd/simpleargs/main.go
nikita-mk: Okay

cmd/simpleargs/main.go
nikita-mk: Sure

cmd/simpleargs/main.go
geekodour: the action that will actually run `./prombench` will be different from the action that will run this cli. 

I meant using `jq` when passing the args in the second action. I think it won't matter if there's a small delay in the second action as the whole starting of benchmark will take sometime anyway. 

would like @krasi-georgiev 's inputs on this

krasi-georgiev:
@geekodour 	had this idea to also set the 'benchmark' label only so that we can display all PRs which had a running benchmarks like
https://github.com/prometheus/prometheus/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc+label%3Abenchmark

this will allow to filter PRs only with a benchmark and  review some old benchmark results

cmd/simpleargs/main.go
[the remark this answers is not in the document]
krasi-georgiev: I also think that we don't need this?  
I think the prombench tool can read env variables so it will be something like: `/prombench .... --auth=$AUTH_FILE` if that is not the case we can extend it to support this format instead of just a file.

cmd/simpleargs/README.md
```diff
@@ -0,0 +1,41 @@
+# simpleargs
```
krasi-georgiev: now that the tool does a lot more than just parsing vars should name it differently
`commentMonitor` sound like a good name.

also best to put these small tools in `./tools/....'

cmd/simpleargs/go.mod
```diff
@@ -0,0 +1,11 @@
+module /home/niki/github.com/prombench/cmd/simpleargs
```
krasi-georgiev: is there any reason why you haven't updated the folder structure?

cmd/simpleargs/main.go
```diff
@@ -2,130 +2,112 @@ package main
...
 
 var regex string
-var eventfile string
-var writepath string
+var input string
+var output string
+var owner string
+var repo string
+var prnumber int
+var releaseVersion string
```
krasi-georgiev: as I mention global variables is an anti patern and leads to problems at a later stage.

cmd/simpleargs/main.go
```diff
@@ -0,0 +1,141 @@
...
+
+		owner = *e.GetRepo().Owner.Login  //Get parameters
+		repo = *e.GetRepo().Name
+		prnumber = *e.GetIssue().Number
+		releaseVersion = arglist[1]
+		
+		arglist = append(arglist, strconv.Itoa(prnumber))
+		writeArgs(arglist) //Save args to file(releaseVersion in ARG_0, prnumber in ARG_1)
+
+		//Posting benchmark start comment
```
krasi-georgiev: Please read more about golang idioms for comments.
Capital letter and full stop for all comments and run go fmt to format the code correctly.

Best to install and use vscode as it does this for you with every save.

cmd/simpleargs/go.mod
nikita-mk: I will update it.

cmd/simpleargs/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+> `/github` is the where GitHub mounts a shared filesystem to be accessed by multiple actions.
+
+> **Important Note**
+>
+> GitHub Actions uses the HCL language for the workflow files, it has a [known issue related to
+> backslash](https://github.com/hashicorp/terraform/issues/4052). So, `\` needs to be replaced by
+> `\\` when specifying the argument in the HCL file.
+
+Example:
```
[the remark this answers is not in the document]
nikita-mk: Should I delete the readme?

nikita-mk:
> @geekodour had this idea to also set the 'benchmark' label only so that we can display all PRs which had a running benchmarks like
> https://github.com/prometheus/prometheus/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc+label%3Abenchmark
> 
> this will allow to filter PRs only with a benchmark and review some old benchmark results

Should we set the labels in the cli tool or have another action for it?

krasi-georgiev:
only if there is some existing action than can do that, otherwise set it here.

Right at the end.

If there is some existing action we can use you can run it after the running the prombench steps.

cmd/simpleargs/README.md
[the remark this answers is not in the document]
krasi-georgiev: yes, see the prombench cli func main()

cmd/tools/commentMonitor/main.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	switch e := event.(type) {
+	case *github.IssueCommentEvent:
+		// Check author association.
+		if err := memberValidation(*e.GetComment().AuthorAssociation); err != nil {
+			log.Printf("Author is not a member or collaborator")
+			os.Exit(78)
+		}
+		log.Printf("Author is member or collaborator")
```
krasi-georgiev: all logs start with small letter.

cmd/tools/commentMonitor/main.go
```diff
@@ -0,0 +1,144 @@
...
+	token := os.Getenv("GITHUB_TOKEN")
+	ctx := context.Background()
+	ts := oauth2.StaticTokenSource(
+		&oauth2.Token{AccessToken: token},
+	)
+	tc := oauth2.NewClient(ctx, ts)
+	client := github.NewClient(tc)
+
+	// Reading event.json.
+	os.MkdirAll(output, os.ModePerm)
```
krasi-georgiev: IIRC `MkdirAll` return an error which needs to be handler.

Don't forget that all errors needs to be handles or logged.

cmd/tools/commentMonitor/main.go
```diff
@@ -0,0 +1,144 @@
...
+	data, err := ioutil.ReadFile(input)
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	// Parsing event.json.
+	event, err := github.ParseWebHook("issue_comment", data)
+	if err != nil {
+		log.Fatalln("could not parse = %v\n", err)
```
krasi-georgiev: Why `\n` isn't this added automatically?
```suggestion
		log.Fatalln("could not parse = %v", err)
```

cmd/tools/commentMonitor/main.go
```diff
@@ -0,0 +1,144 @@
...
+
+func main() {
+	app := kingpin.New(filepath.Base(os.Args[0]), "simpleargs github comment extract")
+	app.Flag("input", "path to event.json").Default("/github/workflow/event.json").StringVar(&input)
+	app.Flag("output", "path to write args to").Default("/github/home").StringVar(&output)
+	app.Arg("regex", "Regex pattern to match").Required().StringVar(&regex)
+	app.HelpFlag.Short('h')
+	kingpin.MustParse(app.Parse(os.Args[1:]))
+
+	// Github client for posting comments.
```
krasi-georgiev: this can be a separate `clt := newClient(os.Getenv("GITHUB_TOKEN"))` and move as close as possible to where you use the client.

cmd/tools/commentMonitor/main.go
```diff
@@ -0,0 +1,144 @@
...
+	case *github.IssueCommentEvent:
+		// Check author association.
+		if err := memberValidation(*e.GetComment().AuthorAssociation); err != nil {
+			log.Printf("Author is not a member or collaborator")
+			os.Exit(78)
+		}
+		log.Printf("Author is member or collaborator")
+
+		// Validate comment.
+		arglist, err := regexValidation(regex, *e.GetComment().Body)
```
krasi-georgiev:
```suggestion
		args, err := regexValidation(regex, *e.GetComment().Body)
```

or use camelCase

```
		argList, err := regexValidation(regex, *e.GetComment().Body)

```

cmd/simpleargs/README.md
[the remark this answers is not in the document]
krasi-georgiev: The readme still includes test on the flags.

see the prombench cli func main()

cmd/tools/commentMonitor/main.go
```diff
@@ -0,0 +1,143 @@
...
+		prnumber := *e.GetIssue().Number
+
+		// Save args to file. Stores releaseVersion in ARG_0 and prnumber in ARG_1.
+		args = append(args, strconv.Itoa(prnumber))
+		writeArgs(args, *output)
+
+		// Posting benchmark start comment.
+		comment := fmt.Sprintf(`Welcome to Prometheus Benchmarking Tool.
+
+The two prometheus versions that will be compared are _**pr-%d**_ and _**%s**_
```
krasi-georgiev: it is easier and less error prone to make all args as %v

```suggestion
The two prometheus versions that will be compared are _**pr-%v**_ and _**%s**_
```

krasi-georgiev (commented):
Just update the readme and the small nit and will merge.

cmd/simpleargs/README.md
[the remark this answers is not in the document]
nikita-mk: I had deleted the README, but didn't commit it. The help message is similar to the prombench cli

cmd/tools/commentMonitor/main.go
```diff
@@ -0,0 +1,144 @@
...
+		filename := fmt.Sprintf("ARG_%v", i)
+		err := ioutil.WriteFile(filepath.Join(output, filename), data, 0644)
+		if err != nil {
+			log.Fatalln(err)
+		}
+		log.Printf(filepath.Join(output, filename))
+	}
+}
+
+func postComment(client *github.Client, owner string, repo string, prnumber int, comment string) error {
```
krasi-georgiev: Best to move all  these funcs after the main func

cmd/tools/commentMonitor/main.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+func main() {
+	app := kingpin.New(filepath.Base(os.Args[0]), "commentMonitor github comment extract")
+	app.HelpFlag.Short('h')
+	input := app.Flag("input", "path to event.json").Short('i').Default("/github/workflow/event.json").String()
+	output := app.Flag("output", "path to write args to").Short('o').Default("/github/home").String()
+	regex := app.Arg("regex", "Regex pattern to match").Required().String()
+
+	kingpin.MustParse(app.Parse(os.Args[1:]))
```
krasi-georgiev, line 24: I think the prombench cli prints some examples when you type a wrong command or with the -h flag so meed to add the same here

nikita-mk: Prombench cli prints examples for the subcommands and not for the flags. Here, we only have flags. Do we still need to print examples?

krasi-georgiev: Yes i think one example is good to have

krasi-georgiev:
Ci is faling

cmd/tools/commentMonitor/main.go
krasi-georgiev: don't you need something like:

https://github.com/prometheus/prombench/blob/a926df5a16e0eb992bf8f71a74670293e7aa72fb/cmd/prombench/prombench.go#L60-L64

cmd/tools/commentMonitor/main.go
```diff
@@ -0,0 +1,144 @@
...
+	"path/filepath"
+	"regexp"
+	"strconv"
+
+	"github.com/google/go-github/v26/github"
+	"golang.org/x/oauth2"
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+const prombenchURL = "http://prombench.prometheus.io"
```
krasi-georgiev: hm , we need to avoid having this as hard coded.
I mean if the url changes we want to avoid braking things.

Maybe add it as a flag and than will provide it as an env variable or a flag when starting this action. $DOMAIN_NAME

nikita-mk: Added it as an env variable in the main.workflow

krasi-georgiev:
Thanks!

PR #211, opened 2019-06-26
set size retention flag

### LINKED ISSUE #69: Use size based retention for prometheus meta
When prometheus supports size based retention will use it to make sure we don't run out of space

PR https://github.com/prometheus/prometheus/pull/4230

nikita-mk:
Resolves #69 

### DIFF
--- a/manifests/cluster-infra/3b_prometheus-meta.yaml
+++ b/manifests/cluster-infra/3b_prometheus-meta.yaml
@@ -7,7 +7,7 @@ spec:
     - ReadWriteOnce
   resources:
     requests:
-      storage: 1000Gi  # If you change this make sure to update the prometheus meta disk retention settings to 50% of the storage.
+      storage: 1000Gi  # If you change this make sure to update the prometheus meta disk retention settings.
 ---
 apiVersion: v1
 kind: ConfigMap
--- a/manifests/cluster-infra/3b_prometheus-meta.yaml
+++ b/manifests/cluster-infra/3b_prometheus-meta.yaml
@@ -1,5 +1,3 @@
-#Need to set the storage.tsdb.retention.size flag in prometheus-meta deployment.
-#80% of the pvc storage
 apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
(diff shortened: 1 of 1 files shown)


manifests/cluster-infra/3b_prometheus-meta.yaml
```diff
@@ -130,7 +132,7 @@ spec:
         args:
         - "--config.file=/etc/prometheus/config/prometheus.yaml"
         - "--storage.tsdb.path=/data"
-        - "--storage.tsdb.retention=90d"
+        - "--storage.tsdb.retention.size=800GB"  #set to 80% of the storage available
```
krasi-georgiev: put the same comment in the yaml file where we set the current disk size.
something like:
1000GB # If you change this make sure to update the prometheus meta disk retention settings.

also lets make it at 50% as when we add logging with loki it will use quite a bit of space as well.
```suggestion
        - "--storage.tsdb.retention.size=500GB"  # 50% of the total storage available.
```

manifests/cluster-infra/3b_prometheus-meta.yaml
```diff
@@ -1,3 +1,5 @@
+#Need to set the storage.tsdb.retention.size flag in prometheus-meta deployment.
+#80% of the pvc storage
```
krasi-georgiev: don't think we need this extra comment here.

manifests/cluster-infra/3b_prometheus-meta.yaml
krasi-georgiev: No I meant leave this one here as an addition to the one in the other yaml file.

manifests/cluster-infra/3b_prometheus-meta.yaml
```diff
@@ -7,7 +7,7 @@ spec:
     - ReadWriteOnce
   resources:
     requests:
-      storage: 1000Gi
+      storage: 1000Gi  # If you change this make sure to update the prometheus meta disk retention settings to 50% of the storage.
```
krasi-georgiev:
```suggestion
      storage: 1000Gi  # If you change this make sure to update the prometheus meta disk retention settings.
```

krasi-georgiev:
Thanks!

PR #214, opened 2019-06-29
[RDY] Updating Grafana to  6.3.0-beta1

geekodour:
This PR updates Grafana to 6.2.5 and performs the migration steps ( https://grafana.com/docs/installation/docker/#migration-from-a-previous-version-of-the-docker-container-to-5-1-or-later ) ~few things still need to be tested.~

### DIFF
--- a/manifests/cluster-infra/grafana_datasource.yaml
+++ b/manifests/cluster-infra/grafana_datasource.yaml
@@ -16,7 +16,7 @@ data:
       type: prometheus
       access: proxy
       orgId: 1
-      url: http://{{ .DOMAIN_NAME }}/prometheus-meta/
+      url: http://prometheus-meta/prometheus-meta/
       isDefault: true
       jsonData:
          graphiteVersion: "1.1"
--- a/manifests/cluster-infra/grafana_deployment.yaml
+++ b/manifests/cluster-infra/grafana_deployment.yaml
@@ -18,6 +18,8 @@ spec:
         component: core
     spec:
       serviceAccountName: prometheus
+      securityContext:
+        runAsUser: 472
       containers:
       - image: grafana/grafana:6.2.5
         name: grafana-core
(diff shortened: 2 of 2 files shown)


manifests/cluster-infra/grafana_datasource.yaml
```diff
@@ -16,7 +16,7 @@ data:
       type: prometheus
       access: proxy
       orgId: 1
-      url: http://{{ .DOMAIN_NAME }}/prometheus-meta/
+      url: http://prometheus-meta/prometheus-meta/
```
krasi-georgiev, line 19: Why this change? can't we use the env var here as well?

geekodour: this will allow grafana to source prometheus even if the domain is not pointed at it. let me know if I should revert it back to using `.DOMAIN_NAME`

krasi-georgiev: I think this is less likely to brake if we change the name of the service at some point.

krasi-georgiev: or maybe put in some global env variable so that if we change the service name it will update everywhere.

krasi-georgiev:
btw grafana 6.3 should be released next week so might be worth to wait for it.

geekodour:
Yes, It'll be good to wait for it.

manifests/cluster-infra/grafana_datasource.yaml
geekodour: I'll revert this change and add the change you mentioned in another PR.

geekodour: this change is required for changes in #227 since the config files don't take any variables.

https://github.com/prometheus/prombench/blob/7a2c225bff4ef50629867aeeebae419f585c4449/config/grafana/prometheus-datasource.yaml

geekodour:
updated to  6.3.0-beta1 

cc @krasi-georgiev

krasi-georgiev:
Thanks

PR #216, opened 2019-07-01
.circleci/config.yml: use prometheus/prometheus orb

simonpasquier:


krasi-georgiev:
Thanks Simon, much appreciated!

PR #219, opened 2019-07-07
[RDY] Added `NodePoolCheck` to prombench nodepools cli

geekodour:
very much work in progress.

* this cli option will enable users to check if nodepools are running or not
* this will be useful when we need to determine whether to start a test based on state of nodepools

Note: I found that this check is necessary for running prombench tests as Deployments when waiting for the previous test to finish before starting a new one when restarting tests.

cc: @krasi-georgiev 

### DIFF
--- a/pkg/provider/gke/gke.go
+++ b/pkg/provider/gke/gke.go
@@ -426,6 +426,28 @@ func (c *GKE) nodePoolRunning(zone, projectID, clusterID, poolName string) (bool
 	return false, nil
 }
 
+// NodePoolCheck checks if nodepools exists and returns error code if exists
+func (c *GKE) NodePoolCheck(*kingpin.ParseContext) error {
+	reqC := &containerpb.CreateClusterRequest{}
+
+	for _, deployment := range c.gkeResources {
+		if err := yamlGo.UnmarshalStrict(deployment.Content, reqC); err != nil {
+			log.Fatalf("error parsing the cluster deployment file %s:%v", deployment.FileName, err)
+		}
+
+		for _, node := range reqC.Cluster.NodePools {
+			b, err := c.nodePoolRunning(reqC.Zone, reqC.ProjectId, reqC.Cluster.Name, node.Name)
+			if err != nil {
+				log.Fatalln("error fetching nodePool info")
+			}
+			if b == true {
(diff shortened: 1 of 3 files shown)


pkg/provider/gke/gke.go
```diff
@@ -415,6 +415,28 @@ func (c *GKE) nodePoolRunning(zone, projectID, clusterID, poolName string) (bool
 	return false, nil
 }
 
+// NodePoolCheck checks if nodepools exists and returns error code if exists
```
```suggestion
// NodePoolCheck returns an error if the nodepool exists.
```

krasi-georgiev (commented):
Why is this WIP?

Usually when I see WIP it means that it is not ready for a review, but this one seems ok. What else is there to add?

geekodour:
the same changes from this PR are added in #235 

can this be closed? or we can add this and I can remove this change from #235 and it will keep #235 smaller.

cc @krasi-georgiev

pkg/provider/gke/gke.go
```diff
@@ -426,6 +426,33 @@ func (c *GKE) nodePoolRunning(zone, projectID, clusterID, poolName string) (bool
...
+		if err := yamlGo.UnmarshalStrict(deployment.Content, reqC); err != nil {
+			log.Fatalf("error parsing the cluster deployment file %s:%v", deployment.FileName, err)
+		}
+
+		for _, node := range reqC.Cluster.NodePools {
+			isRunning, err := c.nodePoolRunning(reqC.Zone, reqC.ProjectId, reqC.Cluster.Name, node.Name)
+			if err != nil {
+				log.Fatalln("error fetching nodePool info")
+			}
+			exists = exists && isRunning
```
krasi-georgiev, line 445: exists is always true, why is it needed?

geekodour: if any of the nodepool is running then `isRunning` will be `false` and then `exists` will be false and will continue to be false till the return statement below.

I updated the function description, now it's ` // NodePoolCheck returns an error if any the nodepools do not exist. `

geekodour: ^ nodepool is not running

krasi-georgiev:
lets continue in https://github.com/prometheus/prombench/pull/235

PR #222, opened 2019-07-15
read the auth data from a file or env variable

krasi-georgiev:
Github actions provide the secrets as env variables and this PR extends prombench to read the json auth data directly from an env variable.

@nikita-mk 	can you please test me and let me know if we need to change anything else.

### DIFF
--- a/pkg/provider/gke/gke.go
+++ b/pkg/provider/gke/gke.go
@@ -21,6 +21,7 @@ import (
 	"gopkg.in/alecthomas/kingpin.v2"
 	yamlGo "gopkg.in/yaml.v2"
 
+	"google.golang.org/api/option"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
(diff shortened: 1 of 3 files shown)


pkg/provider/gke/gke.go
```diff
@@ -60,12 +64,19 @@ type GKE struct {
...
+		log.Fatal("no auth provided! Need to either set the auth flag or the GOOGLE_APPLICATION_CREDENTIALS env variable")
 	}
-	cl, err := gke.NewClusterManagerClient(context.Background())
+
+	// When the auth variable points to a file
+	// put the file content in the variable.
+	if content, err := ioutil.ReadFile(c.Auth); err == nil {
+		c.Auth = string(content)
+	}
+	opts := option.WithCredentialsJSON([]byte(c.Auth))
```
nikita-mk, line 92:
```suggestion
        if decoded, err := base64.StdEncoding.DecodeString(c.Auth); err == nil {
		c.Auth = string(decoded)
	}
	opts := option.WithCredentialsJSON([]byte(c.Auth))
```

nikita-mk: We need to check if the string is base64 encoded and decode it if necesary.
I tested this locally and its working.

krasi-georgiev: Actually I used the answer from https://stackoverflow.com/questions/8571501/how-to-check-whether-a-string-is-base64-encoded-or-not/20145780

to determine when the string is base64 encoded. This way the error handling is more deterministic.

Please test and suggest changes if needed.

nikita-mk: I have tested it with Github Actions also and its working.

krasi-georgiev:
@geekodour 	would you make a quick review and will merge if you don't see any issues with this change.

PR #226, opened 2019-07-21
Changes needed for Github Actions

nikita-mk:

### DIFF
--- a/cmd/prombench/Dockerfile
+++ b/cmd/prombench/Dockerfile
@@ -13,6 +13,8 @@ RUN \
 
 COPY prombench /prombench/prombench
 COPY Makefile /prombench/Makefile
+COPY Makefile.common /prombench/Makefile.common
 COPY manifests/prombench /prombench/manifests/prombench
 
 WORKDIR /prombench
+ENTRYPOINT [ "make", "-C","/prombench/","deploy"] 
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ DOCKER_REPO             ?= prombench
 
 include Makefile.common
 
-PROMBENCH_CMD        = ./prombench
+PROMBENCH_CMD        = /prombench/prombench
 
 ifeq ($(AUTH_FILE),)
 AUTH_FILE = "/etc/serviceaccount/service-account.json"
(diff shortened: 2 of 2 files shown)


krasi-georgiev:
I bit of details why are these needed would be helpful.

geekodour:
can the `GITHUB_WORKSPACE` env var be used here?

https://developer.github.com/actions/creating-github-actions/accessing-the-runtime-environment/#environment-variables

In the `Makefile` you can add something like:
```
PROMBENCH_PATH             ?= ./
.
.
.
-f ${PROMBENCH_PATH}/manifests/prombench/nodepools.yaml
```

then you can probably set `PROMBENCH_PATH` to `/github/workspace` in the env var. not 100% sure though.

krasi-georgiev:
yep that would be a much better approach, but I still fail to understand why running a docker container in github actions requires changing the paths.

geekodour:
Continuation to the previous message, my assumption was wrong because `GITHUB_WORKSPACE` will be the `prometheus/prometheus` when we need `prometheus/prombench` cloning the repo beforehand and placing it in a shared directory will solve this issue.

nikita-mk:
https://developer.github.com/actions/creating-github-actions/accessing-the-runtime-environment/#filesystem
Github actions has `/github/workspace` as working directory. It contains copy of the repository that triggered the workflow.

cmd/prombench/Dockerfile
```diff
@@ -13,6 +13,8 @@ RUN \
 
 COPY prombench /prombench/prombench
 COPY Makefile /prombench/Makefile
+COPY Makefile.common /prombench/Makefile.common
 COPY manifests/prombench /prombench/manifests/prombench
 
 WORKDIR /prombench
+ENTRYPOINT [ "make", "-C","/prombench/","deploy"]
```
nikita-mk, line 20: The `ENTRYPOINT` instruction gets executed in `/github/workspace`.

Contents of the Dockerfile are present in `/prombench` dir.
Setting PROMBENCH_PATH is a good idea. We just need to pass this as an env var. I will test this

krasi-georgiev:
The problem is that the workdir is set to `GITHUB_WORKSPACE` which is `/github/workspace` and it overwrites the default prombench workdir which is `/prombench`

to overcome this problem you use this
```
action "start_benchmark" {
  uses = "docker://prombench/prombench:2.0.1"
  runs = "bash"
  args = ["-c", "cd /prombench && make deploy"]
}
```

this way there would be no need to make any changes to the prombench image and just add the missing `Makefile.common`

nikita-mk:
I tested this and it is worked.

krasi-georgiev:
If it worked do you prefer to keep the PR as is or will update it?

nikita-mk:
Actually, the only change we need is copying Makefile.common. That is being solved in another PR. 
I will close this PR

nikita-mk:
See #228 for parameters to be given as file paths

PR #206, opened 2019-06-23
Added k8s jobs apply, wait and delete.

geekodour:
This will be needed to run [config-bootstrapper](https://github.com/kubernetes/test-infra/tree/master/prow/cmd/config-bootstrapper) 

this will help in getting rid of all the configmaps from `/manifest` directory and enable config files/alert rules etc to be stored and edited directly in yaml format.

Signed-off-by: Hrishikesh Barman <hrishikeshbman@gmail.com>

Issue #117, opened 2018-08-06
Think how to simplify the prombench Dockerimage to NOT build a new one on every config file change


krasi-georgiev:
```
COPY components/prombench/nodepools.yaml /prombench/components/prombench/nodepools.yaml
COPY components/prombench/manifests/benchmark /prombench/components/prombench/manifests/benchmark
```

Currently we copy the configs inside the image itself so any change to these will requires a new image to be build and pushed.

I am thinking if these are either pulled from master from within the prombench image or added as config maps would this simplify the workflow?

krasi-georgiev:
> Create all files present in components/prombench/manifests/benchmark as
config-maps and volume-mount them in prowjobs.
We can use https://github.com/kubernetes/git-sync or https://github.com/WanderaOrg/git2kube to keep the configmap in sync with the git repository.
(@krasi-georgiev might not like adding new dependencies stuck_out_tongue_closed_eyes )

 very good ideas, but new dependancies require more maintenance :) lets leave this as a last resource.



> Implement something like kubectl apply in the prombench-cli tool.
https://kubernetes.io/docs/concepts/cluster-administration/manage-deployment/#kubectl-apply.
When we pass a http address it will pull it from there.

I am leaning towards this idea with the only drawback that we need to either drop support for parsing entyre folders, which is not a very big deal just more verbose (to specify each file) or implement 
https://github.com/src-d/go-git with 
https://github.community/t5/How-to-use-Git-and-GitHub/How-can-I-download-a-specific-folder-from-a-GitHub-repo/m-p/95#M23

> Implement a new tool which clones the prombench repo in the prowjobs and deploys the yaml files in components/prombench/manifests/benchmark
https://github.community/t5/How-to-use-Git-and-GitHub/How-can-I-download-a-specific-folder-from-a-GitHub-repo/m-p/95#M23
https://github.com/src-d/go-git

don't need a new tool for this can all be done in prombench

> Create a new prow-plugin or modify prometheus/test-infra/prow/plugins/updateconfig plugin.
(This method has a low priority because if we decide not to use prow one day, we want the transition to be as smooth as possible.)

yes will try this of everything else fails.

geekodour:
in my experience, the current docker image is not completely standalone. 
- [/components/prombench/manifests/results/authenticate_rbac.yaml](https://github.com/prometheus/prombench/blob/master/components/prombench/manifests/results/authenticate_rbac.yaml)
[...]

PR #227, opened 2019-07-21
Add config bootstrapper

### LINKED ISSUE #223: Update Dockerfile and image for prombench
In addition to some changes done recently, another change that is needed in the `prombench` `Dockerfile` is that `Makefile` now includes `Makefile.common` which uses the `go` binary. 

https://github.com/prometheus/prombench/blob/7d2e3503af0ea3f47561d4b5c6f4837973146f3e/Makefile#L3

https://github.com/prometheus/prombench/blob/7d2e3503af0ea3f47561d4b5c6f4837973146f3e/Makefile.common#L35

So we probably need to add a base image that has the go binary in it, otherwise install go in the current base image inside the `Dockerfile`. 

[Current Dockerfile](https://github.com/prometheus/prombench/blob/master/cmd/prombench/Dockerfile) also do not contain `Makefile.common` which breaks `make` but not having `go` installed in the base image just prints warnings and does not break `make`

geekodour:
This PR adds [config-bootstrapper](https://github.com/kubernetes/test-infra/tree/master/prow/cmd/config-bootstrapper) to prombench. this bootstraps configmaps used in various areas in prombench.

 upgrading of prow with the [updateconfig](https://github.com/kubernetes/test-infra/tree/master/prow/plugins/updateconfig) will be done in a follwing pr. this will allow updating of configmaps directly when prs are merged.

closes #206 
closes #223  ( base image with go binary still needs to be added )
also adds `GOOGLE_APPLICATION_CREDENTIALS` to the readme for #222


README.md
```diff
@@ -24,6 +24,7 @@ export PROJECT_ID=<google-cloud project-id>
 export CLUSTER_NAME=prombench
 export ZONE=us-east1-b
 export AUTH_FILE=<path to service-account.json>
+export GOOGLE_APPLICATION_CREDENTIALS=<path to service-account.json>
```
krasi-georgiev, line 27: lets also add a comment why is this needed and that the long plan is to remove it.

pkg/provider/k8s/k8s.go
```diff
@@ -1257,6 +1326,34 @@ func (c *K8s) statefulSetReady(resource runtime.Object) (bool, error) {
...
+	switch v := resource.GetObjectKind().GroupVersionKind().Version; v {
+	case "v1":
+		client := c.clt.BatchV1().Jobs(req.Namespace)
+
+		res, err := client.Get(req.Name, apiMetaV1.GetOptions{})
+		if err != nil {
+			return false, errors.Wrapf(err, "Checking Job resource:'%v' status failed err:%v", req.Name, err)
+		}
+
+		// current jobReady only works for non-parallel jobs
```
krasi-georgiev: not sure how/why is this comment usefull?


And all golang comments should start with a capital letter and end with a full stop

krasi-georgiev (commented):
seems like many files are just renamed. Can you try to use `git mv` so that it shows the diff only as a file rename ``

pkg/provider/k8s/k8s.go
geekodour: currently we only need non-parallel jobs. incase someone tries to add a parallel job someday and prombench fails this comment can be useful. I've not tested this with parallel jobs so I changed the word to `// Current jobReady function only tested with non-parallel jobs.`

added the changes.

README.md
geekodour: okay

geekodour:
> seems like many files are just renamed. Can you try to use git mv so that it shows the diff only as a file rename 

is this still required? github does show the files that are only renamed like this:
```
manifests/cluster-infra/1_gcloud_rolebinding.yaml → manifests/cluster-infra/1a_gcloud_rolebinding.yaml
```
image example:
image example

cc @krasi-georgiev

pkg/provider/k8s/k8s.go
krasi-georgiev: If that is the case should add it at the top as a comment to this entire function.

krasi-georgiev:
will this allow us to use just the config updater without the other prow components?
If not which are the minimal components we need to have this working?

geekodour:
These are the only config files needed to run `config-bootstrapper` , the configmap `prow-config` added in `plugins.yaml` is not even required but I added it because we'll need it anyway when adding the [updateconfig](https://github.com/kubernetes/test-infra/tree/master/prow/plugins/updateconfig) plugin for prow

`/prombench/config/prow/config.yaml`
```
prowjob_namespace: default
pod_namespace: default
log_level: debug
```
`/prombench/config/prow/plugins.yaml`
```
config_updater:
  maps:
    config/prow/config.yaml:
      name: prow-config
    config/prow/plugins.yaml:
      name: prow-config
    config/prometheus-meta/prometheus.yaml:
      name: prometheus-meta
    config/grafana/prometheus-datasource.yaml:
      name: grafana-datasources
    config/grafana/prometheus-dashboard.yaml:
      name: grafana-dashboard-provision
    config/grafana/dashboards/*.json:
      name: grafana-dashboards
```

no other prow components is needed. That's why `config-bootstrapper` is added in the `manifest/cluster-infra` directory before any prow component is even added.

pkg/provider/k8s/k8s.go
geekodour: okay, will add the change

geekodour:
The prow configmaps added here are not used by prow. that might be a confusion,  I'll add a comment in the new pr.

krasi-georgiev:
My question was more if we would still need to run a full prow installation even after we migrate to GH actions? If that is the case maybe there is some other standalone tool that can do the same job with far less maintenance overhead.

krasi-georgiev:
Can you also explain in more details the problem we are trying to solve.
 
IIRC one problem was that the configs when deploying a test are in the image itself.
 https://github.com/prometheus/prombench/issues/117
I think the solution to this problem should be that these need to be added as config maps.
The second problem is probably when we have changes to other existing config maps. 

Can't we solve both of these problems with simple tools like:  https://github.com/WanderaOrg/git2kube - monitor git files and update the config maps from them
https://github.com/stakater/reloader - restart the pods when the config map changes
There is also the sidecar solution which send a reload signal rather than doing a rolling update: https://github.com/coreos/prometheus-operator/tree/master/cmd/prometheus-config-reloader

With any solution we choose we also need to start  think how to test that after the config updates we can still deploy the bench test as expected.

geekodour:
no we won't need to run a full prow installation to run `config-bootstrapper` we don't need any other prow component. `config-bootstrapper` is not a core prow component itself. 

> IIRC one problem was that the configs when deploying a test are in the image itself.

This PR does not attempt to solve this, the `statefulset` prombench tests pr(will be sending soon) attempts to solve that by cloning and mounting that volume for prombench to use. On the other hand, when using github actions, we should be able to clone the latest prombench and use the `manifests/prombench` and `Makefile` from it.

> Can you also explain in more details the problem we are trying to solve.

Since we're adding more components to prombench, having a way to store configurations in their native form helps. for eg. now we can directly update grafana dashboard json without having to add any extra escape sequence for golang templates.

secondly after we migrate to GitHub actions we'll only need to have the `hook` prow component if we want to have the `update-config` plugin (which will update configmaps directly from github merges) . `config-bootstrapper` is a k8s job does not need any prow component and only updates the configmaps during initial start.

krasi-georgiev:
just realized that in the case where we use the prombench image with github actions it would have no way to access the `manifests/` files as the command runs within the github action environment and NOT in the Google cloud cluster.

krasi-georgiev:
to solve this @nikita-mk 	need to create another action "clonePrombenchRepo" that will clone the prombench repo and save it in `/github/workspace/prombench`
this way the `prombenchDeploy` action would be able to access the manifest files.

So the makefile can be changed to something like:
```
MANIFESTS               ?= manifests/prombench 
# Set the env only if not set. 
# This allows it to set it from github actions to "/github/workspace/prombench/manifests/prombench", 
# but still use the makefile without a docker image locally.

nodepool_create:
	$(PROMBENCH_CMD) gke nodepool create -a ${AUTH_FILE} \
		-f ${MANIFESTS}/nodepools.yaml

PR #228, opened 2019-07-23
Parameters now can be path to file

nikita-mk:
This change will check if the provided parameter is a file path and read its value. The file must contain the value of the parameter.
Does nothing if the parameter is an actual value and not a file path.

### DIFF
--- a/pkg/provider/provider.go
+++ b/pkg/provider/provider.go
@@ -80,6 +80,17 @@ func DeploymentsParse(deploymentFiles []string, deploymentVars map[string]string
 		}
 	}
 
+	for k, v := range deploymentVars {
+		_, err := os.Stat(v)
+		if err == nil && v != "prombench" {
+			val, e := ioutil.ReadFile(v)
+			if e != nil {
+				log.Fatalf("couldn't read var file")
+			}
+			deploymentVars[k] = string(val)
+		}
+	}
+
 	deploymentObjects := make([]Resource, 0)
 	for _, name := range fileList {
 		content, err := applyTemplateVars(name, deploymentVars)


nikita-mk:
Fails if the parameter is a file path but doesn't exist.
This will consider the wrong path as the parameters value.

pkg/provider/provider.go
```diff
@@ -80,6 +80,17 @@ func DeploymentsParse(deploymentFiles []string, deploymentVars map[string]string
 		}
 	}
 
+	for k, v := range deploymentVars {
+		_, err := os.Stat(v)
```
krasi-georgiev, line 84: why not read the file directly like in:
https://github.com/prometheus/prombench/blob/b8b29233ce43946a6e13407dc461025ff42a1966/pkg/provider/gke/gke.go#L75

pkg/provider/provider.go
```diff
@@ -80,6 +80,17 @@ func DeploymentsParse(deploymentFiles []string, deploymentVars map[string]string
...
+		if err == nil && v != "prombench" {
```
krasi-georgiev, line 85: `v != "prombench"` - this looks wrong, why do we need it?

krasi-georgiev (commented):
Have you tested it github actions?
The idea is that with this change we should be able to use the prombench image directly.

pkg/provider/provider.go
nikita-mk: Prombench binary exists in the directory and we are checking if some file with given name exists.
It reads the binary and fails.

pkg/provider/provider.go
nikita-mk: This will also read the binary file "prombench" as `CLUSTER_NAME` parameter has value prombench.

nikita-mk:
I have tested with github actions and it is working.

pkg/provider/provider.go
krasi-georgiev: Sorry I didn't understand what you mean.

nikita-mk: eg. /home/github/ARG_0 file exists and err will be nil. 

We have `CLUSTER_NAME` whose value is prombench. Prombench binary file named "prombench" exists. As err is nil, the condition gets satisfied and `ioutil.ReadFile()` will read the file.

It tries to read the binary file and assign that value to `CLUSTER_NAME`. This results in failure.

krasi-georgiev: In that case lets revert the logic, first check if the env variable exists and only if it doesn't read it from a file.

krasi-georgiev: maybe also implement this for the auth file which I added recently

nikita-mk: We are setting the env variable to the path of the file

krasi-georgiev: I didn't understand what you mean?

krasi-georgiev: ping

geekodour:
this is no longer needed ~after~ if #244 is added, then we can pass the `PR_NUMBER` and `RELEASE` as such:

```

action "start_benchmark" {
  needs = ["start_benchmark_validate"]
  uses = "docker://prombench/prombench:2.0.2"
  args = [
    "export RELEASE=$(cat /github/home/ARG_0) && [ -z $RELEASE ] && export RELEASE=master",
    "export PR_NUMBER=$(cat /github/home/ARG_1);",
    "make",
    "deploy"
  ]
  secrets = ["AUTH_FILE"]
  env = {
    PROJECT_ID="prombench-gsoc",
    CLUSTER_NAME="prombench",
    ZONE="us-central1-a",
    DOMAIN_NAME="http://prombench.prometheus.io",
    PROMBENCH_REPO="https://github.com/testpromorg/prombench.git"
	}
}
```

krasi-georgiev:
Ok

PR #230, opened 2019-07-24
kube-state-metrics

### LINKED ISSUE #76: Add kube-state-metrics deployment and grafana graphs
It will be good to have grafana dashboards that display cluster status, so that users can view what amount of load is being put onto the prometheus servers.
I have tried a few grafana dashboards like [kube-prometheus](https://github.com/coreos/prometheus-operator/tree/master/contrib/kube-prometheus), etc but I wasn't fully satisfied with the graphs.

[grafana-kubernetes-app](https://github.com/grafana/kubernetes-app) looks like a good option. Instead of using grafana-plugin, can convert the [grafana-kubernetes-app-dashboards](https://github.com/grafana/kubernetes-app/tree/master/src/dashboards) for prombench.

geekodour:
closes #76

Adding kube-state-metrics to use the `kube_namespace_created` metric to send alerts to github for long running benchmark tests.

@krasi-georgiev do we need to limit exposed metrics to `kube_namespace_created` only? or for current setup all metrics exposed will be fine? 
https://github.com/kubernetes/kube-state-metrics/blob/master/docs/cli-arguments.md

### DIFF
--- a/manifests/cluster-infra/4_kube-state-metrics.yaml
+++ b/manifests/cluster-infra/4_kube-state-metrics.yaml
@@ -3,7 +3,6 @@ kind: ServiceAccount
 metadata:
   name: kube-state-metrics
   namespace: kube-system
-
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
@@ -101,6 +100,8 @@ spec:
       containers:
       - name: kube-state-metrics
         image: quay.io/coreos/kube-state-metrics:v1.7.1
+        args:
+        - "--collectors=namespaces"
         ports:
         - name: http-metrics
           containerPort: 8080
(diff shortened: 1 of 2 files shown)


manifests/cluster-infra/4_kube-state-metrics.yaml
```diff
@@ -0,0 +1,137 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: kube-state-metrics
+  namespace: kube-system
+
```
krasi-georgiev:
```suggestion
```

krasi-georgiev:
> @krasi-georgiev do we need to limit exposed metrics to kube_namespace_created only? or for current setup all metrics exposed will be fine?

I would say lets keep the exposed metrics to a minimum and only add new ones as we need them.

PR #236, opened 2019-07-29
Added the ability to specify the pull request comment as a template.

VladMasarik:
Added the ability to specify the pull request comment as an environment variable. 
It is possible to reference the environment variables inside the comment using in similar manner: 
`TEMPLATE="The benchmark is starting. Your Github token is {{ index . \"GITHUB_TOKEN\" }}."`

### DIFF
--- a/cmd/tools/commentMonitor/main.go
+++ b/cmd/tools/commentMonitor/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io/ioutil"
@@ -9,6 +10,8 @@ import (
 	"path/filepath"
 	"regexp"
 	"strconv"
+	"strings"
+	"text/template"
 
 	"github.com/google/go-github/v26/github"
 	"golang.org/x/oauth2"
@@ -16,10 +19,12 @@ import (
 )
 
 func main() {
-	app := kingpin.New(filepath.Base(os.Args[0]), "commentMonitor github comment extract\n ./commentMonitor -i /path/event.json -o /path \"^myregex$\"")
+	app := kingpin.New(filepath.Base(os.Args[0]), "commentMonitor github comment extract\n ./commentMonitor -i /path/event.json -o /path \"^myregex$\"\nExample of comment template environment variable: COMMENT_TEMPLATE=\"The benchmark is starting. Your Github token is {{ index . \\\"GITHUB_TOKEN\\\" }}.\"")
 	app.HelpFlag.Short('h')
 	input := app.Flag("input", "path to event.json").Short('i').Default("/github/workflow/event.json").String()
 	output := app.Flag("output", "path to write args to").Short('o').Default("/github/home").String()
+	verifyUser := app.Flag("verify-user", "If set to true, a check, on whether the comment creator is a collaborator or a member of the group, will be executed.").Default("true").Bool()
+	templateEnvVar := app.Flag("template-var", "Name of the environment variable that contains comment template.").Short('t').Default("COMMENT_TEMPLATE").String()
 	regex := app.Arg("regex", "Regex pattern to match").Required().String()
 	kingpin.MustParse(app.Parse(os.Args[1:]))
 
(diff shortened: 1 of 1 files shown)


cmd/tools/commentMonitor/main.go
```diff
@@ -66,27 +66,25 @@ func main() {
...
+		envvar := make(map[string]string)
+		for _, e := range env {
+			tmp := strings.Split(e, "=")
+			envvar[tmp[0]] = tmp[1]
+		}
 
-The Prometheus servers being benchmarked can be viewed at :
-- PR - [prombench.prometheus.io/%v/prometheus-pr](%v/%v/prometheus-pr)
-- %v - [prombench.prometheus.io/%v/prometheus-release](%v/%v/prometheus-release)
+		unformatedComment := os.Getenv("TEMPLATE_BENCH")
```
krasi-georgiev: this needs to be passed as an arg as well.


```
template := app.Arg("template", "Template text so use when posting a github comment. All env variables will be replaces with.......").Required().String()

var unformatedComment string
unformatedComment = os.Getenv(template)
if unformatedComment ==""{
	unformatedComment = os.ReadFile(template)
}

if unformatedComment == "" ...fail
```

cmd/tools/commentMonitor/main.go
```diff
@@ -132,6 +130,10 @@ func writeArgs(args []string, output string) {
 		if err != nil {
 			log.Fatalln(err)
 		}
+		err = os.Setenv(filename, string(data))
```
krasi-georgiev, line 140: why is this needed?

krasi-georgiev (commented):
just a bit if clarification for @nikita-mk .

this change is required so that the same tool can be used running benchmark tests for tsdb. The comment for this test is different so we need a way to set this dynamically. 

After this change the template text would need to be set as an env variable.

krasi-georgiev:
Maybe also add an example template to give an idea of how it would look like.

cmd/tools/commentMonitor/main.go
VladMasarik: Since we want to use env vars in the template and `args` such as PR number and branch are not env vars, I had to export those arguments so that I can use them in the comment.

cmd/tools/commentMonitor/main.go
VladMasarik: Does it have to be an argument or is `flag` enough?

VladMasarik:
> Maybe also add an example template to give an idea of how it would look like.

I hoped that the example in the starting comment is enough, but I added a larger example.

cmd/tools/commentMonitor/main.go
krasi-georgiev: yes flag is better.

krasi-georgiev:
> I hoped that the example in the starting comment is enough, but I added a larger example.


I meant somewhere in main.go so that when people do ... -h it shows an example template to start with.

cmd/tools/commentMonitor/main.go
krasi-georgiev: ack. Could you add a comment like
```
// Also add all comment args as env variables so that these can be used in the comment template.
```

cmd/tools/commentMonitor/main.go
```diff
@@ -66,27 +67,25 @@ func main() {
...
+		envvar := make(map[string]string)
+		for _, e := range env {
+			tmp := strings.Split(e, "=")
+			envvar[tmp[0]] = tmp[1]
+		}
 
-The Prometheus servers being benchmarked can be viewed at :
-- PR - [prombench.prometheus.io/%v/prometheus-pr](%v/%v/prometheus-pr)
-- %v - [prombench.prometheus.io/%v/prometheus-release](%v/%v/prometheus-release)
+		unformatedComment := os.Getenv(templateEnvVar)
```
krasi-georgiev:
```suggestion
		unformatedComment := os.Getenv(*templateEnvVar)
```

cmd/tools/commentMonitor/main.go
VladMasarik: Added the comment.

VladMasarik:
I renamed some variables because only now I noticed that I used the ones when I needed a quick variable name.

krasi-georgiev:
ok, ping me when ready for a final review after adding the skip-member-validation and posting a comments when the validation fails.

krasi-georgiev:
good work, Thanks!

krasi-georgiev:
@nikita-mk 	now you need to set a new env var that would include the template text.

nikita-mk:
I am working on it.

PR #238, opened 2019-08-02
Add cd into prombench directory

nikita-mk:
To make the output of subsequent commands visible in the github actions log.
We just need to pass args now.

### DIFF
--- a/cmd/prombench/clone.sh
+++ b/cmd/prombench/clone.sh
@@ -8,4 +8,4 @@ cp /usr/bin/prombench $PROMBENCH_DIR/
 
 cd $PROMBENCH_DIR
 # execute arguments passed to the image
-$*
+$@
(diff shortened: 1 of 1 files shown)


cmd/prombench/clone.sh
```diff
@@ -6,5 +6,6 @@ git clone $PROMBENCH_REPO $PROMBENCH_DIR
 # copy the prombench binary to the cloned directory
 cp /usr/bin/prombench $PROMBENCH_DIR/
 
+cd $PROMBENCH_DIR
 # execute arguments passed to the image
-$@
+$*
```
krasi-georgiev: why this change?

nikita-mk: We tried this for github actions. But both `*` and `@` work.

krasi-georgiev: So what is the answer to the question why do we need this change?

nikita-mk: In github actions example, `*` was used to accept the args. So we thought that might be the case that github actions was not executing `make` commands. But adding `cd $PROMBENCH_DIR` resolved it. Will revert it back to `@`.

PR #241, opened 2019-08-06
[RDY] Adding Loki

### LINKED ISSUE #157: Display deployment logs
We should look into displaying some logs from some deployments for troubleshooting and debug information.

for example for both tested prometheus servers - we want to see if the stdout/stderr shows any errors during the benchmark.

Currently these are only visible in the GKE console.

geekodour:
Todo:
- [x] Add Loki
- [x] Add Promtail
- [x] Connect grafana, add to datasource provisions

Closes #157 
 

geekodour:
@krasi-georgiev I think it's ready for review.

image

manifests/cluster-infra/6b_loki_stateful_set.yaml
```diff
@@ -0,0 +1,140 @@
...
+  ports:
+    - port: 3100
+      protocol: TCP
+      name: http-metrics
+      targetPort: http-metrics
+  selector:
+    app: loki
+---
+apiVersion: apps/v1
+kind: StatefulSet
```
krasi-georgiev, line 61: why do we need it as a stateful set?

manifests/cluster-infra/grafana_deployment.yaml
```diff
@@ -37,6 +37,8 @@ spec:
             value: "Main Org."
           - name: GF_SECURITY_ADMIN_PASSWORD
             value: "{{ .GRAFANA_ADMIN_PASSWORD }}"
+          - name: GF_USERS_VIEWERS_CAN_EDIT
```
krasi-georgiev, line 40: this means theat anon access also allows to edit the dashboards right?

I am not sure this is a good idea.

manifests/prombench/benchmark/3c_promtail-test_deployment.yaml
```diff
@@ -0,0 +1,249 @@
...
+        target_label: __path__
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: promtail-test-pr-{{ .PR_NUMBER }}
+  namespace: prombench-{{ .PR_NUMBER }}
+  labels:
+    app: promtail
+    promtail: test-pr-{{ .PR_NUMBER }}
```
krasi-georgiev:
```suggestion
    promtail: pr-{{ .PR_NUMBER }}
```

krasi-georgiev:
Can you also open a separate draft PR for updating the readme and the chematics. 
doesn't need to be complete just some start so we don't forget and will update when we start testing this new setup.

manifests/cluster-infra/grafana_deployment.yaml
geekodour: They will be able to `view` and `edit` but not `save` , this option is required if we want to expose the grafana explorer page. (which we need to display loki logs publicly)

https://grafana.com/docs/installation/configuration/#viewers-can-edit

manifests/prombench/benchmark/3c_promtail-test_deployment.yaml
```diff
@@ -0,0 +1,249 @@
...
+        target_label: promtail
+      - replacement: /var/log/pods/*$1/*.log
+        separator: /
+        source_labels: [__meta_kubernetes_pod_uid, __meta_kubernetes_pod_container_name]
+        target_label: __path__
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: promtail-test-pr-{{ .PR_NUMBER }}
```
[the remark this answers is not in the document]
geekodour: okay will add this change, maybe can add another pr to remove the other `test-`s later.
I think it's related to https://github.com/prometheus/prombench/pull/89

manifests/cluster-infra/grafana_deployment.yaml
krasi-georgiev: No problem than.

manifests/prombench/benchmark/3c_promtail-test_deployment.yaml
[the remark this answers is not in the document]
krasi-georgiev: yes that would be great, and can probably remove prow mentions everywhere as well.

manifests/cluster-infra/6b_loki_stateful_set.yaml
geekodour: It's not needed but I think it's some kind of best practice to use statefulset when using persistent storage. If we follow that it'll be ideal to migrate `prometheus-meta` to `StatefulSet` too at some point. I've noticed it in some other repository too.

https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/prometheus/prometheus-statefulset.yaml

But  there's an issue with wait time in prombench for `StatefulSet` which needs more investigation.

Should I change this to Deployment? (The actual loki helm charts are using `StatefulSet`)

manifests/prombench/benchmark/3c_promtail-test_deployment.yaml
```diff
@@ -0,0 +1,249 @@
+apiVersion: v1
```
krasi-georgiev: 3c_promtail-test_deployment.yaml -> 3c_promtail-benchtest_deployment.yaml

manifests/prombench/benchmark/3c_promtail-test_deployment.yaml
```diff
@@ -0,0 +1,249 @@
...
+      affinity:
+        podAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+          - topologyKey: kubernetes.io/hostname
+            labelSelector:
+              matchExpressions:
+              - key: prometheus
+                operator: In
+                values:
+                - test-pr-{{ .PR_NUMBER }}
```
krasi-georgiev: this needs to stay as is because of the other deployments right?

manifests/prombench/benchmark/3c_promtail-test_deployment.yaml
```diff
@@ -0,0 +1,249 @@
...
+      affinity:
+        podAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+          - topologyKey: kubernetes.io/hostname
+            labelSelector:
+              matchExpressions:
+              - key: prometheus
+                operator: In
+                values:
+                - test-{{ normalise .RELEASE }}
```
krasi-georgiev: can we have this as
```suggestion
                - benchtest-{{ normalise .RELEASE }}
```

krasi-georgiev:
[design.zip](https://github.com/prometheus/prombench/files/3509226/design.zip)


Much better now.

I aligned few things, can you try this one?

PR #243, opened 2019-08-07
Set GOOGLE_APPLICATION_CREDENTIALS as an env var in NewGKEClient

geekodour:
removed setting of `GOOGLE_APPLICATION_CREDENTIALS` in Makefile and `README.md`

### DIFF
--- a/pkg/provider/gke/gke.go
+++ b/pkg/provider/gke/gke.go
@@ -441,6 +441,9 @@ func (c *GKE) NewK8sProvider(*kingpin.ParseContext) error {
 		return fmt.Errorf("missing required CLUSTER_NAME variable")
 	}
 
+	// Set GOOGLE_APPLICATION_CREDENTIALS
+	os.Setenv("GOOGLE_APPLICATION_CREDENTIALS", c.Auth)
+
 	// Get the authentication certificate for the cluster using the GKE client.
 	req := &containerpb.GetClusterRequest{
 		ProjectId: projectID,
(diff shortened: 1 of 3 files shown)


pkg/provider/gke/gke.go
```diff
@@ -66,6 +66,7 @@ type GKE struct {
 func (c *GKE) NewGKEClient(*kingpin.ParseContext) error {
 	// Set the auth env variable needed to the gke client.
 	if c.Auth != "" {
+		os.Setenv("GOOGLE_APPLICATION_CREDENTIALS", c.Auth)
```
krasi-georgiev: I would say lets add this to `NewK8sProvider` as it is really needed there.

geekodour: since `NewGKEClient` is called first in prombench,
https://github.com/prometheus/prombench/blob/2eda25459df65351dbdc8d71203ea1840530812d/cmd/prombench/prombench.go#L22

it already sets `c.Auth` to the content of the file:
https://github.com/prometheus/prombench/blob/2eda25459df65351dbdc8d71203ea1840530812d/pkg/provider/gke/gke.go#L73-L77

so if we do `os.Setenv("GOOGLE_APPLICATION_CREDENTIALS", c.Auth)` in `NewK8sProvider` , `GOOGLE_APPLICATION_CREDENTIALS` gets set to the content of `c.Auth` which is not expected here: https://cloud.google.com/docs/authentication/getting-started#setting_the_environment_variable

should we create another variable to store the content of the file then?

PR #244, opened 2019-08-08
move base64 decoding before data is written to the temporary file (NewGKEClient)

geekodour:
The file `GOOGLE_APPLICATION_CREDENTIALS` should not be base64 encoded, so we decode it before we write it to the fs.

### DIFF
--- a/pkg/provider/gke/gke.go
+++ b/pkg/provider/gke/gke.go
@@ -73,36 +73,35 @@ func (c *GKE) NewGKEClient(*kingpin.ParseContext) error {
 	// When the auth variable points to a file
 	// put the file content in the variable.
 	if content, err := ioutil.ReadFile(c.Auth); err == nil {
-		// Set the auth env variable needed to the k8s client.
-		// The client looks for this special variable name and it is the only way to set the auth for now.
-		// TODO Remove when the client supports an auth config option in NewDefaultClientConfig.
-		os.Setenv("GOOGLE_APPLICATION_CREDENTIALS", c.Auth)
 		c.Auth = string(content)
-	} else {
-		// Check if auth data is base64 encoded and decode it.
-		encoded, err := regexp.MatchString("^([A-Za-z0-9+/]{4})*([A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{2}==)?$", c.Auth)
+	}
+
+	// Check if auth data is base64 encoded and decode it.
+	encoded, err := regexp.MatchString("^([A-Za-z0-9+/]{4})*([A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{2}==)?$", c.Auth)
+	if err != nil {
+		return err
+	}
+	if encoded {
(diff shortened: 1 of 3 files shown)


pkg/provider/gke/gke.go
```diff
@@ -79,6 +79,20 @@ func (c *GKE) NewGKEClient(*kingpin.ParseContext) error {
 		os.Setenv("GOOGLE_APPLICATION_CREDENTIALS", c.Auth)
 		c.Auth = string(content)
 	} else {
+		// Check if auth data is base64 encoded and decode it.
```
krasi-georgiev: For complitness I think it is better to check and decode in both cases, when it is a file and when it is a data string in an env var.

geekodour: In that case, if we find that the file is base64  encoded, we'll need to override the file with the decoded data because `GOOGLE_APPLICATION_CREDENTIALS` expects the path of the file, content of the file should not be base64 encoded.

geekodour: or we can:
1. check if it's file or data
2. store the content of the file in `Auth` if it's a file
3. check if `Auth` is encoded, decode it,
4. create a temporary file to store the decoded data
5. `os.SetEnv(GOOGLE_APPLICATION_CREDENTIALS, toTheTempFile)`

what do  you think? this way create a temporary file will be created no matter what and there's only one call to set `GOOGLE_APPLICATION_CREDENTIALS`

krasi-georgiev: yes that sound good.

geekodour: updated.

pkg/provider/gke/gke.go
```diff
@@ -73,36 +73,35 @@ func (c *GKE) NewGKEClient(*kingpin.ParseContext) error {
...
-		}
-		os.Setenv("GOOGLE_APPLICATION_CREDENTIALS", saFile.Name())
+	// Create tempory file to store the credentials.
+	saFile, err := ioutil.TempFile("", "service-account")
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer saFile.Close()
+	if _, err := saFile.Write([]byte(c.Auth)); err != nil {
+		log.Fatal(err)
```
krasi-georgiev: Recently I learned that log.Fatal calles os.Exit and this way defer statments are never executed.
Maybe try to just return the error here.

I would say we should never use log.Fatal and refactor to return the error and handle it in the parent func.

pkg/provider/gke/gke.go
```diff
@@ -103,11 +89,25 @@ func (c *GKE) NewGKEClient(*kingpin.ParseContext) error {
 		c.Auth = string(auth)
 	}
 
+	// Create tempory file to store the credentials.
+	saFile, err := ioutil.TempFile("", "service-account")
+	if err != nil {
+		return err
```
```suggestion
		return errors.Wrap(err,"create temp file")
```

To make it more clear which part of the code returned the error.

The same for the one below

PR #245, opened 2019-08-08
updated prombench docker image to 2.0.2 and docs for GitHub Actions

geekodour:

### DIFF
--- a/manifests/prow/components/prow_internals_2.yaml
+++ b/manifests/prow/components/prow_internals_2.yaml
@@ -217,7 +217,7 @@ data:
         agent: kubernetes
         spec:
           containers:
-          - image: docker.io/prombench/prombench:2.0.1
+          - image: docker.io/prombench/prombench:2.0.2
             imagePullPolicy: Always
             args:
             - "make"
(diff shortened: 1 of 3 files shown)


manifests/prow/components/prow_internals_2.yaml
```diff
@@ -217,7 +217,7 @@ data:
         agent: kubernetes
         spec:
           containers:
-          - image: docker.io/prombench/prombench:2.0.1
```
krasi-georgiev, line 220: why not do it as part of the whole change to github actions with the readme update etc?

geekodour:
With this we'll have to update the `main.workflow` file manually in `prometheus/prometheus` .

I've written a PoC for syncing `main.workflow` files for different repositories.

https://gist.github.com/geekodour/17dbfbcb560d7d665266d84a1825644f

the only downside I see to this automation that, if repo owners want to add non-prombench specific workflows and actions then they'll have to add it to `prometheus/prombench` and prombot will send a pr to sync it which will be not natural.

That's why I created the `/examples` directory with a sample `main.workflow` that repo owners can use however they want. 

currently the only repo that is using prombench is `prometheus/prometheus`

geekodour:
Superseded by #248

PR #251, opened 2019-08-10
some typo fix

yeya24:
Signed-off-by: yeya24 <yb532204897@gmail.com>


.gitignore
```diff
@@ -2,3 +2,5 @@
 /fake-webserver
 /scaler
 /tools/amGithubNotifier/amGithubNotifier
+.idea/
```
krasi-georgiev: these are env specific so best to add them to .git/info/exclude

yeya24: I removed this. PTAL.

PR #240, opened 2019-08-06
Move all tools from /cmd/.. to /tools/...

nikita-mk:
Closes #239 
All cli tools in `/cmd` are moved to `/tools`. Now, only the prombench cli tool is present in`/cmd`.

Signed-off-by: nikita-mk <nikkikokitkar@gmail.com>

PR #253, opened 2019-08-12
moved cmd to tools

krasi-georgiev:
closes https://github.com/prometheus/prombench/pull/240


krasi-georgiev:
unfortunately we should stick to a single go.mod file in the root dir as adding go.mod to each tool messes the tests and the builds.

the builds can be easily fixed, but the tests not so easy.

The PR also fixes few other small issues to allow enabling  `make all` test

cc @nikita-mk 	@geekodour

.circleci/config.yml
```diff
@@ -17,20 +17,13 @@ jobs:
...
     - run:
-        command: make style unused vet build
+        command: make all
         environment:
           # By default Go uses GOMAXPROCS but a Circle CI executor has many
           # cores (> 30) while the CPU and RAM resources are throttled. If we
           # don't limit this to the number of allocated cores, the job is
           # likely to get OOMed and killed.
           GOOPTS: "-p 2"
-    - run: git diff --exit-code
```
krasi-georgiev, line 27: @geekodour 	any idea was why this here?

.circleci/config.yml
```diff
@@ -17,20 +17,13 @@ jobs:
...
-    - prometheus/store_artifact:
```
krasi-georgiev, line 28: the same with these, why do we need them now?

.circleci/config.yml
geekodour: I used them here because they were used in the prometheus/prometheus `.config.yml` 
I think it's a check to verify that no new modules were added and no other files get changed when running `make build` or any of the above make commands run before this is run.

.circleci/config.yml
geekodour: don't think we need them here, in prometheus these are used to build the docker images afaik.

.circleci/config.yml
krasi-georgiev: I don't see it anymore there so shall we leave it out as well?
https://github.com/prometheus/prometheus/blob/master/.circleci/config.yml

.circleci/config.yml
krasi-georgiev: yes this is what I thought.

krasi-georgiev:
@geekodour 	any other comments or can I merge it?

.circleci/config.yml
geekodour: yes i think we can remove it as well.

PR #257, opened 2019-08-16
Add Funcbench feature and update the readme files.

VladMasarik:
This pull request allows usage of the funcbench feature, which is a benchmarking workflow for Prometheus time series database.

### DIFF
--- a/funcbench/README.md
+++ b/funcbench/README.md
@@ -1,0 +1,57 @@
+## Intro
+A tool used as a github action to run a `go test -bench` and compare changes from a PR against another branch. The benchmark is triggered by creating a comment which specifies a branch to compare. The results are then posted back as a PR comment. 
+Comparison is done using [benchcmp](https://godoc.org/golang.org/x/tools/cmd/benchcmp). Arguments for the benchcmp are read from files created by previous action (for example [commentMonitor](/tools/commentMonitor/main.go)), which is responsible for the comment parsing.
+
+## Use
+Tests are triggered by posting a comment in a PR with the following format:
+`/funcbench <branch> <golang test regex>  [-no-race]`
+Specifying which tests to run are filtered by using the standard golang regex format. 
+By default all benchmarks run with `-race` flag enabled and it can be disabled by appending `-no-race` at the end of the comment. 
+
+### Example Github actions workflow file
+```
+on: issue_comment // Workflow is executed when a pull request comment is created.
+name: Benchmark
+jobs:
+  commentMonitor:
+    runs-on: ubuntu-latest
+    steps:
+    - name: commentMonitor
+      uses: docker://prombench/comment-monitor:latest
+      env:
+        COMMENT_TEMPLATE: 'The benchmark has started.' // Body of a comment that is created to announce start of a benchmark.
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} // Github secret token
+      with:
+        args: '"^/funcbench ?(?P<BRANCH>[^ B\.]+)? ?(?P<REGEX>\.|Bench.*|[^ ]+)? ?(?P<RACE>-no-race)?.*$"'
+    - name: benchmark
+      uses: docker://prombench/funcbench:latest
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} // Github secret token
+```
+
+## Set up
+This tools is meant to be used as a Github action. The action itself is, to a large degree, unusable alone, as you need to combine it with another Github action that will provide necessary files to it. At this time, the only action it is supposed to work with, is [comment-monitor](https://github.com/prometheus/prombench/tree/master/tools/commentMonitor).
+- Create Github actions workflow file that is executed when an issue comment is created, `on = "issue_comment"`.
+- Add comment-monitor Github action as a first step.
+- Specify this regex `^/funcbench ?(?P<BRANCH>[^ B\.]+)? ?(?P<REGEX>\.|Bench.*|[^ ]+)? ?(?P<RACE>-no-race)?.*$` in the `args` field of the comment-monitor.
+- Specify this Github action as a pre-built image, build from this source code, or just refer to this repository from the workflow file.
+- Provide a Github token as an environment variable to both comment-monitor and funcbench.
+
+## How to build
+`docker build -t <tag of your choice> .`
+
+## Examples
+Execute benchmark named `FuncName` regex, and compare it with `master` branch.
+ ```
+ /funcbench master BenchmarkFuncName
+ ```
+
+Execute all benchmarks matching `FuncName.*` regex, and compare it with `master` branch.
+ ```
+ /funcbench master FuncName.*
+ ```
+
+Execute all benchmarks, and compere the results with `devel` branch.
+ ```
+ /funcbench devel .
+ ```
--- a/funcbench/main.go
+++ b/funcbench/main.go
@@ -1,0 +1,345 @@
+// Copyright 2019 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package main
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"github.com/google/go-github/github"
+	"github.com/pkg/errors"
+	"golang.org/x/oauth2"
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+type gitClient struct {
+	owner  string
+	branch string
+	repo   string
+}
+type gitHubClient struct {
+	owner            string
+	repo             string
+	latestCommitHash string
+	prNumber         int
+	client           *github.Client
+}
+
+type benchmarkTester struct {
+	bechRegex       string
+	home            string
+	raceFlagEnabled bool
+}
+
+func main() {
+	app := kingpin.New(filepath.Base(os.Args[0]), "benchmark result posting and formating tool.\n-i location of github hook file (even.json)")
+	app.HelpFlag.Short('h')
+	input := app.Flag("input", "path to event.json").Short('i').Default("/github/workflow/event.json").String()
+	kingpin.MustParse(app.Parse(os.Args[1:]))
+
+	data, err := ioutil.ReadFile(*input)
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	event, err := github.ParseWebHook("issue_comment", data)
+	if err != nil {
+		log.Fatalln(err)
+	}
+	var (
+		ghClient  *gitHubClient
+		gitClient *gitClient
+		prNumber  int
+	)
+	switch eventType := event.(type) {
+	case *github.IssueCommentEvent:
+		ghClient = newGitHubClient(eventType)
+		gitClient, err = newGitClient(eventType)
+		if err != nil {
+			log.Fatalln(err)
+		}
+		prNumber = *eventType.GetIssue().Number
+
+	default:
+		log.Fatalln("only issue_comment event is supported")
+	}
+	benchTest, err := newBenchmarkTester()
+	if err != nil {
+		log.Fatalln(err)
+	}
+	logLink := fmt.Sprintf("Full logs at: https://github.com/%s/%s/commit/%s/checks", ghClient.owner, ghClient.repo, ghClient.latestCommitHash)
+
+	if err := os.Chdir(os.Getenv("GITHUB_WORKSPACE")); err != nil {
+		log.Fatalln(err)
+	}
+
+	if err := gitClient.cloneRepository(); err != nil {
+		log.Fatalln(err)
+	}
+
+	if err := gitClient.checkoutPR(prNumber); err != nil {
+		comment := fmt.Sprintf("Switch to a pull request branch failed. %s", logLink)
+		if postCommentErr := ghClient.postComment(comment); postCommentErr != nil {
+			log.Fatalf("Error posting a comment for `checkoutToPullRequest` command execution error. checkoutToPullRequest err:%v, postComment err:%v", err, postCommentErr)
+		}
+		log.Fatalln(err)
+	}
+
+	prResults, err := benchTest.execBenchmark()
+	if err != nil {
+		comment := fmt.Sprintf("Go bench test for this pull request failed. %s", logLink)
+		if postCommentErr := ghClient.postComment(comment); postCommentErr != nil {
+			log.Fatalf("An error: %v occured while processing error: %v", postCommentErr, err)
+		}
+		log.Fatalln(err)
+	}
+
+	if err := gitClient.revertPRChanges(); err != nil {
+		log.Fatalln(err)
+	}
+
+	branchResults, err := benchTest.execBenchmark()
+	if err != nil {
+		comment := fmt.Sprintf("Go bench test for this pull request failed. %s", logLink)
+		if postCommentErr := ghClient.postComment(comment); postCommentErr != nil {
+			log.Fatalf("An error: %v occured while processing error: %v", postCommentErr, err)
+		}
+		log.Fatalln(err)
+	}
+
+	comparisonTable, err := benchTest.compareBenchmarks(branchResults, prResults)
+	if err != nil {
+		comment := fmt.Sprintf("Error: `benchcmp` failed. %s", logLink)
+		if postCommentErr := ghClient.postComment(comment); postCommentErr != nil {
+			log.Fatalf("An error: %v occured while processing error: %v", postCommentErr, err)
+		}
+		log.Fatalln(err)
+	}
+	if err := ghClient.postComment(comparisonTable); err != nil {
+		log.Fatalln(err)
+	}
+
+}
+
+func newGitHubClient(event *github.IssueCommentEvent) *gitHubClient {
+	c := gitHubClient{
+		client:           newClient(os.Getenv("GITHUB_TOKEN")),
+		owner:            *event.GetRepo().Owner.Login,
+		repo:             *event.GetRepo().Name,
+		prNumber:         *event.GetIssue().Number,
+		latestCommitHash: os.Getenv("GITHUB_SHA"),
+	}
+	log.Printf("GH Client: %+v\n", c)
+	return &c
+}
+
+func (c *gitHubClient) postComment(comment string) error {
+	issueComment := &github.IssueComment{Body: github.String(comment)}
+	_, _, err := c.client.Issues.CreateComment(context.Background(), c.owner, c.repo, c.prNumber, issueComment)
+	return err
+}
+
+func newGitClient(event *github.IssueCommentEvent) (*gitClient, error) {
+	_, _, err := execCommand("git", "config", "--global", "user.email", "prombench@example.com")
+	if err != nil {
+		return nil, err
+	}
+	_, _, err = execCommand("git", "config", "--global", "user.name", "Prombench Bot Junior")
+	if err != nil {
+		return nil, err
+	}
+
+	branch, err := ioutil.ReadFile("/github/home/commentMonitor/BRANCH")
+	c := gitClient{
+		branch: string(branch),
+		owner:  *event.GetRepo().Owner.Login,
+		repo:   *event.GetRepo().Name,
+	}
+
+	log.Printf("Git Client: %+v\n", c)
+	return &c, err
+}
+
+func (c *gitClient) cloneRepository() error {
+	_, _, err := execCommand("git", "clone", fmt.Sprintf("https://github.com/%s/%s.git", c.owner, c.repo))
+	if err != nil {
+		return err
+	}
+	err = os.Chdir(c.repo)
+	return err
+}
+
+// checkoutToPullRequest applies changes from the pull request to the working tree
+// of the branch that is being compared.
+func (c *gitClient) checkoutPR(num int) error {
+	_, _, err := execCommand("git", "fetch", "origin", fmt.Sprintf("pull/%d/head:pullrequest", num))
+	if err != nil {
+		return err
+	}
+	_, _, err = execCommand("git", "checkout", c.branch)
+	if err != nil {
+		return err
+	}
+	_, exitCode, err := execCommand("git", "merge", "--squash", "--no-commit", "pullrequest")
+	if err != nil || exitCode != 0 {
+		return errors.Wrap(err, "Pull request merge failed.")
+	}
+	_, _, err = execCommand("git", "reset")
+	return err
+}
+
+func (c *gitClient) revertPRChanges() error {
+	return filepath.Walk(os.Getenv("GITHUB_WORKSPACE"), func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if info.IsDir() || strings.Contains(path, ".git/") || strings.HasSuffix(info.Name(), "test.go") {
+			return nil
+		}
+
+		data, _, err := execCommand("git", "checkout", "--", path)
+		if err != nil {
+			log.Println("Following error occured during checkout: ", data, "If the error is caused by an untracted file, it will be deleted.")
+
+			// New file appeared and is not tracked, then it can be removed.
+			if strings.Contains(data, "did not match any file") {
+				if err := os.Remove(path); err != nil {
+					return errors.Errorf("Error: %v; Command out: %s", err, string(data))
+				}
+				return nil
+			}
+			return err
+		}
+		return nil
+	})
+}
+
+func newBenchmarkTester() (*benchmarkTester, error) {
+	if err := os.Setenv("GO111MODULE", "on"); err != nil {
+		return nil, err
+	}
+
+	benchRegex, err := ioutil.ReadFile("/github/home/commentMonitor/REGEX")
+	if err != nil {
+		return nil, err
+	}
+	raceArgument, err := ioutil.ReadFile("/github/home/commentMonitor/RACE")
+	if err != nil {
+		return nil, err
+	}
+
+	bench := benchmarkTester{
+		raceFlagEnabled: string(raceArgument) != "-no-race",
+		home:            os.Getenv("HOME"),
+		bechRegex:       string(benchRegex),
+	}
+
+	_, _, err = execCommand("go", "get", "golang.org/x/tools/cmd/benchcmp")
+
+	log.Printf("Benchmark Tester: %+v\n", bench)
+	return &bench, err
+}
+
+func (bench *benchmarkTester) compareBenchmarks(old, new string) (string, error) {
+	out, _, err := execCommand(strings.Join([]string{os.Getenv("GOPATH"), "/bin/benchcmp"}, ""), old, new)
+	log.Println("Benchmark comparision output: ", out)
+
+	if strings.Count(out, "\n") < 2 {
+		return out, errors.New("error: `go test` did not match any `BenchmarkXxx` functions")
+	}
+	return formatComment(out), err
+}
+
+func (bench *benchmarkTester) execBenchmark() (string, error) {
+	var (
+		out      string
+		exitCode int
+		err      error
+	)
+	if bench.raceFlagEnabled {
+		out, exitCode, err = execCommand("go", "test", "-bench", fmt.Sprintf("^%s$", bench.bechRegex), "-benchmem", "-v", "./...")
+	} else {
+		out, exitCode, err = execCommand("go", "test", "-bench", fmt.Sprintf("^%s$", bench.bechRegex), "-benchmem", "-race", "-v", "./...")
+	}
+	log.Println("Executing benchmark with REGEX ", bench.bechRegex,
+		"Benchmark output: ", out)
+	if err != nil || exitCode != 0 {
+		return "", errors.Wrap(err, "Benchmark ended with an error.")
+	}
+
+	tempFile, err := ioutil.TempFile("", "benchmark")
+	if err != nil {
+		return "", err
+	}
+
+	if _, err := tempFile.Write([]byte(out)); err != nil {
+		return "", err
+	}
+	err = tempFile.Close()
+	return tempFile.Name(), err
+}
+
+func newClient(token string) *github.Client {
+	ts := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: token})
+	tc := oauth2.NewClient(context.Background(), ts)
+	clt := github.NewClient(tc)
+	return clt
+}
+
+func formatComment(rawTable string) string {
+	tableContent := strings.Split(rawTable, "\n")
+	for i := 0; i <= len(tableContent)-1; i++ {
+		e := tableContent[i]
+		switch {
+		case e == "":
+
+		case strings.Contains(e, "old ns/op"):
+			e = "| Benchmark | Old ns/op | New ns/op | Delta |"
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -1,124 +1,10 @@
-# Automated Prometheus E2E testing and benchmarking.
+## Intro
+The repository provides two main functions:
```
krasi-georgiev:
```suggestion
The repository provides two main projects:
```

README.md
```diff
@@ -1,124 +1,10 @@
...
+- fucnbench
+- prombench
 
-![Prombench Design](design.svg)
+## Prombench
+Benchmarking of the whole Prometheus stack and comparison with the Prometheus master branch. The benchmark is initiated by commenting on a pull request in the Prometheus repository. See Prombench [README.md](/cmd/prombench/README.md) for more information.
```
krasi-georgiev:
```suggestion
Benchmarks and compares 2 different Prometheus versions. The benchmark is initiated by a comment on a pull request in the repository. See Prombench [README.md](/cmd/prombench/README.md) for more information.
```

README.md
```diff
@@ -1,124 +1,10 @@
...
-    -f manifests/prombench/benchmark
-```
-
-## Buliding from source
-To build Prombench and related tools from source you need to have a working Go environment with go modules enabled. Prombench uses [promu](https://github.com/prometheus/promu) to build the binaries.
-```
-make build
-```
+## Funcbench
+Benchmarking of the time series database that Prometheus uses by default. The benchmark is initiated by commenting on a pull request in the Prometheus/TSDB repository. See Funcbench [README.md](/cmd/funcbench/README.md) for more information.
```
krasi-georgiev: We can benchmark any bench functions right? not just the tsdb functions.
```suggestion
Benchmarks and compares a Golang function from 2 different repository branches using the built in Golang benchmarking tooling. The benchmark is initiated by a comment on a pull request in the repository. See Funcbench [README.md](/cmd/funcbench/README.md) for more information.
```

cmd/funcbench/go.mod
```diff
@@ -0,0 +1,13 @@
+module funcbench
```
krasi-georgiev: unfortunately we can't have separate go.mod files for each tool as it messes with testing building etc.

run go build cmd/funcbench/main.go from the repo root and it will add the dependencies to the root go.mod file.

And after that need to remoe go.mod and go.sum from cmd/funcbench

cmd/funcbench/main.go
```diff
@@ -0,0 +1,261 @@
...
+	app.HelpFlag.Short('h')
+	input := app.Flag("input", "path to event.json").Short('i').Default("/github/workflow/event.json").String()
+	kingpin.MustParse(app.Parse(os.Args[1:]))
+
+	data, err := ioutil.ReadFile(*input)
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	// Temporary fix for the new Github actions time format. This makes the time stamps unusable.
```
krasi-georgiev:
```suggestion
	// Temporary fix for the new Github actions time format. This makes the time stamps json parsable.
```

cmd/funcbench/main.go
```diff
@@ -0,0 +1,261 @@
...
+	txt := string(data)
+	reg := regexp.MustCompile("(.*)\"[0-9]+/[0-9]+/2019 [0-9]+:[0-9]+:[0-9]+ [AP]M(.*)")
+	txt = reg.ReplaceAllString(txt, "$1\"2019-06-11T09:26:28Z$2")
+	data = []byte(txt)
+	// End of the temporary fix
```
krasi-georgiev:
```suggestion
	// End of the temporary fix.
```

cmd/funcbench/main.go
```diff
@@ -0,0 +1,261 @@
...
+		os.Chdir(os.Getenv("GITHUB_WORKSPACE"))
+		data, err := exec.Command("git", "clone", fmt.Sprintf("https://github.com/%s/%s.git", owner, repo)).CombinedOutput()
+		log.Println(string(data))
+		if err != nil {
+			log.Fatalln(err)
+		}
+		os.Chdir(repo)
+
+		latestCommitHash := os.Getenv("GITHUB_SHA")
+		log.Println(latestCommitHash)
```
krasi-georgiev: debug log?

cmd/funcbench/main.go
```diff
@@ -0,0 +1,261 @@
...
+		}
+		err = ioutil.WriteFile(oldPath, data, 0644)
+		if err != nil {
+			log.Fatalln(err)
+		}
+
+		data, err = exec.Command(strings.Join([]string{os.Getenv("GOPATH"), "/bin/benchcmp"}, ""), oldPath, newPath).CombinedOutput()
+		if err != nil {
+			log.Println(string(data))
+			if err := postComment(clt, owner, repo, prNumber, fmt.Sprintf("Error: `benchcmp` failed. Check https://github.com/%s/%s/commit/%s/checks for more information.", owner, repo, latestCommitHash)); err != nil {
```
krasi-georgiev: Best to put this in some var instead of repeating it so many times
```
logsLink := fmt.Sprintf("https://github.com/%s/%s/commit/%s/checks for full logs.", owner, repo, latestCommitHash)
```

cmd/prombench/README-prombench.md
```diff
@@ -0,0 +1,124 @@
+# Automated Prometheus E2E testing and benchmarking.
```
krasi-georgiev: why is it called README-prombench.md and not just README.md?

Also for git to detect and show a file as moved you should use
git mv README.md cmd/prombench/README.md

or `git add -A .` , this will detect when a file was move.

This makes reviews a lot easyer, knowing that the file was not modified and just moved, I don't have to read it to see if there any changes there.

krasi-georgiev (commented):
No DCO - see https://github.com/prometheus/prometheus/wiki/DCO-signing

file conflict, need to git pull the latest commits from master.

failing tests. need to add the license header like in the other files

krasi-georgiev:
just remembered that also need it it to the promu config so it gets build as part of the CI tests.
https://github.com/prometheus/prombench/blob/master/.promu.yml

README.md
VladMasarik: I don't understand the meaning of the changed sentence.

krasi-georgiev: I updated it a bit, is this more clear?

krasi-georgiev: I updated it again to simplify, no need to include more details as these should be mentioned in the extended readme. It will avoid having to modify if the design changes at some point.

README.md
VladMasarik: True

cmd/funcbench/main.go
VladMasarik: yes, it makes them parsable, but it also replaces them with a constant, which means that using them is useless. So I am not sure how to deal with this situation.

krasi-georgiev: If you are not using it than can merge as is and refactor once they answer your support ticket.

cmd/prombench/README-prombench.md
VladMasarik: The `README-prombench.md` file contains some text, and it was previously in the prombench folder as the default `README.md`. I am not sure who created it, but before deleting I wanted to ask whether it should be kept or deleted. So, what should we do about it?

Thanks for the suggestion, I will use it next time.

krasi-georgiev: > Thanks for the suggestion, I will use it next time.

you need to use for this time :smile: 

Otherwise I can't read the PR properly.

> The README-prombench.md file contains some text, and it was previously in the prombench folder as the default README.md. I am not sure who created it, but before deleting I wanted to ask whether it should be kept or deleted. So, what should we do about it?

combine the two and just name it as README.md

VladMasarik: How can I move it since it is already moved? Do I have to move it back, remove the current `README.md`, commit the changes, move it to `prombench` folder again, add the new `README.md` again, and commit? Would that work?

cmd/funcbench/main.go
```diff
@@ -175,6 +175,16 @@ func main() {
 			data, err = exec.Command("git", "checkout", "--", path).CombinedOutput()
 			if err != nil {
 				log.Println(string(data))
+
+				// New file appeared and is not tracked, then it can be removed
+				if strings.Contains(string(data), "did not match any file") {
+					data, err = exec.Command("rm", path).CombinedOutput()
+					if err != nil {
+						log.Println(string(data))
+						log.Fatalln(err)
```
krasi-georgiev: `Fatalln` and `Println` can take more than a single argument like
`log.Println("err:", err,"data:", string(data))`

cmd/prombench/README-prombench.md
krasi-georgiev: Yes need to do it from the begining.

krasi-georgiev:
also need to move diagram.svg and diagram.xml to cmd/prombench

funcbench/main.go
```diff
@@ -0,0 +1,281 @@
...
+	switch e := event.(type) {
+	case *github.IssueCommentEvent:
+
+		//Variable setup.
+		branch := readFile("/github/home/ARG_0")
+		if branch == "" {
+			branch = "master"
+		}
+		issueComment := readFile("/github/home/ARG_1")
+		race := readFile("/github/home/ARG_2")
```
krasi-georgiev: I think you can have this as 

```suggestion
		race := readFile("/github/home/ARG_2") != "-no-race"
```

this way race will be a boolean

funcbench/main.go
```diff
@@ -0,0 +1,281 @@
...
+			}
+			if info.IsDir() || strings.Contains(path, ".git/") || strings.HasSuffix(info.Name(), "test.go") {
+				return nil
+			}
+
+			data, err = exec.Command("git", "checkout", "--", path).CombinedOutput()
+			if err != nil {
+				log.Println(string(data))
+
+				// New file appeared and is not tracked, then it can be removed
```
krasi-georgiev:
```suggestion
				// New file appeared and is not tracked, then it can be removed.
```

funcbench/main.go
```diff
@@ -0,0 +1,281 @@
...
+				if strings.Contains(string(data), "did not match any file") {
+					data, err = exec.Command("rm", path).CombinedOutput()
```
krasi-georgiev: When would such file appear and why do you have to remove it?

funcbench/main.go
```diff
@@ -0,0 +1,281 @@
...
+		// Benchmark the comparing branch.
+		if race == "-no-race" {
+			cmd = exec.Command("go", "test", "-bench", fmt.Sprintf("^%s$", issueComment), "-benchmem", "-v", "./...")
+		} else {
+			cmd = exec.Command("go", "test", "-bench", fmt.Sprintf("^%s$", issueComment), "-benchmem", "-race", "-v", "./...")
+		}
+		data, err = cmd.CombinedOutput()
+		log.Println(string(data))
+		if err != nil || cmd.ProcessState.ExitCode() != 0 {
+			if err := postComment(clt, owner, repo, prNumber, fmt.Sprintf("Go test on branch %s failed. %s", branch, logLink)); err != nil {
```
krasi-georgiev:
```suggestion
			if err := postComment(clt, owner, repo, prNumber, fmt.Sprintf("Go bench test on branch %s failed. %s", branch, logLink)); err != nil {
```

funcbench/main.go
```diff
@@ -0,0 +1,281 @@
...
+		}
+
+		// Compare the benchmarks.
+		data, err = exec.Command(strings.Join([]string{os.Getenv("GOPATH"), "/bin/benchcmp"}, ""), oldPath, newPath).CombinedOutput()
+		if err != nil {
+			log.Println(string(data))
+			if err := postComment(clt, owner, repo, prNumber, fmt.Sprintf("Error: `benchcmp` failed. %s", logLink)); err != nil {
+				log.Fatalln(err)
+			}
+			log.Fatalln(err)
```
krasi-georgiev: you can simplify this by just calling log.Fatal once.

funcbench/main.go
```diff
@@ -0,0 +1,281 @@
...
+		}
+		if strings.Count(string(data), "\n") < 2 {
+			if err := postComment(clt, owner, repo, prNumber, fmt.Sprintf("Error: `go test` did not match any benchmarked functions. %s", logLink)); err != nil {
```
krasi-georgiev:
```suggestion
			if err := postComment(clt, owner, repo, prNumber, fmt.Sprintf("Error: `go test` did not match any `BenchmarkXxx` functions. %s", logLink)); err != nil {
```

funcbench/main.go
```diff
@@ -0,0 +1,281 @@
...
+			}
+			tableContent[i] = e
+		}
+
+		if err := postComment(clt, owner, repo, prNumber, strings.Join(tableContent, "\n")); err != nil {
+			log.Fatalln(err)
+		}
+
+	default:
+		log.Fatalln("simpleargs only supports issue_comment event")
```
krasi-georgiev:
```suggestion
		log.Fatalln("only issue_comment event is supported")
```

funcbench/README.md
```diff
@@ -0,0 +1,52 @@
+## Intro
```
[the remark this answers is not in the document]
VladMasarik: I added every new idea from your suggestion.

"How to build image" seciton is 16 lines under this comment.

funcbench/main.go
VladMasarik: When I started testing Prometheus, it created a file called `lock`. When the program executed `git checkout -- lock`, an error occurred because `lock` is not tracked, and therefore, cannot be checked out. This lead to a failure.

funcbench/main.go
VladMasarik: Wouldn't the error be overwritten by the `postComment` function?

funcbench/main.go
krasi-georgiev: ok thanks.

funcbench/main.go
krasi-georgiev: yeah it is a bit confusing as now if `postComment` fails you will never see the `exec.Command` error as  `log.Fatalln(err)` exits the program immediately

I would say there is too much logic in the main func. Lets brake each step into smaller funcs and  do some proper error handling and wrapping. `errors.Wrap`

funcbench/README.md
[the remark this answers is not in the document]
krasi-georgiev: In a readme I would expect to know what the tool does by just reading the first 1-2 sentences. In the current readme it is not quite clear until the second paragraph.

In the format I suggested it is quite clear what is the workflow and makes it more clear that it expects some arguments that are passed by another action.

funcbench/README.md
```diff
@@ -0,0 +1,52 @@
...
+- Add comment-monitor Github action as a first step.
+- Specify this regex `^/benchmark ?([^ B\.]+)? ?(\.|Bench.*|[^ ]+)? ?(-no-race)?.*$` in the `args` field of the comment-monitor.
+- Specify this Github action as a pre-built image, build from this source code, or just refer to this repository from the workflow file.
+- Provide a Github token as an environment variable to both comment-monitor and funcbench.
+
+## How to build
+`docker build -t <tag of your choice> .`
+
+## Examples
+### Example comments
```
krasi-georgiev: An example how to run a bench test by an exact func name would also be useful.

funcbench/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Use
+After a successful set up, you can create a comment in a pull request to execute the benchmarks. The form of the comment is following:
+`/benchmark [branch] <golang test regex>  [-no-race]`
+If no branch is specified, `master` branch is used. The golang test regex is used to execute benchmarks that match the regex. By defualt the benchmarks run with `-race` flag enabled, disable this by specifying `-no-race` in the comment. After a successful run, a table with results will be displayed in the comment section.
+
+## Set up
+- Create Github actions workflow file that is executed when an issue comment is created, `on = "issue_comment"`.
+- Add comment-monitor Github action as a first step.
+- Specify this regex `^/benchmark ?([^ B\.]+)? ?(\.|Bench.*|[^ ]+)? ?(-no-race)?.*$` in the `args` field of the comment-monitor.
```
```suggestion
- Specify this regex `^/funcbench ?([^ B\.]+)? ?(\.|Bench.*|[^ ]+)? ?(-no-race)?.*$` in the `args` field of the comment-monitor.
```

funcbench/README.md
```diff
@@ -0,0 +1,52 @@
...
+- Specify this Github action as a pre-built image, build from this source code, or just refer to this repository from the workflow file.
```
krasi-georgiev: It would be much more clear if you just show the example first with comments for what each env var is for.

funcbench/README.md
```diff
@@ -0,0 +1,52 @@
...
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        args: '"^/benchmark ?([^ ]+)? ?([^ ]+)? ?(-no-race)?.*$"'
+    - name: benchmark
+      uses: docker://prombench/funcbench:latest
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+```
+
+## What it does [SHORT]
```
krasi-georgiev: As I mentioned lets put the overview at the top.

funcbench/README.md
```diff
@@ -0,0 +1,52 @@
...
+It acquires arguments for `go test` benchmarks, executes and compares the benchmarks based on the provided arguments. Then it posts results into pull request it was called from.
+
+## What it does [LONG]
```
krasi-georgiev: It is only an additional sentence, why not combine it in a single paragraph?

funcbench/main.go
VladMasarik: This one is copied from the comment monitor, should I fix it there as well?

krasi-georgiev: @geekodour 	is updating it now so can do it as part of his PR

funcbench/README.md
VladMasarik: I wanted to add more, but we switched to another task and I forgot what I wanted to say.

funcbench/README.md
VladMasarik: I hoped it would evident from the fact that we use regex to match the functions.

krasi-georgiev: 3 examples:
Bench a func by an exact match
Bench all functions with the same suffix
Bench all functions from all modules.

funcbench/main.go
```diff
@@ -0,0 +1,318 @@
...
+	"github.com/google/go-github/github"
+	"github.com/pkg/errors"
+	"golang.org/x/oauth2"
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+type githubClient struct {
+	event                                          *github.IssueCommentEvent
+	client                                         *github.Client
+	owner, repo, branch, latestCommitHash, logLink string
```
krasi-georgiev: you can put these in the lines to avoid this long wrapping

funcbench/main.go
```diff
@@ -0,0 +1,318 @@
...
+	}
+
+	switch eventType := event.(type) {
+	case *github.IssueCommentEvent:
+		ghClient := githubClient{event: eventType}
+		if err := ghClient.initialize(); err != nil {
+			log.Fatalln(err)
+		}
+		benchExec := benchmarkHandler{}
+		if err := benchExec.initialize(); err != nil {
```
krasi-georgiev: you can have a constructor here like:

```suggestion
		type benchTest struct {.... or type benchGroup struct {

		if bench,err := newBencTest();
		if  err != nil {
			log.Fatalln(err)
		}
```

than you can have
```
bench.StartNew()
bench.StartOld()
```

funcbench/main.go
```diff
@@ -0,0 +1,318 @@
...
+	// End of the temporary fix.
+
+	event, err := github.ParseWebHook("issue_comment", data)
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	switch eventType := event.(type) {
+	case *github.IssueCommentEvent:
+		ghClient := githubClient{event: eventType}
```
krasi-georgiev: How about an api that is something like:

```
type gitClient
gitClt,err := newGitClient() // All git setup goes in here - name, email and anyting git related
gitClt.ClonePR
gitClt.CloneBranch

type githubClient
ghClt,err := newGithubClient()
ghClt.PostComment


type benchTest struct {.... 

bench,err := newBencTest(gitClient);
if  err != nil {
	log.Fatalln(err)
}

bench.StartPR(pr)
bench.StartBranch(branch)
bench.Results()
```

krasi-georgiev (commented):
It is an improvement, but we can do better.

funcbench/main.go
VladMasarik: What are you expecting to be in a function called `gitClt.ClonePR`. If, I assume right, that it is a function, as you have put brackets in only certain parts of your example.

Does your example hint that I should create 3 clients? Also, to what degree is you example a pseudo code? Do you imagine that the main function would contain functions named as in your example and all I would add is error checking?

funcbench/main.go
```diff
@@ -0,0 +1,362 @@
...
+	repo := *event.GetRepo().Name
+	latestCommitHash := os.Getenv("GITHUB_SHA")
+
+	return &gitHubClient{
+		client:           newClient(os.Getenv("GITHUB_TOKEN")),
+		owner:            owner,
+		repo:             repo,
+		prNumber:         *event.GetIssue().Number,
+		latestCommitHash: latestCommitHash,
+		logLink:          fmt.Sprintf("Full logs at: https://github.com/%s/%s/commit/%s/checks", owner, repo, latestCommitHash),
```
krasi-georgiev: don't think this should be part of the github client, can be just a variable in the main function

funcbench/main.go
```diff
@@ -0,0 +1,362 @@
...
+	_, err = execCommand("git", "reset")
+	return err
+}
+
+func (c *gitClient) fetchPullRequest() error {
+	_, err := execCommand("git", "fetch", "origin", fmt.Sprintf("pull/%d/head:pullrequest", c.prNumber))
+	return err
+}
+
+func (c *gitClient) checkoutToComparisonBranch() error {
```
krasi-georgiev: something like this or `resetPRChanges` describes better of what it actually does right?
```suggestion
func (c *gitClient) revertPRChanges() error {
```

also I don't think you need e separate func `checkoutPullRequestFile` here. can just directly put the code there as logically it does what the func name suggest so no need to separate.


btw I found this which explains very well when it is good to put some code into a new func.
https://dave.cheney.net/practical-go/presentations/qcon-china.html#_rather_than_commenting_a_block_of_code_refactor_it

funcbench/main.go
```diff
@@ -0,0 +1,362 @@
...
+	}
+	return nil
+}
+
+func (c *gitClient) cloneRepository() error {
+	_, err := execCommand("git", "clone", fmt.Sprintf("https://github.com/%s/%s.git", c.owner, c.repo))
+	return err
+}
+
+func (c *gitClient) checkoutToPullRequest() error {
```
krasi-georgiev: can you put a comment for this function what it does exactly as I don't quite understand it and it will be a good reminder for when we read it in few weeks time.

btw I think we should have keep the pr var in the main function and just pass it here as an argument. When I see `checkoutToPullRequest` I already ask which pr do you checkout to?
I think it is much logical if you have something like

```suggestion
func (c *gitClient) checkoutPR(num int) error {
```

funcbench/main.go
```diff
@@ -0,0 +1,362 @@
...
+	}
+	cmd, err := execCommand("git", "merge", "--squash", "--no-commit", "pullrequest")
+	if err != nil || cmd.ProcessState.ExitCode() != 0 {
+		return errors.Wrap(err, "Pull request merge failed.")
+	}
+	_, err = execCommand("git", "reset")
+	return err
+}
+
+func (c *gitClient) fetchPullRequest() error {
```
krasi-georgiev: don't think we need a separate func for this as the usual git workflow when you checkout ot a branch is to fetch it first so I think this should be just part of `checkoutToPullRequest`

funcbench/main.go
```diff
@@ -0,0 +1,356 @@
...
+
+}
+
+func readFile(name string) (string, error) {
+	data, err := ioutil.ReadFile(name)
+	return string(data), err
+
+}
+
+func execCommand(command ...string) (*exec.Cmd, error) {
```
krasi-georgiev: why not refactor it to return the output and the exit code and use it everywhere?

funcbench/main.go
```diff
@@ -0,0 +1,356 @@
...
+	raceArgument, err := readFile("/github/home/commentMonitor/RACE")
+	home := os.Getenv("HOME")
+
+	bench := benchmarkTester{
+		raceFlagEnabled:          raceArgument != "-no-race",
+		home:                     home,
+		bechRegex:                bechRegex,
+		comparisonBranchFilePath: strings.Join([]string{home, "/old.txt"}, ""),
+		pullRequestFilePath:      strings.Join([]string{home, "/new.txt"}, ""),
+	}
```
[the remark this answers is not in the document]
VladMasarik: Do I have to use `exec` for it? If I am going to remove it from Dockerfile, why not just download it right away and not check whether there is `benchcmp` or not.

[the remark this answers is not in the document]
krasi-georgiev: > Do I have to use exec for it?

yep, don't think there is another way

>  If I am going to remove it from Dockerfile, why not just download it right away and not check whether there is benchcmp or not.

yep that is a good idea

funcbench/main.go
```diff
@@ -0,0 +1,347 @@
...
+		out      string
+		exitCode int
+		err      error
+	)
+	if bench.raceFlagEnabled {
+		out, exitCode, err = execCommand("go", "test", "-bench", fmt.Sprintf("^%s$", bench.bechRegex), "-benchmem", "-v", "./...")
+	} else {
+		out, exitCode, err = execCommand("go", "test", "-bench", fmt.Sprintf("^%s$", bench.bechRegex), "-benchmem", "-race", "-v", "./...")
+	}
+	log.Println(out)
```
krasi-georgiev: Add some more context to the logs.

funcbench/main.go
```diff
@@ -0,0 +1,347 @@
...
+}
+
+func (bench *benchmarkTester) compareBenchmarks(old, new string) (string, error) {
+	out, _, err := execCommand(strings.Join([]string{os.Getenv("GOPATH"), "/bin/benchcmp"}, ""), old, new)
+	log.Println(out)
+
+	if strings.Count(out, "\n") < 2 {
+		return out, errors.New("error: `go test` did not match any `BenchmarkXxx` functions")
+	}
+	return formatFinalComment(out), err
```
krasi-georgiev:
```suggestion
	return formatComment(out), err
```

funcbench/README.md
```diff
@@ -0,0 +1,53 @@
+## Intro
+A tool used as a github action to run a `go test -bench` and compare changes from a PR against another branch. The benchmark is triggered by creating a comment which specifies a branch to compare. The results are then posted back as a PR comment. 
+Comparison is done using [benchcmp](https://godoc.org/golang.org/x/tools/cmd/benchcmp). Arguments for the benchcmp are read from files created by previous action (for example [commentMonitor](/tools/commentMonitor/main.go)), which is responsible for the comment parsing.
+
+## Use
+Tests are triggered by posting a comment in a PR with the following format:
+`/funcbench <branch> <golang test regex>  [-no-race]`
```
krasi-georgiev, line 7: use ``` and a new line to make it on a new line

funcbench/README.md
```diff
@@ -0,0 +1,53 @@
...
+Specifying which tests to run are filtered by using the standard golang regex format. 
+By default all benchmarks run with `-race` flag enabled and it can be disabled by appending `-no-race` at the end of the comment. 
+After a successful run a github comment is created with the results.
```
```suggestion
```

this is already mentioned in the first paragraph

funcbench/main.go
```diff
@@ -0,0 +1,354 @@
...
+
+	if err := gitClient.checkoutPR(prNumber); err != nil {
+		comment := fmt.Sprintf("Switch to a pull request branch failed. %s", logLink)
+		if postCommentErr := ghClient.postComment(comment); postCommentErr != nil {
+			log.Fatalf("Error posting a comment for `checkoutToPullRequest` command execution error. checkoutToPullRequest err:%v, postComment err:%v", err, postCommentErr)
+		}
+		log.Fatalln(err)
+	}
+
+	newBenchmark, err := benchTest.execBenchmark()
```
krasi-georgiev:
```suggestion
	prResults, err := benchTest.execBenchmark()
```

funcbench/main.go
```diff
@@ -0,0 +1,354 @@
...
+			log.Fatalf("An error: %v occured while processing error: %v", postCommentErr, err)
+		}
+		log.Fatalln(err)
+	}
+
+	if err := gitClient.revertPRChanges(); err != nil {
+		log.Fatalln(err)
+	}
+
+	oldBenchmark, err := benchTest.execBenchmark()
```
krasi-georgiev:
```suggestion
	branchResults, err := benchTest.execBenchmark()
```

funcbench/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Example Github actions workflow file
```
krasi-georgiev, line 11: not sure if we should keep this example file or just put a link to the one that will be in the prometheus repo.
Having it here means that when we change things need to also update the example where if we just put a link to the prometheus repo it will always be up to date.

funcbench/main.go
```diff
@@ -0,0 +1,354 @@
...
+
+	_, _, err = execCommand("go", "get", "golang.org/x/tools/cmd/benchcmp")
+
+	log.Printf("Benchmark Tester: %+v\n", bench)
+	return &bench, err
+}
+
+func (bench *benchmarkTester) compareBenchmarks(old, new string) (string, error) {
+	out, _, err := execCommand(strings.Join([]string{os.Getenv("GOPATH"), "/bin/benchcmp"}, ""), old, new)
+	log.Println(out)
```
krasi-georgiev: please add more context - I mean something like:
```suggestion
	log.Println("benchcmp result:", out)
```

the same for the other log that is missing context.

funcbench/main.go
```diff
@@ -0,0 +1,354 @@
...
+		bechRegex:       string(benchRegex),
+	}
+
+	_, _, err = execCommand("go", "get", "golang.org/x/tools/cmd/benchcmp")
+
+	log.Printf("Benchmark Tester: %+v\n", bench)
+	return &bench, err
+}
+
+func (bench *benchmarkTester) compareBenchmarks(old, new string) (string, error) {
```
krasi-georgiev, line 259:
```suggestion
func (bench *benchmarkTester) compareBenchmarks(file1, file2 string) (string, error) {
```

in the context of the function there is no new and old, just compares 2 files.

funcbench/main.go
```diff
@@ -0,0 +1,354 @@
...
+			return nil
+		}
+
+		data, _, err := execCommand("git", "checkout", "--", path)
+		if err != nil {
+			log.Println(data)
+
+			// New file appeared and is not tracked, then it can be removed.
+			if strings.Contains(data, "did not match any file") {
+				data, _, err = execCommand("rm", path)
```
krasi-georgiev: I think you can use https://golang.org/pkg/os/#Remove

funcbench/main.go
krasi-georgiev: or maybe call it before and after

just old and new sounds a bit strange. Not a blocker and can leave it as is if you prefer.

funcbench/main.go
```diff
@@ -0,0 +1,354 @@
...
+		bechRegex:       string(benchRegex),
+	}
+
+	_, _, err = execCommand("go", "get", "golang.org/x/tools/cmd/benchcmp")
+
+	log.Printf("Benchmark Tester: %+v\n", bench)
+	return &bench, err
+}
+
+func (bench *benchmarkTester) compareBenchmarks(file1, file2 string) (string, error) {
```
krasi-georgiev: Hm, I think I was wrong here, the context here is still old and new as in the examples
https://godoc.org/golang.org/x/tools/cmd/benchcmp

sorry my bad.

funcbench/README.md
VladMasarik: The example, is just an example. As long as it is a valid example I think it is okay, and it does not have to be directing to something useful. I think we should modify the example only if we modify the `funcbench`. Also, it might be possible that the link will be broken and we might never find out. So I would rather keep it as in the README as a text.

krasi-georgiev: ok

PR #258, opened 2019-08-16
Filenames same as env var

### LINKED ISSUE #250: Enable named files for arguments in commentMonitor
Currently when using commentMonitor we're getting the `PR_NUMBER` as such:
`export PR_NUMBER=$(cat /github/home/ARG_1);,`

I think it will be better to have the filename of known args as their env var names, think this can be implemented in `writeArgs` by checking the value of `data` to a  list of known regex for the envvars and when it's not in the list of known regex we can add it the name `ARG_X` and so on.
https://github.com/prometheus/prombench/blob/156828803d0c72fbe024dccd7804a702ec1cf23f/cmd/tools/commentMonitor/main.go#L133

So that we can do something like:
`export PR_NUMBER=$(cat /github/home/PR_NUMBER);,`

nikita-mk:
Closes #250 
Also added the temporary fix for the new Github actions time format.

### DIFF
--- a/tools/commentMonitor/main.go
+++ b/tools/commentMonitor/main.go
@@ -152,9 +152,9 @@ func writeArgs(args []string, output string) {
 	// More regex can be added.
 	regex := make(map[string]*regexp.Regexp)
 	// Regex for PR number.
-	regex["PR_NUMBER"] = regexp.MustCompile("(?m)^([0-9]+)\\s*$")
+	regex["PR_NUMBER"] = regexp.MustCompile(`(?m)^([0-9]+)\s*$`)
 	// Regex for Release version.
-	regex["RELEASE"] = regexp.MustCompile("(?mi)^(master|v[0-9]+\\.[0-9]+\\.[0-9]+\\S*)\\s*$")
+	regex["RELEASE"] = regexp.MustCompile(`(?mi)^(master|v[0-9]+\.[0-9]+\.[0-9]+\S*)\s*$`)
 
 	var filename string
 	for i, arg := range args[1:] {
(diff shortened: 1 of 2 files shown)


nikita-mk:
cc @krasi-georgiev @geekodour @VladMasarik

tools/commentMonitor/main.go
```diff
@@ -141,9 +149,24 @@ func regexValidation(regex string, comment string) ([]string, error) {
 }
 
 func writeArgs(args []string, output string) {
+	// More regex can be added.
+	regex := make(map[string]*regexp.Regexp)
+	// Regex for PR number.
+	regex["PR_NUMBER"] = regexp.MustCompile("(?m)^([0-9]+)\\s*$")
+	// Regex for Release version.
+	regex["RELEASE"] = regexp.MustCompile("(?mi)^(master|[0-9]+\\.[0-9]+\\.[0-9]+\\S*)\\s*$")
```
geekodour: the `v` prefix was made mandatory when commenting to avoid confusion in the future.
https://github.com/prometheus/prombench#trigger-tests-via-a-github-comment

I think this regex will need the little bit of modification for that, also do we need double backslash when using it inside a gofile? `\\` I think this should work, not 100% sure though:
```
(?mi)^(master|v[0-9]+\.[0-9]+\.[0-9]+\S*)\s*$
```

nikita-mk: Build fails if `\` used instead of `\\`

nikita-mk: Worked after using \` instead of `"`

krasi-georgiev:
closed in https://github.com/prometheus/prombench/pull/262

PR #259, opened 2019-08-16
Add support for creating configmap directly from file

geekodour:
example
```
./prombench gke configmap apply -a $AUTH_FILE --name=grafana-provision --namespace=default -f dashboard.json -f datasource.json -v PROJECT_ID:$PROJECT_ID -v ZONE:$ZONE -v CLUSTER_NAME:$CLUSTER_NAME
```


krasi-georgiev:
why do we need this? Can't we use the existing deployment commands?

geekodour:
This is an enhancement and can help with creating configmaps with contents which have conflicting syntax with go templates.

Actual format
```
"legendFormat": "{{ node }} - Cached", 
```
v/s the escaping that we're using currently:
https://github.com/prometheus/prombench/blob/ce7de2f047789896e1715826d3d8a051b786d8ec/manifests/cluster-infra/grafana_dashboard_prow_bench.yaml#L120

geekodour:
This does not support using `DeploymentVars` though. mostly can be used for static config files.

krasi-georgiev:
lets revisit as if we add to many ways to do same thing it will get confusing very quickly.

geekodour:
okay, just for a reminder for later:

This will help in updating the configmap directly when the file is updated using `on:push` on `branch:master` of `prometheus/prombench` (prombench itself will use github actions for this)

secondly, it will be good but not mandatory to have these config files in a separate `configs` directory.

PR #260, opened 2019-08-16
Check for ReadyReplicas instead when waiting for StatefulSet

geekodour:
Wait till pod becomes ready in the prombench cli


krasi-georgiev:
Should we do the same for the other commands?

geekodour:
The `CurrentReplicas` is working as expected with `Deployment` so I decided to leave it as is.

PR #264, opened 2019-08-20
[RDY] Custom wait for prombench in StatefulSet

geekodour:
This PR attempts to add custom wait for complete deployment of a prombench test, so that github action waits till the error or completion in deploying a statefulset.

Todo:
- [x] ~Custom annotation: retry count~ (6xGlobalRetryCount for statefulSet)
- [x] ~Custom annotation: exit on error~(Made default behavior)
- [x] ~Custom annotation: wait on update~(Made default behaviour)
- [x] add readiness probe

### DIFF
--- a/pkg/provider/k8s/k8s.go
+++ b/pkg/provider/k8s/k8s.go
@@ -462,6 +462,23 @@ func (c *K8s) statefulSetApply(resource runtime.Object) error {
 		req.Namespace = "default"
 	}
 
+	// set annotations
+	var retryCount int
+
+	retryCount = provider.GlobalRetryCount
+	if count, ok := req.Annotations["prometheus.io/prombench.retry_count"]; ok {
+		intCount, err := strconv.Atoi(count)
+		if err != nil {
+			return fmt.Errorf("%v: format of .retry_count annotation wrong", err)
+		}
+		retryCount = intCount
+	}
+	if exitOnError, ok := req.Annotations["prometheus.io/prombench.exit_on_error"]; ok {
+		if exitOnError == "true" {
+			provider.ExitOnError = true
+		}
+	}
+
 	switch v := resource.GetObjectKind().GroupVersionKind().Version; v {
 	case "v1":
 		client := c.clt.AppsV1().StatefulSets(req.Namespace)
@@ -486,6 +503,14 @@ func (c *K8s) statefulSetApply(resource runtime.Object) error {
 				return errors.Wrapf(err, "resource update failed - kind: %v, name: %v", kind, req.Name)
 			}
 			log.Printf("resource updated - kind: %v, name: %v", kind, req.Name)
+			if waitOnUpdate, ok := req.Annotations["prometheus.io/prombench.wait_on_update"]; ok {
+				if waitOnUpdate == "true" {
+					return provider.RetryUntilTrue(
+						fmt.Sprintf("applying statefulSet:%v", req.Name),
+						retryCount,
+						func() (bool, error) { return c.statefulSetReady(resource) })
+				}
+			}
 			return nil
 		} else if _, err := client.Create(req); err != nil {
 			return errors.Wrapf(err, "resource creation failed - kind: %v, name: %v", kind, req.Name)
@@ -495,20 +520,6 @@ func (c *K8s) statefulSetApply(resource runtime.Object) error {
 		return fmt.Errorf("unknown object version: %v kind:'%v', name:'%v'", v, kind, req.Name)
 	}
 
-	if exitOnError, ok := req.Annotations["prometheus.io/prombench.exit_on_error"]; ok {
-		if exitOnError == "true" {
-			provider.ExitOnError = true
-		}
-	}
-
-	retryCount := provider.GlobalRetryCount
-	if count, ok := req.Annotations["prometheus.io/prombench.retry_count"]; ok {
-		intCount, err := strconv.Atoi(count)
-		if err != nil {
-			return fmt.Errorf("%v: format of .retry_count annotation wrong", err)
-		}
-		retryCount = intCount
-	}
 	return provider.RetryUntilTrue(
 		fmt.Sprintf("applying statefulSet:%v", req.Name),
 		retryCount,
--- a/pkg/provider/k8s/k8s.go
+++ b/pkg/provider/k8s/k8s.go
@@ -465,8 +465,8 @@ func (c *K8s) statefulSetApply(resource runtime.Object) error {
 	// set annotations
 	retryCount := provider.GlobalRetryCount
 	waitOnUpdate := false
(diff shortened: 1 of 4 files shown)


geekodour:
ping @krasi-georgiev

manifests/prombench/stateful-set.yaml
```diff
@@ -4,6 +4,10 @@ metadata:
   name: prombench-test-{{ .PR_NUMBER }}
   labels:
     app: prombench-test-{{ .PR_NUMBER }}
+  annotations:
+    prometheus.io/prombench.exit_on_error: 'true'
+    prometheus.io/prombench.wait_on_update: 'true'
```
krasi-georgiev: why not have these as cli arg to `resource apply` ? This way the cli help file will show this arg and it will be clear how to use without any need to look inside the code to know that such a functionality exists.

Preferable use it for all apply functions and name `waitOnUpdate` or  otherwise just name it `statefulSet.waitOnUpdate` or something similar to indicate that it is only for statefull set


btw don't we want this behavior in all cases? 
I mean we always want to make sure that the update is completed?

manifests/prombench/stateful-set.yaml
```diff
@@ -4,6 +4,10 @@ metadata:
   name: prombench-test-{{ .PR_NUMBER }}
   labels:
     app: prombench-test-{{ .PR_NUMBER }}
+  annotations:
+    prometheus.io/prombench.exit_on_error: 'true'
```
krasi-georgiev: don't we want this in all cases?

manifests/prombench/stateful-set.yaml
```diff
@@ -4,6 +4,10 @@ metadata:
...
+    prometheus.io/prombench.wait_on_update: 'true'
+    prometheus.io/prombench.retry_count: '90'
```
krasi-georgiev: again I think best to set this as a cli arg

manifests/prombench/stateful-set.yaml
geekodour: I thought so too, should the default behavior be changed then?

The only time I found NOT exiting on error useful was the time `/manifest/prombench/benchmark` is getting deployed and if some deployment times out, the other deployment after that continue to get deployed. (but this behavior is undesirable sometimes) let me know if we should change this.

If we change this then we also won't need the global var that you mentioned : `ExitOnError = false `

manifests/prombench/stateful-set.yaml
geekodour: waiting is currently only done on `namespace - delete`,  `Deplyment, StatefulSet, DaemonSet - create`, will add waiting on `Deplyment, StatefulSet, DaemonSet - update` too.

This will also get rid of the requirement of `prometheus.io/prombench.wait_on_update: 'true' ` annotation

manifests/prombench/stateful-set.yaml
geekodour: I think keeping `retry_count` as an annotation makes more sense because we might want different resources to have different retry count. having it as a cli arg will mean the same retry count will be enforced on all of the manifest files, similar to what `GlobalRetryCount` is currently doing

krasi-georgiev: ok in that case can add it as a cli arg - `-retryCount=60 -retryCountResourceName(or retryCountResourceFilename)=manifesFilename.yaml` or some similiar aproach.

my main concern here is that if we start setting options from different places it will get confusing.

prombench/manifests/prombench/stateful-set.yaml
```diff
@@ -43,7 +43,7 @@ spec:
         imagePullPolicy: Always
         args:
         # sleep forever after make deploy succeeded
-        - "make deploy && echo 'deploy succeeded, now sleeping' && while sleep 86400; do :; done"
+        - "make deploy && touch /tmp/READY && echo 'deploy succeeded, now sleeping' && while sleep 86400; do :; done"
```
krasi-georgiev, line 48: can you also comment the workflow with `/tmp/READY` why is it created and than why it is removed etc.

prombench/manifests/prombench/stateful-set.yaml
```diff
@@ -70,13 +70,13 @@ spec:
               - "/bin/sh"
               - "-c"
               - "rm /tmp/READY && until make all_nodepools_running; do echo waiting for nodepools to be created; sleep 10; done; make clean; kill -9 -1"
-              # wait until all_nodepools_running returns a success code then run make clean and kill the container
+              # Remove ready status, wait until all_nodepools_running returns a success code then run make clean and kill the container.
```
krasi-georgiev: Would you mind putting the comment above the command to make it consistent with the other places and also it is the most common way of commenting.

pkg/provider/provider.go
```diff
@@ -40,11 +40,11 @@ type Resource struct {
 // RetryUntilTrue returns when there is an error or the requested operation returns true.
 func RetryUntilTrue(name string, retryCount int, fn func() (bool, error)) error {
 	for i := 1; i <= retryCount; i++ {
+		time.Sleep(globalRetryTime)
```
krasi-georgiev, line 43: why did you move it here?

geekodour: for waiting on update, it takes sometime (readiness probe+threshold) for the pod to become unready/ready. so we wait before we check the status. because when updating, the previous state will be `ready` and we need to wait sometime for it to change.

prombench/manifests/prombench/stateful-set.yaml
```diff
@@ -42,8 +42,9 @@ spec:
         image: docker.io/prombench/prombench:2.0.2
         imagePullPolicy: Always
         args:
-        # sleep forever after make deploy succeeded
-        - "make deploy && echo 'deploy succeeded, now sleeping' && while sleep 86400; do :; done"
+        # - Create /tmp/READY to marmake k the StatefulSet as ready once make deploy succeeds
```
krasi-georgiev:
```suggestion
        # Create /tmp/READY to mark the StatefulSet as ready once make deploy succeeds.
```

pkg/provider/k8s/k8s.go
```diff
@@ -482,17 +483,19 @@ func (c *K8s) statefulSetApply(resource runtime.Object) error {
...
 		}
-		log.Printf("resource created - kind: %v, name: %v", kind, req.Name)
 	default:
 		return fmt.Errorf("unknown object version: %v kind:'%v', name:'%v'", v, kind, req.Name)
 	}
+
 	return provider.RetryUntilTrue(
 		fmt.Sprintf("applying statefulSet:%v", req.Name),
-		provider.GlobalRetryCount,
+		6*provider.GlobalRetryCount, // StatefulSet used to start test
```
krasi-georgiev: Don't quite understand this comment.

PR #267, opened 2019-08-22
added date check github action fix

geekodour:
temporary check to check if github actions fixed the date issue mentioned in : https://github.com/google/go-github/issues/1254

### DIFF
--- a/tools/commentMonitor/main.go
+++ b/tools/commentMonitor/main.go
@@ -90,6 +90,10 @@ func main() {
 		txt = reg.ReplaceAllString(txt, "$1\"2019-06-11T09:26:28Z$2")
 		data = []byte(txt)
 		log.Println("temp fix active")
+	} else {
+		log.Println(`WARNING: Github actions outputs correct date format so can
+		remove the workaround fix in the code commentMonitor.
+		https://github.com/google/go-github/issues/1254`)
 	}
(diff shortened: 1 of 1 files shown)


tools/commentMonitor/main.go
```diff
@@ -83,11 +83,14 @@ func main() {
...
 	reg := regexp.MustCompile("(.*)\"[0-9]+/[0-9]+/2019 [0-9]+:[0-9]+:[0-9]+ [AP]M(.*)")
-	txt = reg.ReplaceAllString(txt, "$1\"2019-06-11T09:26:28Z$2")
-	data = []byte(txt)
-	log.Println("temp fix active")
+	m := reg.FindSubmatch(data)
+	if m != nil {
+		txt := string(data)
+		txt = reg.ReplaceAllString(txt, "$1\"2019-06-11T09:26:28Z$2")
+		data = []byte(txt)
+		log.Println("temp fix active")
```
krasi-georgiev, line 92: Someone that is not familiat with this workaround wouldn't know what this means when looking at the logs.

Let have something like:

If the regex is not found than the date is probably fixed so only than output a log like:
`WARNING: Github  actions outputs correct date format so can remove the workaround fix in the code commentMonitor. https://github.com/google/go-github/issues/1254`

PR #270, opened 2019-08-26
[WIP] [prombench] Log on existing nodepool and no error on service re-apply/update

### LINKED ISSUE #44: When applying some resource files getting an error:  Invalid value: "": must be specified for an update
prombench gke resource apply -a $AUTH_FILE -v ZONE:$ZONE -v CLUSTER_NAME:$CLUSTER_NAME \
-v GCLOUD_SERVICEACCOUNT_CLIENTID:$GCLOUD_SERVICEACCOUNT_CLIENTID \
-f components/prow/manifests/rbac.yaml -f components/prow/manifests/nginx-controller.yaml

for example applying this twice renders the error below

error while applying a resource err:apply resources from manifest file:components/prow/manifests/nginx-controller.yaml: resource update failed - kind: Service, name: default-http-backend: Service "default-http-backend" is invalid: [metadata.resourceVersion: Invalid value: "": must be specified for an update, spec.clusterIP: Invalid value: "": field is immutable]

geekodour:
Closes #44 

and we need to just log it if nodepool already exists, otherwise if there's an error when applying prombench components this does into a crash loop because `make deploy` is run and it errors out `nodepool already exits` which is not a problem when trying to recover from a crash.

this is different from benchmark restarts where the initContainer is run again too, but in this case just the main container crashes (if there's an error in deploying) and when it tries to redeploy it finds that there's an existing nodepool and crashes again. 

this PR fixes it, now there will be attempts to redeploy components after they fail automatically. and once there is a timeout in github actions. we'll see the fail comment on the PR.

also updated the `WORKDIR` because `make clean` was running from another directory since we didn't have a `WORKDIR` set.

### DIFF
--- a/pkg/provider/gke/gke.go
+++ b/pkg/provider/gke/gke.go
@@ -322,7 +322,10 @@ func (c *GKE) NodePoolCreate(*kingpin.ParseContext) error {
 				})
 
 			if err != nil {
-				log.Fatalf("Couldn't create cluster nodepool '%v', file:%v ,err: %v", node.Name, deployment.FileName, err)
+				if st, ok := status.FromError(err); ok && st.Code() != codes.AlreadyExists {
+					log.Fatalf("Couldn't create cluster nodepool '%v', file:%v ,err: %v", node.Name, deployment.FileName, err)
+				}
+				log.Printf("nodepool already exists '%v', file:%v ,err: %v", node.Name, deployment.FileName, err)
 			}
 
 			err = provider.RetryUntilTrue(
--- a/pkg/provider/k8s/k8s.go
+++ b/pkg/provider/k8s/k8s.go
@@ -771,12 +771,19 @@ func (c *K8s) serviceApply(resource runtime.Object) error {
 		for _, l := range list.Items {
 			if l.Name == req.Name {
 				exists = true
+				if req.ResourceVersion == "" {
+					req.ResourceVersion = l.ResourceVersion
+				}
+				if req.Spec.ClusterIP == "" {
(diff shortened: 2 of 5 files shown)


geekodour:
https://github.com/prometheus/prombench/pull/264
introduced some inconsistency to the how `StatefulSets` work with our tests because prombench now exits on error. This PR is WIP to make the tests stable.

geekodour:
@krasi-georgiev should we consider rolling back to just logging on error when deploying manifest. exiting on error definitely should be the default behavior but since we worked on the StatefulSet not assuming that behavior of prombench it's causing some issues now in some edge case. (eg. cancelling a benchmark when it is getting started)

Will keep working on this PR to make StatefulSet work when prombench exits on error

geekodour:
```
/prombench start --> suppose benchmark error exits in some stage of resource_apply

/prombench cancel --> (while the above is ongoing before the error exit happens)  --> `make clean` is run
```

so at some point both `make clean` and `make deploy` are running, so if the container exits with some error in `make deploy` , `make clean` won't be able to run to completion. 

We needed error on exit because github actions wants error code to know if some error happened and run commentMonitor according to that. 

I thought using an init system and running two processes under PID1 will let us keep running the container till `make clean` exits and manually calls `kill -9 -1` after that.

I opened an issue with the tini init system regarding that: https://github.com/krallin/tini/issues/139

But I am not sure how should we proceed with this. rolling back to just logging on error and only erroring out on the statefulset run from github actions is an option. we can add a flag `--exit-on-error` flag for that.

cc @krasi-georgiev

geekodour:
See #284

PR #274, opened 2019-08-29
order bench results by magnitutde

krasi-georgiev:
cc @VladMasarik 	

for long bench results like this one I think it is best to have the results sorted by default
https://github.com/prometheus-community/prometheus/pull/7#issuecomment-526161459

### DIFF
--- a/funcbench/main.go
+++ b/funcbench/main.go
@@ -257,7 +257,7 @@ func newBenchmarkTester() (*benchmarkTester, error) {
 }
 
 func (bench *benchmarkTester) compareBenchmarks(old, new string) (string, error) {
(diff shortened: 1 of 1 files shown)


krasi-georgiev:
shoot wrong repo

VladMasarik:
Sounds reasonable. One more thing, what about benchmarks with long results? They are what we expect, but aren't they a bit, too long? Is not interfering with the ability to read comments easily? I feel like the resulting comment holds too much information and it makes it impractical.

krasi-georgiev:
that is already part of https://github.com/prometheus/prombench/pull/279

> Sounds reasonable. One more thing, what about benchmarks with long results? They are what we expect, but aren't they a bit, too long? Is not interfering with the ability to read comments easily? I feel like the resulting comment holds too much information and it makes it impractical.

I think it is not that bad, but I remember that there was a flag to only show top .... results.

PR #275, opened 2019-08-29
Use the alpine image and disable CGO

krasi-georgiev:
reduces image size by 50%

cc @VladMasarik 	


VladMasarik:
LGTM, Thank you!

PR #276, opened 2019-08-29
Build prometheus only for prometheus builder

geekodour:
Note: This will build both prometheus and promtool on commits before https://github.com/prometheus/prometheus/pull/5966


geekodour:
@krasi-georgiev I've tested this, works as expected. prometheus gets built before 30 tries and prombench test starts successfully now. can be merged if no other change is needed.

PR #278, opened 2019-08-30
[WIP] Added exit-on-error flag and set WORKDIR in Dockerfile

geekodour:
Added some changes to the prombench cli and Dockerfile.

Change in the Dockerfile was necessary because running `make` expects the Makefile to be in the working directory when running from the `preStopHook`


geekodour:
see #284

PR #279, opened 2019-08-30
varios refactorings

krasi-georgiev:
added a config to show the file line with each log.
removed some logs entries that might expose secrets by mistake.
refactored the "revert pr changes" function.
added `-mag` to benchcmp to sort the results.

cc @VladMasarik 

### DIFF
--- a/funcbench/main.go
+++ b/funcbench/main.go
@@ -48,6 +48,8 @@ type benchmarkTester struct {
 }
 
 func main() {
+	// Show file line with each log.
+	log.SetFlags(log.Ltime | log.Lshortfile)
 	app := kingpin.New(filepath.Base(os.Args[0]), "benchmark result posting and formating tool.\n-i location of github hook file (even.json)")
 	app.HelpFlag.Short('h')
 	input := app.Flag("input", "path to event.json").Short('i').Default("/github/workflow/event.json").String()
@@ -84,10 +86,6 @@ func main() {
 		log.Fatalln(err)
 	}
 	logLink := fmt.Sprintf("Full logs at: https://github.com/%s/%s/commit/%s/checks", ghClient.owner, ghClient.repo, ghClient.latestCommitHash)
-
-	if err := os.Chdir(os.Getenv("GITHUB_WORKSPACE")); err != nil {
-		log.Fatalln(err)
-	}
 
 	if err := gitClient.cloneRepository(); err != nil {
 		log.Fatalln(err)
@@ -145,7 +143,6 @@ func newGitHubClient(event *github.IssueCommentEvent) *gitHubClient {
 		prNumber:         *event.GetIssue().Number,
 		latestCommitHash: os.Getenv("GITHUB_SHA"),
 	}
-	log.Printf("GH Client: %+v\n", c)
 	return &c
 }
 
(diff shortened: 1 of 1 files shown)


funcbench/main.go
```diff
@@ -205,26 +206,19 @@ func (c *gitClient) checkoutPR(num int) error {
...
-	return filepath.Walk(os.Getenv("GITHUB_WORKSPACE"), func(path string, info os.FileInfo, err error) error {
+	return filepath.Walk(filepath.Join(os.Getenv("GITHUB_WORKSPACE"), c.repo), func(path string, info os.FileInfo, err error) error {
 		if err != nil {
 			return err
 		}
-		if info.IsDir() || strings.Contains(path, ".git/") || strings.HasSuffix(info.Name(), "test.go") {
-			return nil
+		if info.IsDir() && info.Name() == ".git" {
+			return filepath.SkipDir
 		}
```
VladMasarik: I do not think this `if` statement is necessary. AFAIK, it will skip only `.git` directory, but not files inside the `.git` directory, and you already covered the case of skipping files inside `.git` directory since, AFAIK there are no `.go` file in the `.git` directory.

Also, do you think there might be a case where we want to checkout files that don't end with `.go`?

Other than that, great suggestion!

funcbench/main.go
```diff
@@ -172,22 +169,26 @@ func newGitClient(event *github.IssueCommentEvent) (*gitClient, error) {
...
-	err = os.Chdir(c.repo)
-	return err
+	return nil
 }
 
 // checkoutToPullRequest applies changes from the pull request to the working tree
 // of the branch that is being compared.
 func (c *gitClient) checkoutPR(num int) error {
+	if err := os.Chdir(filepath.Join(os.Getenv("GITHUB_WORKSPACE"), c.repo)); err != nil {
+		return errors.Wrap(err, "changing to GITHUB_WORKSPACE dir")
```
VladMasarik, line 190: Personal interest question, is it better to report what is happening when the error occurred or report why the error occurred?

funcbench/main.go
```diff
@@ -172,22 +169,26 @@ func newGitClient(event *github.IssueCommentEvent) (*gitClient, error) {
 		repo:   *event.GetRepo().Name,
 	}
 
-	log.Printf("Git Client: %+v\n", c)
 	return &c, err
 }
 
 func (c *gitClient) cloneRepository() error {
+	if err := os.Chdir(os.Getenv("GITHUB_WORKSPACE")); err != nil {
+		log.Fatalln(err)
```
VladMasarik:
```suggestion
		return err
```

funcbench/main.go
krasi-georgiev: I don't quite understand, can you give an example

funcbench/main.go
krasi-georgiev: yes you are right I tested it and it still scans the files inside that dir. I though it would skip the entire folder but unfortunately it doesn's.
https://play.golang.org/p/pPAVotsvLs8

Removed it.

funcbench/main.go
VladMasarik: When an error occurs, what is better error description:
"changing to GITHUB_WORKSPACE dir" or
"Could not change directory to XXX"?

VladMasarik:
LGTM

funcbench/main.go
krasi-georgiev: I think either one works.

PR #280, opened 2019-09-02
Added supervisord to prombench docker image

geekodour:
#264 introduced exit on error behavior for apply/delete in prombench. This became a problem because the container ran that script as PID1 and would restart the container if `make deploy` or `make clean` errors out resulting in not letting any other process run on the container.

This PR uses a combination of [tini init system](https://github.com/krallin/tini) and  [supervisord](http://supervisord.org/) to solve it since the container started by the statefulset to start/stop a test needs to be able to run multiple scripts at once.

supervisorctl also gives a very nice way to check status of the various scripts inside the container, for eg:
```
/prombench # supervisorctl status
clean                            STOPPED   Not started
deploy                           STOPPED   Not started
init                             RUNNING   pid 9, uptime 0:00:34
signal-supervisor                RUNNING   pid 8, uptime 0:00:34
```

Closes #278 
Closes #270 

Changes Added:
- ~Log fatal instead of returning error in `AllNodepoolsRunning` and `AllNodepoolsDeleted`~
- ~Revert `Makefile` clean target to `clean: resource_delete nodepool_delete` , as sometimes `nodepool_delete` takes longer and times out.~
- Added [tini init system](https://github.com/krallin/tini) so that we can run it as PID 1
- Added [supervisord](http://supervisord.org/) since [tini only tracks a single child](https://github.com/krallin/tini/issues/139#issuecomment-525944484). (supervisord is [recommended by docker](https://docs.docker.com/config/containers/multi-service_container/))
- Updated prombench Dockerfile to use `WORKDIR`
- removed initContainer from `stateful-set.yaml`, put that logic into `deploy.sh`
- ~split `prombench/manifests/prombench/benchmark/1c_cluster-role-binding.yaml` into two files as the `Role` and `RoleBinding` were getting deleted when deleting the namespace. this gets rid of the error message `Role: loadgen-scaler` and `RoleBinding: loadgen-scaler` not found when running `make clean`~

Additionally, with this approach we are also able to see the logs for `make clean` in loki which was not possible previously(with statefulsets)
https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#debugging-hook-handlers

cc @krasi-georgiev


geekodour:
See #284

PR #282, opened 2019-09-04
change order of deletion in resource_delete

geekodour:
Gets rid of the following error, this error is fatal now as prombench exits on error.
```
11:51:21 gke.go:576: error while deleting objects from a manifest file err:error deleting 'manifests/prombench/benchmark/1c_cluster-role-binding.yaml' err:resource delete failed - kind: Role, name: loadgen-scaler: roles.rbac.authorization.k8s.io "loadgen-scaler" not found
make: *** [Makefile:41: resource_delete] Error 1
```

cc @krasi-georgiev


krasi-georgiev:
do you know why this happens?

geekodour:
```
resource_delete:
	$(PROMBENCH_CMD) gke resource delete -a ${AUTH_FILE} \
		-f manifests/prombench/benchmark/1a_namespace.yaml \
        -f manifests/prombench/benchmark/1c_cluster-role-binding.yaml
```
this target deletes the namespace first and also deletes the roles and rolebindings. but the clusterroles are not part  of the namespace so they need to be deleted separately.

since `loadgen-scaler` role and rolebindings were in `1c_cluster-role-binding.yaml` and were already deleted when deleting the namespace. hence this error

krasi-georgiev:
I think these are executed in the sequence they are provided so maybe can combine  nodepool_delete and resource_delete and just have something like:

```
delete:
	$(PROMBENCH_CMD) gke resource delete -a ${AUTH_FILE} \
        -f manifests/prombench/benchmark/1c_cluster-role-binding.yaml \
		-f manifests/prombench/benchmark/1a_namespace.yaml \				
		-f manifests/prombench/nodepools.yaml
```

this will delete the role bindings first, than the namespace and only than the nodepools, right?

geekodour:
@krasi-georgiev tested by changing the order, working now. 
about,
```
-f manifests/prombench/nodepools.yaml
```
we are using a separate command `./prombench gke nodepool delete` so kept that  as is.

krasi-georgiev:
sorry I didn't understand, why do we need `nodepool_delete` and `resource_delete`, why not just have one?

geekodour:
@krasi-georgiev the prombench commands for each is different,

https://github.com/prometheus/prombench/blob/0498f012444b7b550e2780b8d0011ca3d93e8c01/prombench/Makefile#L47

vs

https://github.com/prometheus/prombench/blob/0498f012444b7b550e2780b8d0011ca3d93e8c01/prombench/Makefile#L41

krasi-georgiev:
ooh yes, I completely forgot about that.

PR #285, opened 2019-09-13
*: bump Go version to 1.13

simonpasquier:


geekodour:
@simonpasquier @krasi-georgiev  won't we need a change here too? :

https://github.com/prometheus/prombench/blob/8fa36bf45520bfca6a42ba7b850750428c01a4a8/prombench/temp/prometheus-builder/Dockerfile#L1

krasi-georgiev:
yep defensively.

geekodour:
Just saw prometheus/prometheus is still using 1.12 so maybe not? 
https://github.com/prometheus/prometheus/blob/master/.promu.yml

simonpasquier:
@geekodour once https://github.com/prometheus/prometheus/pull/6020 is merged, yes you would need to bump to 1.13 there too.

PR #290, opened 2019-09-18
[DO NOT MERGE] Added container resources limits to fake-webserver Deployment

geekodour:
After upgrading to 1.13.x in GKE, fake-webserver fails to scale up because k8s puts some limits on memory usage. So we explicitly set limits for fake-webserver. The problem persists in 1.14.x too.

After adding these resource limits and requests, prombench tests run as expected.

See #286 
See https://github.com/kubernetes/kubernetes/issues/73572

The set limit is arbitrary, had to set something above 2GiB but was unsure how much fake-webserver consumes.

more on differences in limits and requests for resources:
https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#how-pods-with-resource-limits-are-run

### DIFF
--- a/prombench/manifests/prombench/benchmark/2_fake-webserver.yaml
+++ b/prombench/manifests/prombench/benchmark/2_fake-webserver.yaml
@@ -27,7 +27,7 @@ data:
             requests:
               memory: 80Mi
             limits:
-              memory: 7600Mi
+              memory: 99999Mi
             ports:
             - name: metrics1
               containerPort: 8080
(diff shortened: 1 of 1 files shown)


krasi-georgiev:
could you find out if there is way to set these as unlimited or to a high enough number to not have to think out it.

geekodour:
from the docs at https://kubernetes.io/docs/tasks/configure-pod-container/assign-memory-resource/#if-you-do-not-specify-a-memory-limit

> The Container has no upper bound on the amount of memory it uses. 

not sure why this behavior is happening, maybe more investigation is needed but the current approach is done using setting the limit to ~8GB for a container which seems to be working when scaled to `100` by the prombench `scaler`. I'll try to look more into this.

PR #295, opened 2019-10-10
Merged and added _noparse suffix to grafana dashboard configmaps

geekodour:
Added `_noparse` suffix to grafana dashboard configmap files added in #293 as well removed the extra `{{` used to escape golang template.

Currently `subPath` is being used to point grafana to the dashboards, so if we want grafana to reflect changes made to the dashboard configmaps, we'll need to redeploy grafana. When not using `subPath` we don't face this issue, so just updating the configmap will show the changes in the grafana dashboards. Hence merged both of the dashboards into one configmap.

From the docs:
>  Note: A container using a ConfigMap as a subPath volume will not receive ConfigMap updates.

more info on the `subPath` problem:
- https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#mounted-configmaps-are-updated-automatically
- https://github.com/kubernetes/kubernetes/issues/50345

Merged `/manifests/cluster-infra/grafana_dashboard_prombench.yaml` and
`/manifests/cluster-infra/grafana_dashboard_node_metrics.yaml`
into `/manifests/cluster-infra/grafana_dashboard_dashboards_noparse.yaml`


prombench/manifests/cluster-infra/grafana_deployment.yaml
```diff
@@ -72,12 +66,9 @@ spec:
...
         configMap:
           name: grafana-dashboard-provision
-      - name: grafana-dashboard-prombench
+      - name: grafana-dashboard-dashboards
         configMap:
-          name: grafana-dashboard-prombench
-      - name: grafana-dashboard-node-metrics
-        configMap:
-          name: grafana-dashboard-node-metrics
+          name: grafana-dashboard-dashboards
```
krasi-georgiev: why not just 
```suggestion
          name: grafana-dashboards
```

the same at the other place

geekodour: okay adding the change

PR #298, opened 2019-10-25
[WIP] Added webhook support for commentMonitor

geekodour:
This PR removes the ability to read `event.json` and act on its content, instead it runs as a webhook and emits `repository_dispatch` events. 


geekodour:
cc @krasi-georgiev

tools/commentMonitor/main.go
```diff
@@ -14,219 +14,217 @@
...
-					log.Fatalf("%v : couldn't post comment", err)
-				}
-				os.Exit(1)
-			}
-			log.Println("author is a member or collaborator")
+		// Validate regex.
+		err := cmClient.validateRegex()
+		if err != nil {
+			//log.Println(err) // Don't log on failure.
+			http.Error(w, "comment validation failed", http.StatusBadRequest)
```
krasi-georgiev: with most api's when you send a non 200 ok reply the same request is repeated which is not what we want here. Best to just return 200

I think `validateRegex()` should use bool as the return statment

tools/commentMonitor/client.go
```diff
@@ -0,0 +1,129 @@
...
+	commentTemplate string
+}
+
+// Validate comment if regexString provided.
+func (c *commentMonitorClient) validateRegex() error {
+	for _, e := range c.eventMap {
+		c.regex = regexp.MustCompile(e.RegexString)
+		if c.regex.MatchString(c.ghClient.commentBody) {
+			c.commentTemplate = e.CommentTemplate
+			c.eventType = e.EventType
```
krasi-georgiev, line 43: why do we need this variable?

krasi-georgiev (commented):
also don't forget to update the main prombench readme

tools/commentMonitor/client.go
geekodour: These are later used in `createRepositoryDispatch` call in the `extractArgs` function in this file.

krasi-georgiev: yep I sew how it is used now. thanks

PR #307, opened 2019-11-24
[RDY] Added webhook secret to comment-monitor

geekodour:
This PR adds check for the webhook secret for comment-monitor. This will allow webhook requests coming from GitHub only.

I've tested this with a new cluster, it's working as expected.


tools/commentMonitor/main.go
```diff
@@ -73,6 +78,13 @@ func main() {
 	if len(cmConfig.eventMap) == 0 {
 		log.Fatalln("eventmap empty")
 	}
+
+	// Get webhook secret.
+	cmConfig.whSecret, err = ioutil.ReadFile(cmConfig.whSecretFilePath)
```
krasi-georgiev: I think best to read the secret on every request. This way if you change the secret in k8s you won't need to restart the tool.

geekodour: I added a `/-/reload` endpoint to reload the config file and webhook secret. It's working locally, yet to test it on a running cluster.

krasi-georgiev: I think a reload endpoint is an overkill at this point, why not just read the file at every request?

geekodour: With `/-/reload` it is reloading the `eventmap.yml` file as well as the webhook secret file. If we do it with every request then the eventmap and the webhook will be loaded for every `IssueComment` event on `prometheus/prometheus` on GitHub. Will loading both of these files be a good idea on every request? I thought, as `eventmap.yml` is like the configuration file for comment-monitor it'll be a good idea to add `/-/reload` similar to Prometheus. Please let me know if we should read both of the files on every request, I'll add the change.

krasi-georgiev: yep it is absolutely fine to read those on every request. I did a test recently and the kernel caches non chnaged files in memory so esensitally these are read from memory when not changed.

geekodour: @krasi-georgiev added the change.

krasi-georgiev:
yep , good to go, Thanks!

Issue #310, opened 2019-12-18
Make sure that Prombench uses sole tenants.


krasi-georgiev:
https://cloud.google.com/compute/docs/nodes/create-nodes

If the tests run in a host with shared VMs than this means that the results will not be consistent between different VMs

krasi-georgiev:
concluded that this is not needed

PR #311, opened 2019-12-18
Run funcbench via GKE

nevill:
xref #310 

it includes:
- [x] Run `funcbench` as a Kubernetes Job
- [x] Drop param `-i event.json` from `funcbench` cli, add `org / repo / branch / prNumber / regex / raceEnabled / postComment` as input parameters
- [ ] Replace docker images `nevill/prombench`, `nevill/funcbench` once the latest are released
- [ ] Let `funcbench` run on Sole-tenant nodes

Once it's merge, we also have to update Github actions in main repository

### DIFF
--- a/funcbench/manifests/benchmark/2_job.yaml
+++ b/funcbench/manifests/benchmark/2_job.yaml
@@ -30,6 +30,8 @@ metadata:
     app: prometheus
     funcbench: test-{{ .PR_NUMBER }}
 spec:
+  # never re-create a new pod
+  backoffLimit: 0
   template:
     metadata:
       namespace: funcbench-{{ .PR_NUMBER }}
--- a/funcbench/manifests/benchmark/2_job.yaml
+++ b/funcbench/manifests/benchmark/2_job.yaml
@@ -21,8 +21,8 @@ data:
   RACE: {{ .RACE }}
   REGEX: {{ .REGEX }}
 ---
-apiVersion: apps/v1
-kind: Deployment
+apiVersion: batch/v1
+kind: Job
 metadata:
   name: funcbench-test-{{ .PR_NUMBER }}
   namespace: funcbench-{{ .PR_NUMBER }}
@@ -30,11 +30,6 @@ metadata:
     app: prometheus
     funcbench: test-{{ .PR_NUMBER }}
 spec:
-  replicas: 1
-  selector:
-    matchLabels:
-      app: prometheus
-      funcbench: test-{{ .PR_NUMBER }}
   template:
     metadata:
       namespace: funcbench-{{ .PR_NUMBER }}
@@ -42,6 +37,7 @@ spec:
         app: prometheus
         funcbench: test-{{ .PR_NUMBER }}
     spec:
+      restartPolicy: Never
       initContainers:
       - name: clean-up
         image: busybox:latest
--- a/funcbench/manifests/benchmark/2_job.yaml
+++ b/funcbench/manifests/benchmark/2_job.yaml
@@ -26,18 +26,10 @@ kind: Job
 metadata:
   name: funcbench-test-{{ .PR_NUMBER }}
   namespace: funcbench-{{ .PR_NUMBER }}
-  labels:
-    app: prometheus
-    funcbench: test-{{ .PR_NUMBER }}
 spec:
   # never re-create a new pod
   backoffLimit: 0
   template:
-    metadata:
-      namespace: funcbench-{{ .PR_NUMBER }}
-      labels:
-        app: prometheus
-        funcbench: test-{{ .PR_NUMBER }}
     spec:
       restartPolicy: Never
       initContainers:
--- a/funcbench/manifests/benchmark/2_job.yaml
+++ b/funcbench/manifests/benchmark/2_job.yaml
@@ -1,26 +1,3 @@
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: workflow
-  namespace: funcbench-{{ .PR_NUMBER }}
-data:
-  event.json: |
-    {
-      "action": "created",
-      "issue": {
-        "number": {{ .PR_NUMBER }}
-      },
-      "repository": {
-        "name": "{{ .GITHUB_REPO }}",
-        "owner": {
-          "login": "{{ .GITHUB_ORG }}"
-        }
-      }
-    }
-  BRANCH: {{ .BRANCH }}
-  RACE: {{ .RACE }}
-  REGEX: {{ .REGEX }}
----
 apiVersion: batch/v1
 kind: Job
 metadata:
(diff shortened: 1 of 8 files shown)


pkg/provider/k8s/k8s.go
```diff
@@ -179,6 +180,10 @@ func (c *K8s) ResourceApply(deployments []Resource) error {
 				err = c.customResourceApply(resource)
 			case "statefulset":
 				err = c.statefulSetApply(resource)
+			case "job":
+				err = c.jobApply(resource)
+				// Even job deletion fails, it will be removed together with nodepool.
```
krasi-georgiev:
```suggestion
				// Ignore the error as even when job deletion fails, it will be removed together with nodepool.
```

pkg/provider/k8s/k8s.go
```diff
@@ -1282,6 +1354,37 @@ func (c *K8s) statefulSetReady(resource runtime.Object) (bool, error) {
...
+	switch v := resource.GetObjectKind().GroupVersionKind().Version; v {
+	case "v1":
+		client := c.clt.BatchV1().Jobs(req.Namespace)
+
+		res, err := client.Get(req.Name, apiMetaV1.GetOptions{})
+		if err != nil {
+			return false, errors.Wrapf(err, "Checking Job resource:'%v' status failed err:%v", req.Name, err)
+		}
+
+		// current jobReady only works for non-parallel jobs
```
krasi-georgiev: Please follow Golang commentary guidelines.
https://golang.org/doc/effective_go.html#commentary

funcbench/manifests/benchmark/2_job.yaml
```diff
@@ -0,0 +1,78 @@
...
+        env:
+          - name: GITHUB_WORKSPACE
+            value: /build/src
+          - name: GITHUB_TOKEN
+            value: {{ .GITHUB_TOKEN }}
+          - name: GITHUB_SHA
+            value: {{ .LAST_COMMIT_SHA }}
+          - name: GOTMPDIR
+            value: /build/tmp
+          - name: TMPDIR
```
krasi-georgiev: why do we need these 2 env vars? GOTMPDIR and TMPDIR

can we use the default linux tmp dir for all these?

also why do we need GITHUB_WORKSPACE?

funcbench/manifests/benchmark/2_job.yaml
```diff
@@ -0,0 +1,78 @@
...
+      "repository": {
+        "name": "{{ .GITHUB_REPO }}",
+        "owner": {
+          "login": "{{ .GITHUB_ORG }}"
+        }
+      }
+    }
+  BRANCH: {{ .BRANCH }}
+  RACE: {{ .RACE }}
+  REGEX: {{ .REGEX }}
```
krasi-georgiev: what are these for?

funcbench/manifests/benchmark/2_job.yaml
```diff
@@ -0,0 +1,78 @@
...
+      initContainers:
+      - name: clean-up
+        image: busybox:latest
+        command: ['sh', '-c', 'rm -fr /build/* && mkdir /build/tmp && mkdir /build/src']
+        volumeMounts:
+        - name: instance-ssd
+          mountPath: /build
+      containers:
+      - name: funcbench
+        image: nevill/funcbench:latest
```
krasi-georgiev: we should use the official images `prombench/funcbench`

funcbench/manifests/nodepools.yaml
```diff
@@ -0,0 +1,15 @@
+zone: {{ .ZONE }}
+projectid: {{ .PROJECT_ID }}
+cluster:
+  name: {{ .CLUSTER_NAME }}
+  nodepools:
+  # These node-pools will be deployed on triggering benchmark
+  - name: funcbench-{{ .PR_NUMBER }}
+    initialnodecount: 1
+    config:
+      machinetype: n1-highmem-4
```
krasi-georgiev: Is this the smalles machine that can be run as Sole-tenant nodes. If not please set to the lowest one possible to keep the cost of running these very low.

krasi-georgiev (commented):
Any changes required to the funcbnech cli?

funcbench/manifests/benchmark/2_job.yaml
nevill: These 3 variables `GITHUB_WORKSPACE`, `GOTMPDIR`, `TMP_DIR` are used for generated testing binary, to make sure tests are run on the attached local SSD disk. But it can be replaced by specify disktype to `pd-ssd` when creating a VM. These will be removed later.

funcbench/manifests/benchmark/2_job.yaml
nevill: They are the inputs of `funcbench`.
See https://github.com/prometheus/prombench/blob/aad5c780d5cb2ef3e866f49f2b4a500416217bf1/funcbench/main.go#L231-L235
and 
https://github.com/prometheus/prombench/blob/aad5c780d5cb2ef3e866f49f2b4a500416217bf1/funcbench/main.go#L165

nevill:
> Any changes required to the funcbnech cli?

No, there isn't any right now. But @geekodour should make sure the docker image is up to date.

funcbench/manifests/benchmark/2_job.yaml
krasi-georgiev: why not just use the linux tmp dir?
/tmp

funcbench/manifests/benchmark/2_job.yaml
krasi-georgiev: why not set them as env vars?

geekodour: @nevill commentMonitor no longer works by sharing variables in files. To avoid confusion, I think it'll be best to replace these with command line arguments or environment variable whichever is more preferable here. I might be misunderstanding aswell, can you share the github actions workflow file used to trigger funcbench here?

krasi-georgiev: yes actually I also meant command line variable

nevill: @geekodour the workflow file is [here](https://github.com/nevill/prometheus/blob/master/.github/workflows/benchmark.yml)

funcbench/manifests/benchmark/2_job.yaml
nevill: because `/tmp` is not on a SSD disk.

krasi-georgiev: in this case this doesn't matter as there is no requirement for high io

funcbench/manifests/benchmark/2_job.yaml
krasi-georgiev: this is to be changed to cli args right?

nevill: right, will update this soon

nevill:
#343 will replace this one.

PR #318, opened 2020-01-27
comment-monitor fixes

geekodour:
Added the following fixes:

- Won't fatal exit comment-monitor if unexpected event is received.
- Updated Loki explorer link in Github comment.
- Added helpful message about adding the version if someone simply uses `/prombench`

cc @krasi-georgiev


prombench/manifests/cluster-infra/7a_commentmonitor_configmap_noparse.yaml
```diff
@@ -26,6 +26,12 @@ data:
       comment_template: |
         Benchmark cancel is in progress.
 
+    - event_type: noop
+      regex_string: (?mi)^/prombench\s*$
+      comment_template: |
+        Please add the version number to compare against.
```
krasi-georgiev, line 32:
```suggestion
        Please add the branch to compare against.
```

geekodour: why should this be the branch?

krasi-georgiev: aah yeah you are right is it using the docker image.

PR #218, opened 2019-07-04
Auto build and push docker images

nikita-mk:
See #100

PR #320, opened 2020-01-30
Use CircleCI to build and publish docker images automatically

### LINKED ISSUE #100: Auto build and push all used docker images.
Currently we build and push images on every code change. 
Lets use circle or Travis to automate this.

nevill:
This is another effort to fix #100 .
Basically, it will
1. build all the binaries with `promu`
2. build and push docker image one by one


nevill:
@geekodour Please take an early review. 
I am not sure about the version semantic of this project, I am using `YYYYMMDD` as a version tag  now. Also, you may have a look at `filters` in the `workflows` definitions.

Makefile.common
```diff
@@ -109,7 +99,7 @@ endif
 %: common-% ;
 
 .PHONY: common-all
-common-all: precheck style check_license lint unused build test
+common-all: precheck style check_license lint build test unused
```
geekodour: @nevill I  think we should not make changes to the `Makefile.common` file, overriding them in the `Makefile` should do because the `Makefile.common` is updated by @prombot when changes to `Makefile.common` are made in prometheus/prometheus

@krasi-georgiev can comment more on this if we decide to change this.

nevill: Yes, I also noticed PR #196 , which is where `Makefile.common` came in the project. I will try to make my changes into `Makefile` instead.

IMO `Makefile.common` is quite redundant, for example, `GO111MODULE` is always `on` in this project. It's not friendly to new contributors as well.

More interestingly, from [promu document](https://docs.google.com/document/d/1Ql-f_aThl-2eB5v3QdKV_zgBdetLLbdxxChpy-TnWSE/edit#heading=h.24x0bg1hyuak)
> The Makefile.COMMON approach was abandoned for Prometheus and Alertmanager eventually.

tools/commentMonitor/Dockerfile
```diff
@@ -1,16 +1,6 @@
-# NOTE: building this requires it to have the build context of the repository root
-FROM golang:1.12-alpine AS builder
+FROM quay.io/prometheus/busybox:latest
+LABEL maintainer="The Prometheus Authors <prometheus-developers@googlegroups.com>"
```
geekodour, line 11: `commentMonitor` and `amGithubNotifier` are built successfully when using multi-staged builds, is there any specific reason to revert this other than to have a consistent way to build every tool? @krasi-georgiev suggested we should move towards multi-staged builds, so I think the ones that are working can be kept that way if possible? otherwise, we add this change.

cc: @krasi-georgiev

.circleci/config.yml
```diff
@@ -11,11 +15,62 @@ executors:
...
+    - run:
+        name: Setup environment variables
+        command: |
+          echo "export TAG=`date +%Y%m%d`" >> $BASH_ENV
+          echo "DOCKER_REPO=<< parameters.registry >>/${DOCKER_ORG:-<< parameters.organization >>}" >> $BASH_ENV
+          echo "DOCKER_IMAGE=<< parameters.image_name >>" >> $BASH_ENV
+    - when:
+        condition: <<parameters.dockerfile_path>>
+        steps:
+        - run: docker build -t $DOCKER_REPO/$DOCKER_IMAGE:$TAG -f << parameters.dockerfile_path >> << parameters.context_path >>
```
geekodour: Did you try using the `Makefile.common` docker commands here? setting, `DOCKERFILE_PATH`. `DOCKER_IMAGE_NAME`, `DOCKER_IMAGE_TAG`, `DOCKERBUILD_CONTEXT` in the `Setup environment variables` step should enable the following command to build the image using `make docker` similarly for `make docker-publish` and `make docker-manifest`.

What `DOCKER_IMAGE_TAG` should be is worth discussing, currently it is `YYYYMMDD` like
@nevill mentioned.

cc: @krasi-georgiev

geekodour (commented):
I'd like @krasi-georgiev  to review my comments :smile:

.circleci/config.yml
geekodour: `DOCKERFILE_PATH` and `DOCKERBUILD_CONTEXT` were added in https://github.com/prometheus/prometheus/pull/5669

Some usage of `DOCKERBUILD_CONTEXT` was reduced within the prombench repo in https://github.com/prometheus/prombench/pull/316/files

The prombench cli still does make use of it, which can also be made not to use `DOCKERBUILD_CONTEXT` by adding the following,

```Dockerfile
COPY prombench/prombench /usr/bin/prombench
COPY prombench/cmd/prombench/clone.sh /usr/bin/clone.sh
```
to 
https://github.com/prometheus/prombench/blob/8681448850bdc38dcee93d43513cacda710a4d6e/prombench/cmd/prombench/Dockerfile#L12

I've not tested this but it should work, I somehow forgot to add that in #316

tools/commentMonitor/Dockerfile
nevill: What is a multi-staged build? `store_artifact` them, then use it later ? Seems it only applies to `scaler`, `fake-webserver` and `prombench`.

.circleci/config.yml
nevill: I see the target `common-docker` in `Makefile.common`. But it appends `linux` in the middle of image name. I think this project will not create any binary running on OS other than linux, or arch other than amd64, this is so trivial.

tools/commentMonitor/Dockerfile
krasi-georgiev: no multistage is for the docker builds. Basically keeps the final docker image very small because it includes only the binary

https://docs.docker.com/develop/develop-images/multistage-build/

nevill: @simonpasquier suggested using `quay.io/prometheus/busybox:latest` in #218. (It contains some root CAs, so we don't need to copy them when building`fake-webserver`.)

I think the size of image is mainly affected by the binary size and base image size.  Compare the image size of my latest builds docker.io/nevill/{prombench, commentmonitor, amgithubnotifier, fake-webserver, scaler, load-generator} with docker.io/prombench/{prombench, commentmonitor, amgithubnotifier, fake-webserver, scaler, load-generator}, they're almost the same.

nevill: There are two jobs in this circle CI config. The first is `build_and_test`, which will create the binaries, and cache them (by calling `persist_to_workspace`). In the next job `publish`, we just need to COPY the binaries from cache into the image under constructing. That's why every `go build` is removed in `Dockerfile`.

krasi-georgiev: This will remove the posibility to build the docker images locally which is usefull sometimes.

`GO111MODULE=on go build -a -tags netgo -o /go/bin/commentMonitor` would cause downlading the source files of all dependancies so even if doesn't make big difference now it is better to keep the multistage builds. Ofcourse if there is no good reason to have it as a single stage.

> @simonpasquier suggested using quay.io/prometheus/busybox:latest in #218. (It contains some root CAs, so we don't need to copy them when buildingfake-webserver.)

that is a good suggestion so should use it if possible.

.circleci/config.yml
krasi-georgiev: I think it is fine as is, I even perfer that when it is a single command to have it like this instead of going to the makefile to see what is actually run

.circleci/config.yml
```diff
@@ -24,12 +79,48 @@ jobs:
...
+          branches:
             only: /.*/
+    - publish:
+        requires:
+          - build_and_test
+        context: testinfra
+        filters:
+          branches:
+            only:
+              - dockerbuild
```
krasi-georgiev: why do we need this?

.gitignore
```diff
@@ -2,3 +2,4 @@
 /scaler
 /amGithubNotifier
 /commentMonitor
+build/
```
krasi-georgiev: why do we need this?

I usually prefer to put most thing in .git/exclude/info locally instead of maintaining a global gitignore file. Most ignores are local env specific so best to keep them as local configs.

Makefile
```diff
@@ -1,3 +1,10 @@
...
 
-include Makefile.common
\ No newline at end of file
+.PHONY: all
+all: precheck style check_license lint build test unused
+
+include Makefile.common
+
+.PHONY: unused
+unused:
```
krasi-georgiev: Why can't we use the unused command from the commond makefile?

funcbench/.gitignore
```diff
@@ -0,0 +1 @@
+/funcbench
```
krasi-georgiev: can't this be in the global .gitignore file so we maintain only a single file?

prombench/cmd/prombench/Dockerfile
```diff
@@ -1,15 +1,26 @@
 # NOTE: building this requires it to have the build context of the repository root
```
krasi-georgiev, line 1: that is not true anymore right?

.circleci/config.yml
```diff
@@ -11,11 +15,62 @@ executors:
...
+        steps:
+        - run: docker build -t $DOCKER_REPO/$DOCKER_IMAGE:$TAG -f << parameters.dockerfile_path >> << parameters.context_path >>
+    - unless:
+        condition: <<parameters.dockerfile_path>>
+        steps:
+        - run: docker build -t $DOCKER_REPO/$DOCKER_IMAGE:$TAG << parameters.context_path >>
+    - run:
+        name: Push images to registry
+        command: |
+          docker images
```
krasi-georgiev: do we need this?

nevill: This helps for debugging.
To make sure the image is created as we expected, i.e. with the CORRECT image name, tag.

.circleci/config.yml
nevill: Well, I was testing my CircleCI on branch `dockerbuild` instead of `master`. Will remove it.

prombench/cmd/prombench/Dockerfile
nevill: Yes, it's Still true now. If we want to build it locally, have to run `docker build -t prombench:latest -f prombench/cmd/prombench/Dockerfile .` from repository root.

.circleci/config.yml
nevill: Unfortunately, I send this PR from branch `dockerbuild` of my fork, that's why CircleCI check can be passed here. If remove this line, I need to send another new PR.

.gitignore
nevill: Will remove it.

funcbench/.gitignore
nevill: sure, will do it.

.circleci/config.yml
krasi-georgiev: ok

Makefile
nevill: in my environment, it will run `GO111MODULE=on go mod tidy` here, clear go.mod and go.sum, then result the next step `git diff --exit-code -- go.sum go.mod` failed.
I think because we don't have any .go file under root.
`go build` will automatically update go.mod / go.sum, it's safe to run `git diff --exit-code -- go.sum go.mod` here.
`go mod tidy` should be run before commit.

nevill: Oh, found why. It's because of my environment using symbolic link as repository root. `go mod tidy` doesn't follow symbolic link. 
So, I will leave `unused` to `Makefile.common`.

.circleci/config.yml
```diff
@@ -11,11 +15,62 @@ executors:
...
+        default: "."
+    steps:
+    - run:
+        name: Setup environment variables
+        command: |
+          echo "export TAG=`date +%Y%m%d`" >> $BASH_ENV
+          echo "DOCKER_REPO=<< parameters.registry >>/${DOCKER_ORG:-<< parameters.organization >>}" >> $BASH_ENV
+          echo "DOCKER_IMAGE=<< parameters.image_name >>" >> $BASH_ENV
+    - when:
+        condition: <<parameters.dockerfile_path>>
```
simonpasquier: I don't see that `dockerfile_path` isn't always given so I wouldn't have the condition.

.circleci/config.yml
```diff
@@ -11,11 +15,62 @@ executors:
...
+        steps:
+        - run: docker build -t $DOCKER_REPO/$DOCKER_IMAGE:$TAG -f << parameters.dockerfile_path >> << parameters.context_path >>
+    - unless:
+        condition: <<parameters.dockerfile_path>>
```
simonpasquier: same here

.circleci/config.yml
```diff
@@ -11,11 +15,62 @@ executors:
...
+          docker images
+          echo -n $<< parameters.password >> | docker login -u $<< parameters.login >> --password-stdin << parameters.registry >>
+          docker push $DOCKER_REPO/$DOCKER_IMAGE:$TAG
 jobs:
-  test:
+  build_and_test:
     executor: golang
     steps:
-    - prometheus/setup_environment
+    - checkout
```
simonpasquier: why changing this?

.circleci/config.yml
```diff
@@ -11,11 +15,62 @@ executors:
     docker:
     - image: circleci/golang:1.13
 
+commands:
```
simonpasquier: what prevents reusing https://circleci.com/orbs/registry/orb/prometheus/prometheus#commands-publish_images? The `docker` target can be overriden in `Makefile` if we don't need to support multi-arch builds.

simonpasquier (commented):
I'm late for commenting but I'd prefer if we stick with one way of building binaries. That is using `make build` instead of `go build ...` in the Dockefile.

.circleci/config.yml
nevill: Well, everything comes from my comment on [`Makefile.common`](https://github.com/prometheus/prombench/pull/320#discussion_r373179288).

When I discussed with @krasi-georgiev , I felt that we choose to use cli tools like `go build` or `promu` as much as possible instead of `make`. That's why `make docker*` are not used completely here. And if we don't use `make`, orb/prometheus (which contains many `make` steps) can be omitted. Plus, when I was new to the project, given by a big `Makefile.common` + `orb` is so scary, it would be fast to complete without too many dependencies. 😈 

Anyhow, I will try `Makefile` way either. But can you @simonpasquier give me a clue on usage of `Makefile.common` in my previous comment ? Will it be dropped ?

nevill:
@simonpasquier I'm using `orb/promethues` to config the workflow now, please review it again.

.circleci/config.yml
simonpasquier: > Plus, when I was new to the project, given by a big Makefile.common + orb is so scary, it would be fast to complete without too many dependencies. smiling_imp

I agree that it's a bit daunting at first. But they exist for a good reason: they simplify a lot the maintenance work.

> Will it be dropped ?

no the comment you've found predates `Makefile.common`.

Makefile
```diff
@@ -1,3 +1,17 @@
 DOCKER_REPO             ?= prombench
 
-include Makefile.common
\ No newline at end of file
+.PHONY: all docker docker-publish docker-manifest
```
simonpasquier: (nit) I prefer to have the `.PHONY` stanza defined next to the target it applies to.

```
.PHONY: all
all: ...

.PHONY: docker
docker: ...
```

.circleci/config.yml
krasi-georgiev: >  But can you @simonpasquier give me a clue on usage of Makefile.common in my previous comment ? Will it be dropped ?

sorry I should have answered that as well. First the common makefiles was used for building the binaries and it seems that there were some problems with that which led to the development of the promu tool. After that we noticed many common commands for linting and testing go code so decided to share those ammong all repos with yet another common makefile. Now looking at this again it might be a good idea to include as much as possible in the promu tool to reduce the size of the common makefile.

Makefile
nevill: fixed

krasi-georgiev:
> I'm late for commenting but I'd prefer if we stick with one way of building binaries. That is using make build instead of go build ... in the Dockefile.

we had a discussion with @nevill and for now I prefer to keep the Dockerfiles complitely indipendent from promu. Can change later if needed. If you think that this will cause issues please let me know.

.circleci/config.yml
```diff
@@ -24,12 +31,80 @@ jobs:
...
-    - test:
+    - build_and_test:
         filters:
-          tags:
+          branches:
             only: /.*/
+    - publish:
+        requires:
+          - build_and_test
+        context: testinfra
```
nevill: @simonpasquier do you know any predefined context with credentials of prombench@docker.io ?

nevill: some hints: the image created here should be pushed under `docker.io/prombench` instead of `docker.io/prom`, context variable DOCKER_LOGIN / DOCKER_PASSWORD should be set accordingly.

geekodour:
@nevill did any of the changes affect how the images are built manually? in that case, i think updating individual `README.md`s will be a good addition to this PR.

eg. https://github.com/prometheus/prombench/blob/master/prombench/cmd/prombench/README.md

nevill:
@geekodour READMEs are updated

funcbench/Dockerfile
```diff
@@ -1,21 +1,16 @@
...
-WORKDIR /
-COPY . /
+COPY funcbench/*.go ./
+COPY go.* ./
 
-RUN CGO_ENABLED=0 go build main.go
+RUN CGO_ENABLED=0 GO111MODULE=on go build -a -tags netgo -o /go/bin/funcbench
 
-FROM golang:alpine
+FROM golang:1.13-alpine
```
simonpasquier, line 10: `quay.io/prometheus/busybox:latest`?

simonpasquier: ok  scratch that, I missed that you need `git`.

simonpasquier:
> If you think that this will cause issues please let me know.

I see no technical problem. It's not consistent compared to what the other repositories are doing but this can be improved later.

simonpasquier:
LGTM

funcbench/Dockerfile
nevill: Yes, this is another place we can improve, e.g. use `go-git` to save `git` cli.

PR #325, opened 2020-02-21
funcbench: Allow running locally on existing repo.

bwplotka:
This will allow reusing funcbench for literally any Go projects including [Thanos](http://thanos.io/) and [Cortex](https://github.com/cortexproject/cortex) :hugs: 

##  Changes:
* CLI mode
* Ability to use commit
* Bit of refactoring
* Placeholder for Sub-benchmarks mode: TODO and will do in other PR.
* Moving to git library
* Using worktrees
* Ability to cache local results.

Signed-off-by: Bartlomiej Plotka <bwplotka@gmail.com>


geekodour:
@krasi-georgiev operations of funcbench will be done using comment-monitor as well, no? Currently the trigger is `issue_comment`.

@bwplotka  modifying how funcbench will be run in Github Actions is not part of this PR right?

bwplotka:
That's correct. I adjusted it for local run, tried to not changed how GitHub action part works (:

roidelapluie:
You should create a git worktree to avoid the clean state requirement.

bwplotka:
I don't know if I **should** but that's a good idea.

geekodour:
@bwplotka  think the failing tests are due to missing headers in `bench.go` and `env.go`

bwplotka:
Yea, it's not fully complete - idea to move to workdirs sounds good, would like to experiment with it.

funcbench/env.go
```diff
@@ -0,0 +1,224 @@
...
+		return nil, err
+	}
+
+	email, err := c.exec("git", "config", "--get", "user.email")
+	if err != nil {
+		return nil, err
+	}
+
+	if email != "" {
+		return nil, errors.Errorf("expected fresh machine, got git user configured: %s", email)
```
krasi-georgiev: why is this check?

krasi-georgiev (commented):
@geekodour , @nevill 

Since we are not gong to use GH actions, I am wondering if can even avoid parsing the event.json file. I think prombench doesn't parse the event file and we should probably to the same here to simplify the logic. @geekodour , @nevill  Could you have a look and confirm if we could do that?

UPDATE:
> @krasi-georgiev operations of funcbench will be done using comment-monitor as well, no? Currently the trigger is issue_comment.

Sorry I just saw the other comments and yes lets confirm if we can avoid parsing the event.json file. This will simplify the workflow quite a bit.

nevill:
I think it's good to use params to pass the options ( pull request number / branch name / regex, etc ) instead of files. In fact, I did it on my local branch.

nevill (changes requested):

krasi-georgiev:
> I think it's good to use params to pass the options ( pull request number / branch name / regex, etc ) instead of files. In fact, I did it on my local branch.

Yes that is true, main qustion is if this would be possible with the comment monitor and the GH actions.

geekodour:
@nevill worked on removing event.json in #311 , should I make another PR just to remove `event.json` and use `repository_dispatch` arguments for that before merging this one? 

cc: @krasi-georgiev

funcbench/env.go
bwplotka: To not mess up your own git config - I am running this locally.

bwplotka:
@geekodour Feel free to do, I will then adapt my PR.

Let's keep things focused. I don't want to break anything on GH action side, I want it to be usable and clear for CLI use. Looks we are all on board on this. Can we make sure there is NO magic config file unless we need secrets? It's hard to discover, so flags/args would be preferred.

> Since we are not going to use GH actions, I am wondering if can even avoid parsing the event.json file. 

@krasi-georgiev so... what's the plan? Can we keep this discussion public? We have now new CNCF slack channel `#benchmarking`

krasi-georgiev:
> Since we are not going to use GH actions, I am wondering if can even avoid parsing the event.json file.

that comment was misleading, we are still using GH actions just in a slightly different way. First all GH hooks are processed by another component called `commentMonitor` which runs 24/7 on GKE and that when the GH comment is the expected format the `commentMonitor` triggers a GH action. We decided to do it this way because triggering GH action directly from a comment was slow(each time it downloads a container) and also each comment would trigger a GH action(this flooded the GH actions UI with lots of triggered test) which would then check the format of the comment to decide whether it needs to run any additional steps.

geekodour:
> @geekodour Feel free to do, I will then adapt my PR.

@bwplotka on it.

bwplotka:
Thanks all for clarifications. Let's continue this then. Will notify you
when my PR will be fully ready (need to finish subtest comparing)

bwplotka:
Example usage: `funcbench --result-cache "_dev/funcbench" -l -v --no-race -t 30s master BenchmarkSeries` (Will put it in README)

nevill:
> Example usage: `funcbench --result-cache "_dev/funcbench" -l -v --no-race -t 30s master BenchmarkSeries` (Will put it in README)

Currently CGO is disabled, we cannot really do `go test -race`. So here `--no-race` here can be removed.

go.mod
```diff
@@ -9,18 +9,27 @@ require (
...
 	golang.org/x/oauth2 v0.0.0-20190226205417-e64efc72b421
+	golang.org/x/sys v0.0.0-20200223170610-d5e6a3e2c0ae // indirect
+	golang.org/x/tools v0.0.0-20190729092621-ff9f1409240a
 	google.golang.org/api v0.5.0
 	google.golang.org/genproto v0.0.0-20190516172635-bb713bdc0e52
 	google.golang.org/grpc v1.19.0
 	gopkg.in/alecthomas/kingpin.v2 v2.2.6
 	gopkg.in/inf.v0 v0.9.1 // indirect
-	gopkg.in/yaml.v2 v2.2.2
+	gopkg.in/src-d/go-git.v4 v4.13.1
```
nevill, line 32: Great to see `go-git` is finally used.

bwplotka:
This PR is actually almost ready to go, just one thing missing. 

`go-git` is not perfect. I don't really like their API. It's really confusing and missing basic functionalities (: But... :man_shrugging: Nothing better.

krasi-georgiev:
> Currently CGO is disabled, we cannot really do go test -race. So here --no-race here can be removed.

hm we should enable it than, we want to catch race conditions in the bench tests

nevill:
CGO is disabled since #275 . Can benchmark and race detection be separate ?

bwplotka:
Yes, I think we should 100% not detect race while benchmarking. We should have unit tests for that. I would remove it.

bwplotka:
Also @nevill we created a lib in Thanos: https://github.com/thanos-io/thanos/blob/master/pkg/testutil/testorbench.go which allows to run benchmark test as test as well, so we can reuse our tests :heart: This is where I would use race detection. Even as a separate GH action. 

cc @krasi-georgiev

bwplotka:
This should be ready to go. Subbenchmark Comparision is **not implemented** for now and I would vote for merging like this to not increase this PR even more (: 

PTAL @nevill @krasi-georgiev @geekodour :hugs: 

I use this heavily locally and plan to add to Thanos as GithubAction.

bwplotka:
Used this version (local mode) in Thanos: https://github.com/thanos-io/thanos/pull/2233

I wonder if we really need this pr comment based magic approach.... I think local mode is good enough. I might vote to remove the special pr id flow even at some point.

In https://github.com/thanos-io/thanos/pull/2233 we execute all micro benchmarks for every PR, every push to master and every pushed tag (with cache) and compare in a smart way. No need to any manual interaction. 

If that will happen to just work nicely for Thanos, I will propose the same to Prometheus.

bwplotka:
I guess some on demand action with result on pr comment might still make sense for certain tests against certain branch 🤔

roidelapluie:
Have to considered the ecological impact on "wasting" benchmark cpu time on every pr?

roidelapluie:
> Have to considered the ecological impact on "wasting" benchmark cpu time on every pr?

[[ Note that I don't intend to launch a big debate here ; my point is that it is probably to keep that manually triggered to avoid waste; even if at that scale that will not save the planet. ]]

bwplotka:
We never optimized for this so I could argue then why we run CI for every
PR for unrelated code branches, why we cross build and why we build website
on every commit. Still ecology is a valid point. Would be nice to have
pointers from GitHub itself. What If they have those machines running
anyway? Then there is almost no difference so we are quite blind here. I
would say starting bigger discussion about it, discussion that CI provider
will join as well would be a good direction before micro energy optimizing
here. But definitely good point you raised, thanks.


Anyway it's up to us how we configure GitHub actions for this, so
discussion is not relevant for this PR.

krasi-georgiev:
> In thanos-io/thanos#2233 we execute all micro benchmarks for every PR, every push to master and every pushed tag (with cache) and compare in a smart way. No need to any manual interaction.

this will be expencive and usually not needed for every PR.
You still need a GKE node to make sure you get consistent results so I do you think Thanos would be sponsored to run those tests for every PR and every commit to a PR? I certainly know that for Prometheus that would cost too much and like @roidelapluie mentioned just wasting resources.

bwplotka:
Anyway, this discussion does not influence this PR, so it's ready to go.

krasi-georgiev:
thanks, I will let @geekodour do the final review.

geekodour:
> I wonder if we really need this pr comment based magic approach.... I think local mode is good enough. I might vote to remove the special pr id flow even at some point.

There will be some changes introduced to make the PR flow simpler. Also, the GitHub related parts will need changes as well as quite a few changes in the README, but it'll be good if we don't include those in this PR. The refactorings will be similar to #336 so, will be able to add those changes once this PR is merged.

Just to confirm, as per comments in https://github.com/thanos-io/thanos/pull/2233 we will need manual intervention right? and funcbench would be run in GKE for now.

cc: @krasi-georgiev @bwplotka

krasi-georgiev:
yes we leave the original idea to triger tests with a gh comment and the bench tests will run in GKE than the results will be posted back to github

bwplotka:
> There will be some changes introduced to make the PR flow simpler. Also, the GitHub related parts will need changes as well as quite a few changes in the README, but it'll be good if we don't include those in this PR. The refactorings will be similar to #336 so, will be able to add those changes once this PR is merged.

:+1: 

Yes. the https://github.com/thanos-io/thanos/pull/2233 was experiment from Thanos side, looks like from Prometheus side the idea of just launching manually is the way to go for now. This PR is agnostic really to this decision (:

krasi-georgiev:
> Yes. the thanos-io/thanos#2233 was experiment from Thanos side, looks like from Prometheus side the idea of just launching manually is the way to go for now. This PR allows both (:

but what about the misleading results for the ns/op when you run those on gh actions?

actually shall we try to run another one on gh actions with a longer running bench test again just to confirm that gh actions is not reliable for cpu tests @nevill what do you think?

bwplotka:
> but what about the misleading results for the ns/op when you run those on gh actions?

What if we can specify machine size, then those ns/op will make sense. Also, we can hide those `ns/op` for this mode or something. (:

bwplotka:
Anyway I will repeat 3rd time, This PR is not the best place to discuss this as well as ecology and global warming @krasi-georgiev xD

nevill:
> actually shall we try to run another one on gh actions with a longer running bench test again just to confirm that gh actions is not reliable for cpu tests @nevill what do you think?

I don't think it's needed.
As @geekodour mentioned above, `funcbench` will be run in a GKE node, that's the target, please move on to it.

funcbench/bench.go
```diff
@@ -0,0 +1,142 @@
...
+
+	if _, err := ioutil.ReadFile(filepath.Join(b.resultCacheDir, fileName)); err == nil {
+		fmt.Println("Found previous results for ", fileName, b.benchFunc, "Reusing.")
+		return filepath.Join(b.resultCacheDir, fileName), nil
+	}
+
+	// TODO(bwplotka): Allow memprofiles.
+	extraArgs := []string{"-benchtime", b.benchTime.String()}
+	extraArgs = append(extraArgs, "-timeout", b.benchTimeout.String())
+	benchCmd := []string{"bash", "-c", strings.Join(append(append([]string{"cd", pkgRoot, "&&", "go", "test", "-run", "^$", "-bench", fmt.Sprintf("^%s$", b.benchFunc), "-benchmem"}, extraArgs...), "./..."), " ")}
```
nevill, line 84: The real command to be executed is difficult to know, can we add a log here to print the whole command in verbose mode? or do
```go
cmdStr := "cd /some/dir && go test -run . -bench ."
benchCmd := string.Split(cmdStr)
```
probably `func (c *commander) exec` is a better place to log.

funcbench/bench.go
```diff
@@ -0,0 +1,142 @@
...
+		return nil, errors.Wrapf(err, "open %s", beforeFile)
+	}
+
+	after, err := parseFile(afterFile)
+	if err != nil {
+		return nil, errors.Wrapf(err, "open %s", afterFile)
+	}
+
+	// Use benchcmp library directly.
+	// TODO(bwplotka): benchstat is new thing - we might add choice for funcbench to choose from? (:
```
nevill, line 127: Surely `benchcmp` will be dropped. I prefer `benchstat`.

funcbench/env.go
```diff
@@ -0,0 +1,229 @@
...
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"github.com/google/go-github/github"
```
nevill, line 24: We will use `github.com/google/go-github/v29/github`.

funcbench/bench.go
bwplotka: We have the command logged below

funcbench/bench.go
bwplotka: I would not be that sure, let's decide later.

funcbench/env.go
bwplotka: When it was decided and why? (:

geekodour: @nevill let's keep those changes for another pr like I mentioned in my previous comment. This pr is already is adding a lot of things.

geekodour: @bwplotka one of the tools (commentMonitor) was using the `repository_dispatch` event, v29 adds official support for it in the api so we would not want to use multiple versions of go-github here. So it will be just a refactoring change, would not really make it part of this pr if not really required (previous comment)

bwplotka: I see, if we could do this in separate PR it would be awesome :heart:

bwplotka:
Merging as discussed offline with @krasi-georgiev on #benchmark channel (:

PR #330, opened 2020-02-24
Add new input params org / repo / branch / prNumber 

nevill:
It will drop the parameter `-input`.
This is actually part of #311 , and be picked up as discussed in #325 .

Signed-off-by: Nevill <nevill.dutt@gmail.com>

### DIFF
--- a/funcbench/main.go
+++ b/funcbench/main.go
@@ -38,50 +38,60 @@ type gitHubClient struct {
 	repo             string
 	latestCommitHash string
 	prNumber         int
+	comment          bool
 	client           *github.Client
 }
 
 type benchmarkTester struct {
-	bechRegex       string
-	home            string
-	raceFlagEnabled bool
+	bechRegex string
+	home      string
 }
 
 func main() {
 	// Show file line with each log.
 	log.SetFlags(log.Ltime | log.Lshortfile)
-	app := kingpin.New(filepath.Base(os.Args[0]), "benchmark result posting and formating tool.\n-i location of github hook file (even.json)")
+	app := kingpin.New(
+		filepath.Base(os.Args[0]),
+		"benchmark result posting and formating tool.",
+	)
 	app.HelpFlag.Short('h')
-	input := app.Flag("input", "path to event.json").Short('i').Default("/github/workflow/event.json").String()
+
+	owner := app.Flag("owner", "It's a Github owner or organisation name.").Default("prometheus").Short('o').String()
+	repo := app.Flag("repo", "This is the repository name.").Default("prometheus").Short('r').String()
+
+	prNumber := app.Flag("pullrequest", "It's a new pull request ID, the benchmark will be run using a copy of its code.").Short('p').Required().Int()
+	branch := app.Arg("branch", "A name of branch, will be used as a baseline in benchmark.").Required().String()
+	regex := app.Arg("regex", "A regular expression will be used in '-bench regexp' during 'go test', check 'go help testflag' for details.").Default(".").String()
+
+	postComment := app.Flag("comment", "If set to false, it will not post the benchmark result back to Github. This is mainly for local testing.").Default("true").Bool()
+
 	kingpin.MustParse(app.Parse(os.Args[1:]))
 
-	data, err := ioutil.ReadFile(*input)
-	if err != nil {
-		log.Fatalln(err)
-	}
+	log.Printf(
+		"Start testing %s/%s PR-%d VS branch %s, on benchmark `%s`. With comment: %v.\n",
+		*owner, *repo, *prNumber, *branch, *regex, *postComment,
+	)
 
-	event, err := github.ParseWebHook("issue_comment", data)
-	if err != nil {
-		log.Fatalln(err)
-	}
 	var (
 		ghClient  *gitHubClient
 		gitClient *gitClient
-		prNumber  int
 	)
-	switch eventType := event.(type) {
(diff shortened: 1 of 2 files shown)


geekodour:
think we can also get rid of https://github.com/prometheus/test-infra/blob/5f83f79e86a7180ef3d217da67f6313f54c5426a/funcbench/main.go#L88

Since it does no longer matter if we're running `funcbench` inside a container inside github actions, think we should get rid of use of `GITHUB_WORKSPACE` env var and replace it with a more generic name.

nevill:
> think we can also get rid of
> 
> https://github.com/prometheus/test-infra/blob/5f83f79e86a7180ef3d217da67f6313f54c5426a/funcbench/main.go#L88

@geekodour are you referring to the wrong line ? `logLink` is used with error message, which is needed there.
If you refer to `GITHUB_WORKSPACE`, I think we need to keep it. Better rename to `GIT_WORKSAPCE` or so. But in this PR, let's focus on the input parameters.

geekodour:
@nevill yes I was referring to `logLink` since funcbench will be run in GKE from Github Actions, so the logs will most probably end up in loki if we wish to persist the logs. That also means funcbench could do without `latestCommitHash` but let's keep removing those changes for another PR.

> If you refer to GITHUB_WORKSPACE, I think we need to keep it. Better rename to GIT_WORKSAPCE or so. But in this PR, let's focus on the input parameters.

Yes, let's keep this PR smaller then. Since @bwplotka used `GITHUB_WORKSPACE` env var in #325 I thought adding the change here will also add the changes in #325 , but we can refactor that later aswell. no problem.

Otherwise this looks good to me.

cc: @krasi-georgiev

funcbench/main.go
```diff
@@ -50,38 +51,49 @@ type benchmarkTester struct {
...
 	log.SetFlags(log.Ltime | log.Lshortfile)
-	app := kingpin.New(filepath.Base(os.Args[0]), "benchmark result posting and formating tool.\n-i location of github hook file (even.json)")
+	app := kingpin.New(
+		filepath.Base(os.Args[0]),
+		"benchmark result posting and formating tool.",
+	)
 	app.HelpFlag.Short('h')
-	input := app.Flag("input", "path to event.json").Short('i').Default("/github/workflow/event.json").String()
+
+	owner := app.Flag("owner", "github owner or organisation name").Default("prometheus").Short('o').String()
```
bwplotka: Let's start having help in full sentence and below.

funcbench/main.go
```diff
@@ -50,38 +51,49 @@ type benchmarkTester struct {
...
+	repo := app.Flag("repo", "github repository name").Default("prometheus").Short('r').String()
+
+	prNumber := app.Flag("pr", "pull request ID").Short('p').Required().Int()
```
bwplotka: Let's not be lazy and name it `pull-request`.

Why we need this though?

bwplotka: What about just `initial-branch`? You can put then `pulls/xxx/head` etc..

funcbench/main.go
```diff
@@ -50,38 +51,49 @@ type benchmarkTester struct {
...
+	branch := app.Arg("branch", "branch name to test against").Required().String()
+	regex := app.Arg("regex", "regex to specify which benchmarks to run").Default(".").String()
+
+	raceEnabled := app.Flag("race", "branch name to test against").Default("false").Bool()
```
bwplotka: Help is copy-pasta. Also if it does not work let's remove it, maybe?

funcbench/main.go
```diff
@@ -50,38 +51,49 @@ type benchmarkTester struct {
...
+	postComment := app.Flag("comment", "if post result as comment to original pull request, mainly for debugging").Default("true").Bool()
```
bwplotka: I don't get this help ): What exactly this option does?

funcbench/main.go
```diff
@@ -220,30 +223,22 @@ func (c *gitClient) revertPRChanges() error {
...
-	}
-	raceArgument, err := ioutil.ReadFile("/github/home/commentMonitor/RACE")
-	if err != nil {
+	if err := os.Setenv("CGO_ENABLED", "0"); err != nil {
 		return nil, err
 	}
 
 	bench := benchmarkTester{
-		raceFlagEnabled: string(raceArgument) != "-no-race",
+		raceFlagEnabled: raceEnabled,
```
bwplotka:
```suggestion
		isRaceEnabled: isRaceEnabled,
```

bwplotka (commented):
Nice, just some comments.

funcbench/main.go
geekodour: maybe a `--dry-run` flag can substitute this case better?

funcbench/main.go
nevill: Sorry, what is `initial-branch`? Like in this PR, the original branch name of my fork is `funcbench-args`, so initial-branch is `funcbench-args`? I assume that `xxx` in `pulls/xxx/head` are always digits.

funcbench/main.go
nevill: OK, this param will be removed.

bwplotka (approved):
Awesome, let's merge this, so I can rebase and finish the refactor (: 


Thanks!

funcbench/main.go
bwplotka: Let's merge like this, I will show on my PR what I meant (:

PR #337, opened 2020-03-01
Added make back to prombench Dockerfile.

geekodour:
Installing `make` is essential for running the prombench docker image since the base alpine image does not have it by default. It was removed in #320 and got overlooked, this PR adds it back.

cc: @krasi-georgiev @nevill 

Signed-off-by: Hrishikesh Barman <plain.hrishikeshbman@gmail.com>


nevill:
Oh, `promethues-builder` requires it.

PR #342, opened 2020-03-17
Fix for funcbench

nevill:
Globally update go-github to v29 since #336.
Fix the checkout process of funcbench under Github environment.

@geekodour @bwplotka please review it.


bwplotka (approved):
LGTM,

How you verified this?

funcbench/env.go
```diff
@@ -98,14 +99,16 @@ type GitHub struct {
 }
 
 func newGitHubEnv(ctx context.Context, e environment, owner, repo string, prNumber int) (Environment, error) {
-	if err := os.Chdir(os.Getenv("GITHUB_WORKSPACE")); err != nil {
+	rpath := os.Getenv("GITHUB_WORKSPACE")
```
geekodour: Can we change this env var name to something more appropriate since we'll be using Github Actions just to launch funcbench and not to actually run it?

geekodour (commented):
I was working on a similar PR with a few more changes. But if these changes are working as expected, we can add these changes.

~I was just wondering if we could pull out the GitHub commenting part and just pass the local output into some env var and use the ~Github Actions~ `prombench` manifest file for `funcbench` to post the output comment if required. I don't understand why to handle things differently for GitHub and local. Is there something I am missing? (having the comment functionality inside funcbench will make it cleaner though, just a thought)~

EDIT: (No longer wondering)

Also, I was trying to run funcbench locally and where I am getting this error:

```
funcbech02:44:44 main.go:47: error creating SSH agent: "SSH agent requested but SSH_AUTH_SOCK not-specified"compareTargetRef
```
This requires me to setup my ssh-agent to proceed, is this a required step or something else? o.O

cc: @bwplotka @nevill

nevill:
> LGTM,
> 
> How you verified this?

do `export WORKSPACE=/tmp` before running, I updated README to add this.

funcbench/env.go
nevill: Will rename to `WORKSPACE`, and change to:
1. try to read from cli flag `workspace` if set
2. try to read env `WORKSPACE` if set
3. try to use current directory

I think at some point we will merge with local mode 😈

funcbench/env.go
```diff
@@ -98,14 +99,23 @@ type GitHub struct {
 }
 
 func newGitHubEnv(ctx context.Context, e environment, owner, repo string, prNumber int) (Environment, error) {
-	if err := os.Chdir(os.Getenv("GITHUB_WORKSPACE")); err != nil {
-		return nil, errors.Wrap(err, "changing to GITHUB_WORKSPACE dir")
+	if e.home == "" {
+		e.home = os.Getenv("WORKSPACE")
```
geekodour: I think you can just move this out of the if block?

nevill: Update it.
Basically it reads from `-w` flag first, then environment variable `WORKSAPCE`, in the end, set to current directory.

funcbench/main.go
```diff
@@ -80,6 +80,7 @@ The special syntax Nx means to run the benchmark N times
 If d is 0, the timeout is disabled.
 The default is 2 hours (2h).`).Default("2h").Duration()
 	resultCacheDir := app.Flag("result-cache", "Directory to store output for given func name and commit sha. Useful for local runs.").String()
+	workspace := app.Flag("workspace", "A directory where source code will be cloned to.").Short('w').String()
```
geekodour: > Basically it reads from -w flag first, then environment variable WORKSAPCE, in the end, set to the current directory.

You can use the `.Default` here no?

Furthermore, think we should be using one way to set the workspace. But let's go forward with this change as we've not written any tests for this yet and will remove the env var eventually.

nevill: Yes, udpated.

PR #266, opened 2019-08-22
[RDY] Moved prombench to subdirectory prombench

geekodour:
See #257

PR #343, opened 2020-03-25
Funcbench on GKE

nevill:
This is a replacement of #311 :)

You can check [my repository](https://github.com/nevill/prometheus/blob/master/.github/workflows/funcbench.yml) to see the corresponding workflow file.

Currently, the whole process takes over 70 minutes, but the automatically created token only can [last 60 minutes](https://help.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token#about-the-github_token-secret), so we have to use personal token here. In future, a better solution would be to send the benchmark result via commentMonitor.

@krasi-georgiev @geekodour Please review it.


funcbench/manifests/benchmark/2_job.yaml
```diff
@@ -0,0 +1,35 @@
...
+  backoffLimit: 0
+  template:
+    spec:
+      restartPolicy: Never
+      containers:
+      - name: funcbench
+        image: nevill/funcbench:20200324
+        imagePullPolicy: Always
+        args:
+          - "-v"
```
geekodour: I think passing the full flag names for funcbench here will improve readability.

funcbench/manifests/benchmark/2_job.yaml
```diff
@@ -0,0 +1,35 @@
...
+          - "-p"
+          - "{{ .PR_NUMBER }}"
+          - "-w"
+          - "/build"
+          - "{{ .BRANCH }}"
+          - "{{ .REGEX }}"
+        env:
+          - name: GITHUB_TOKEN
+            value: "{{ .GITHUB_TOKEN }}"
+          - name: GITHUB_SHA
```
geekodour: I am removing the use of `GITHUB_SHA` in an upcoming PR as it is not of any use inside funcbench, we can either wait for it to get merged or merge this before.

geekodour (commented):
I think the file location for the manifest files should be:

`funcbench/manifests/benchmark/1_namespace.yaml` -> `prombench/manifest/funcbench/benchmark/1_namespace.yaml` , similar for others.

and `make` targets can be added to `/prombench/Makefile` with appropriate funcbench specific target names.

funcbench/manifests/benchmark/2_job.yaml
geekodour: also please change the name of the image to  ` docker.io/prominfra/funcbench:master` from `nevill/funcbench:20200324`

nevill: Probably I need another PR to change the image name ? Otherwise you cannot really test my workflow. Because `prominfra/funcbench:master` is not sync with this PR. When `prombench` executes the manifest, it cannot find the appropriate docker image.

funcbench/manifests/benchmark/2_job.yaml
nevill: Either way is fine. Just to mention that, `GITHUB_SHA` is still used in [funcbench](https://github.com/prometheus/test-infra/blob/master/funcbench/env.go#L129) now.

nevill:
> and `make` targets can be added to `/prombench/Makefile` with appropriate funcbench specific target names.

@geekodour You mean to add a new target like `make funcbench_resource_apply` here ?

geekodour:
> You mean to add a new target like make funcbench_resource_apply here ?

yes. it'll simplify the github action workflow aswell.

funcbench/manifests/benchmark/2_job.yaml
geekodour: okay we keep it this way for testing now then, just a reminder to change it to the correct name later in the same PR before merging.

nevill:
Updated Makefile and moved manifest files.

geekodour:
@nevill can you update the README for funcbench in #343 with manual instructions to run funcbench on gke like this: 
https://github.com/prometheus/test-infra/tree/master/prombench#start-a-benchmarking-test-manually

funcbench/README.md
```diff
@@ -45,53 +50,67 @@ Errors out if there are no sub-benchmarks.
...
-
-This tool is meant to be used with a Github action. The action itself is, to a large degree, unusable alone, as you need
-to combine it with another Github action that will provide necessary files to it. At this time, the only action it is
-supposed to work with, is [comment-monitor](https://github.com/prometheus/prombench/tree/master/tools/commentMonitor).
+        BRANCH: ${{ github.event.client_payload.BRANCH }}
+        GITHUB_ORG: prominfra
+        GITHUB_REPO: prometheus
+        REGEX: ${{ github.event.client_payload.REGEX }}
+      with:
+        args: make funcbench_resource_apply
```
geekodour, line 69: Would this not need deleting the funcbench namespace or deleting the nodepool afterwards?

nevill: Added more steps below to recycle them.

geekodour:
I think it'll be good to persist the job for sometime after funcbench finishes so that funcbench users are able to see the logs if needed. 

https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/#ttl-mechanism-for-finished-jobs

currently gke jobs logs will probably be only available via stackdriver right?

funcbench/README.md
```diff
@@ -1,17 +1,22 @@
...
+> By default all benchmarks run without `-race` flag (#275).
 
 The benchmark can be triggered by creating a comment which specifies a branch to compare. The results are then posted back as a PR comment.
 The benchmark can be also trigger and used as CLI command, without GitHub hook.
 
 Comparison is done using [benchcmp](https://godoc.org/golang.org/x/tools/cmd/benchcmp).
-Arguments for the benchcmp are read from files created by previous action (for example [commentMonitor](/tools/commentMonitor/main.go)),
-which is responsible for the comment parsing.
+
+Arguments for the benchcmp are read from files created by previous action (for example [commentMonitor](/tools/commentMonitor/main.go)), which is responsible for the comment parsing.
```
geekodour: > Arguments for the benchcmp are read from files created by previous action

We're no longer doing this, we're getting the arguments from the `repository_dispatch` event.

prombench/Makefile
```diff
@@ -45,3 +45,28 @@ all_nodepools_deleted:
...
+	$(PROMBENCH_CMD) gke nodepool delete -a ${AUTH_FILE} \
+		-v ZONE:${ZONE} -v PROJECT_ID:${PROJECT_ID} -v CLUSTER_NAME:${CLUSTER_NAME} -v PR_NUMBER:${PR_NUMBER} \
+		-f manifests/funcbench/nodepools.yaml
+
+funcbench_resource_apply:
+	$(PROMBENCH_CMD) gke resource apply -a ${AUTH_FILE} \
+		-v ZONE:${ZONE} -v PROJECT_ID:${PROJECT_ID} -v CLUSTER_NAME:${CLUSTER_NAME} \
+		-v PR_NUMBER:${PR_NUMBER} -v GITHUB_TOKEN:${GITHUB_TOKEN} \
+		-v GITHUB_ORG:${GITHUB_ORG} -v GITHUB_REPO:${GITHUB_REPO} \
+		-v LAST_COMMIT_SHA:${LAST_COMMIT_SHA} -v BRANCH:${BRANCH} -v REGEX:${REGEX} \
```
geekodour: Should we change `BRANCH` to `TARGET` ?

prombench/manifests/funcbench/benchmark/2_job.yaml
```diff
@@ -0,0 +1,35 @@
...
+  namespace: funcbench-{{ .PR_NUMBER }}
+spec:
+  # never re-create a new pod
+  backoffLimit: 0
+  template:
+    spec:
+      restartPolicy: Never
+      containers:
+      - name: funcbench
+        image: docker.io/nevill/funcbench:20200324
```
geekodour: docker.io/prominfra/funcbench:master

geekodour (commented):
Some minor changes.

nevill:
> I think it'll be good to persist the job for sometime after funcbench finishes so that funcbench users are able to see the logs if needed.
> 
> https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/#ttl-mechanism-for-finished-jobs
> 
> currently gke jobs logs will probably be only available via stackdriver right?

I discussed with @krasi-georgiev about it some time ago. Currently we just remove everything, including node pools, otherwise, we have to wait until job gets recycled, then remove node pools, which includes more steps. We can add a flag to set if we keep the job and node pool for debugging, but this can be a new feature.

prombench/Makefile
nevill: Not sure about it. Some places (like, in `REAMD.md`) are still calling it `branch`. I think we use branch most of the time in the comments, let's keep it like this first.

geekodour: we can get rid of `LAST_COMMIT_SHA` variable now.

funcbench/README.md
```diff
@@ -28,8 +28,56 @@ The benchmark can be triggered by creating a comment in a PR which specifies a b
...
 
+#### Setup
+
 The comment is handled by [comment-monitor](https://github.com/prometheus/test-infra/tree/master/tools/commentMonitor) and then the parsed arguments are handed over to funcbench(if using Github Actions) or to [prombench](https://github.com/prometheus/test-infra/tree/master/prombench) if using funcbench with GKE.
 
+- Create GitHub actions workflow file (see below) that is executed when an `repository_dispatch` event is on.
+- Also add GKE auth info (PROJECT_ID / ZONE / CLUSTER_NAME) as secrets
+- Read LAST_COMMIT_SHA / BRANCH / PR_NUMBER / REGEX from event payload into environment variables
+- Ignore this if your benchmark lasts less than 60 minutes.
+  - Provide a GitHub token as a secret, here named `PERSONAL_TOKEN`, to avoid conflicts with default `GITHUB_TOKEN`.
```
geekodour: I don't see any second mention of `PERSONAL_TOKEN`. Is something missing?

funcbench/README.md
```diff
@@ -28,8 +28,56 @@ The benchmark can be triggered by creating a comment in a PR which specifies a b
 /funcbench <branch/commit> <benchmark function regex>
 ```
 
+#### Setup
+
 The comment is handled by [comment-monitor](https://github.com/prometheus/test-infra/tree/master/tools/commentMonitor) and then the parsed arguments are handed over to funcbench(if using Github Actions) or to [prombench](https://github.com/prometheus/test-infra/tree/master/prombench) if using funcbench with GKE.
 
+- Create GitHub actions workflow file (see below) that is executed when an `repository_dispatch` event is on.
```
geekodour, line 35: Just a reminder: Once we finalize on the workflow with GitHub statuses and action failure handling, I think linking this section to the workflow directly will be a good idea.

geekodour (commented):
Would like @krasi-georgiev to take a final look.

funcbench/README.md
nevill: No, it's only set in the workflow file.

funcbench/README.md
```diff
@@ -28,8 +28,53 @@ The benchmark can be triggered by creating a comment in a PR which specifies a b
...
+    name: Running funcbench
+    if: github.event.action == 'funcbench_start'
+    runs-on: ubuntu-latest
+    env:
+      AUTH_FILE: ${{ secrets.PROMBENCH_GKE_AUTH }}
+      CLUSTER_NAME: << cluster name >>
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      PROJECT_ID: << project id >>
+      PR_NUMBER: ${{ github.event.client_payload.PR_NUMBER }}
+      TEST_INFRA_REPO: https://github.com/prometheus/test-infra.git # put your repo here if test your own test-infra
```
krasi-georgiev: Why do we need this link?

funcbench/README.md
```diff
@@ -28,8 +28,53 @@ The benchmark can be triggered by creating a comment in a PR which specifies a b
...
+      uses: docker://prominfra/prombench:latest
+      with:
+        args: make funcbench_nodepool_create
+    - name: Run funcbench
+      uses: docker://prominfra/prombench:latest
+      env:
+        BRANCH: ${{ github.event.client_payload.BRANCH }}
+        GITHUB_ORG: prominfra
+        GITHUB_REPO: prometheus
+        GITHUB_TOKEN: ${{ secrets.PERSONAL_TOKEN }} # if benchmark lasts more than 60 minutes, have to create your own token, and put it into PERSONAL_TOKEN
```
```suggestion
        GITHUB_TOKEN: ${{ secrets.PERSONAL_TOKEN }} # The GH action token lasts up to 60min so using PERSONAL_TOKEN guarantees that can post back the results even when the bench tests takes longer.
```

funcbench/README.md
```diff
@@ -28,8 +28,53 @@ The benchmark can be triggered by creating a comment in a PR which specifies a b
...
+        REGEX: ${{ github.event.client_payload.REGEX }}
```
krasi-georgiev: hm, why not provide the regex directly here? It will be easier to change and understand.

funcbench/README.md
```diff
@@ -28,8 +28,53 @@ The benchmark can be triggered by creating a comment in a PR which specifies a b
...
+      with:
+        args: make funcbench_resource_apply
+    - name: Recycle all
+      if: always()
```
krasi-georgiev: is that nececary?

pkg/provider/k8s/k8s.go
```diff
@@ -179,6 +180,10 @@ func (c *K8s) ResourceApply(deployments []Resource) error {
 				err = c.customResourceApply(resource)
 			case "statefulset":
 				err = c.statefulSetApply(resource)
+			case "job":
+				err = c.jobApply(resource)
+				// Ignore the error as even when job deletion fails, it will be removed together with nodepool.
+				_ = c.jobDelete(resource)
```
krasi-georgiev: Why is this delete necessary?

prombench/manifests/funcbench/benchmark/2_job.yaml
```diff
@@ -0,0 +1,33 @@
...
+        imagePullPolicy: Always
+        args:
+          - "--verbose"
+          - "--owner"
+          - "{{ .GITHUB_ORG }}"
+          - "--repo"
+          - "{{ .GITHUB_REPO }}"
+          - "--github-pr"
+          - "{{ .PR_NUMBER }}"
+          - "--workspace"
```
krasi-georgiev: can you remind me what is this arg and why is it needed?

krasi-georgiev (commented):
Nice job, very easy to follow and understand!

just a few small questions and can merge it.

prombench/manifests/funcbench/benchmark/2_job.yaml
nevill: `workspace` is a replacement of env variable `GITHUB_WORKSPACE`, it's the directory where the source will be cloned to.

funcbench/README.md
geekodour: I think this stayed after testing it with the prometheus-community repo for prombench, It can also be seen with the [prombench workflow file.](https://github.com/prometheus/prometheus/blob/386aea77744eac455c7430bd33cd0c736dba8a87/.github/workflows/prombench.yml#L28) keeping it here does make it easier to test though, as one will just need to copy and paste the workflow and just replace the variables.

geekodour: Should we remove it? as it's set to the same link in the Dockerfile?

pkg/provider/k8s/k8s.go
nevill: It's inherited from #311, at that time, we didn't have the step `make funcbench_resource_delete`. Now we can keep previous step and remove this line.

funcbench/README.md
nevill: see the next comments.

funcbench/README.md
nevill: I use it for testing my fork. In real case, we don't need to set it.

funcbench/README.md
nevill: Since it's an example triggered by `repository_dispatch` event. ;)

funcbench/README.md
krasi-georgiev: my question was really - what does this link specify? Why do we need to specify the test-infra repo? Does it download any files from there?

funcbench/README.md
krasi-georgiev: I still don't understand why do we need to specify it elsewhere and not directly here?

Unless there is any real use case or a benefit from this don't see why it should be passed from another component. It defies the KISS principle :smile: 

@geekodour what do you think?

funcbench/README.md
krasi-georgiev: sorry I am not sure which comments are you referring to here?

prombench/manifests/funcbench/benchmark/2_job.yaml
krasi-georgiev: can you give me an example of why do we need to have this as an argument and why can't we always use the GITHUB_WORKSPACE env?

krasi-georgiev:
IIRC funcbnech can also be run locally right? If that is true it would be a good idea to also add a section on how to run locally. If you prefer to do this in a separate PR please open an issue so that we don't forget about it.

Having the examples of how to run locally without GH actions would also give some good context on the usage of some cli arguments.

nevill:
> Having the examples of how to run locally without GH actions would also give some good context on the usage of some cli arguments.

Yes, it's [there](https://github.com/prometheus/test-infra/tree/master/funcbench#usage-examples) after #346 got merged.

prombench/manifests/funcbench/benchmark/2_job.yaml
nevill: It's a change introduced by another [pull request](https://github.com/prometheus/test-infra/pull/346#issue-401923405). In this PR, we update the example based on the change.

funcbench/README.md
nevill: Basically, it will clone the TEST_INFRA_REPO to WORKSPACE, then run `prombench` against TEST_INFRA_REPO, in this case, we can update the manifests without rebuilding the docker image.
As a reference, this `TEST_INFRA_REPO` was introduced in #266.

funcbench/README.md
nevill: `REGEX` is an input from user, we need comment-monitor to extract it out, then pass it to GH action via `repository_dispatch` event.
Do you suggest switching back to `issue_comment` event ?

funcbench/README.md
nevill: I'm referring to [this](https://github.com/prometheus/test-infra/pull/343#discussion_r408710386). Line 186 `_ = c.jobDelete(resource)` is already removed from code.
So we have to remove the `Job` resource everytime it finishes, no matter it's successful or failed.

prombench/manifests/funcbench/benchmark/2_job.yaml
nevill: Or, are you suggesting `--workspace ${GITHUB_WORKSPACE}` ?

funcbench/README.md
krasi-georgiev: yeah I remember now, don't think we should expose this in the workflow file as there is a very, very small chance someone would ever need to change this. Let's remove it from the example and leave it set to the default.

funcbench/README.md
krasi-georgiev: no I was referring only to the `if: always()` line. Would this execute always even without this line?

prombench/manifests/funcbench/benchmark/2_job.yaml
krasi-georgiev: no I mean do you even need to set this? Why not have some default so clone to `/tmp` and that is it?

I mean the flag can stay in the funcbench tool as an option if someone needs to set it something different than the default, but why can't see skip setting it in the workflow file and leave it to some default option?

fewer flags to think about less mental baggage :smile:

krasi-georgiev: I see that the default is `/tmp/funcbench` so why do you need to set it to `/build` ?

funcbench/README.md
nevill: Yes, we can remove this line. It just emphasizes that it's always executed. In real case, we will have `if: failure()` and `if: success()` behind. Using `if: always()` makes sure that the exit code of step `Run funcbench` can be passed down the chain (not interrupted by the exit code of step `Recycle all`).

geekodour (commented):
@nevill can't we have this in the docker image? Or maybe just put this in the readme? Does not having this break something? Sorry i removed it without notice and forgot to mention! 

context : ~b28e84f~ 732eac7

krasi-georgiev (approved):

nevill:
> @nevill can't we have this in the docker image? Or maybe just put this in the readme? Does not having this break something? Sorry i removed it without notice and forgot to mention!
> 
Because we don't have `gcc` in docker image, but CGO relies on it. It's introduced by [#275](https://github.com/prometheus/test-infra/pull/275/files#diff-d4ecb9c1cc7e2fbe0ec77c6deffa7eb2), and was mentioned in README but removed  [here](https://github.com/prometheus/test-infra/pull/346/files#diff-a164d67aaecfad99caff7e1c6c6b2357L58).

nevill:
I removed `GO111MODULE=on` from `newGitHubEnv`, since our image is based go-1.13, and prometheus source is outside `GOPATH`.

geekodour:
merged. Thanks!

PR #346, opened 2020-04-10
Refactoring and documenting funcbench

geekodour:
This PR does not add any new feature expect ~`--dryrun`~ `--nocomment` flag, It's mostly refactoring, reordering and fixing some bugs/missing features.

Changes:
- Updated readme and removed all old Github Action references.
- Removed the use of `GITHUB_WORKSPACE` completely inside funcbench and added command line arguments into a configuration variable.
- Improved logging and comments
- Renamed `funcbench` to `startBenchmark`
- Renamed `compareTargetRef` to `getTargetInfo`
- Added ~`--dryrun`~ `--nocomment` flag
- Added test for `formatCommentToMD`
- Enables tests to be run from commit hash as the target

Please let me know if I should split this PR into multiple PRs by the commits.
cc: @krasi-georgiev @nevill @bwplotka

### DIFF
--- a/funcbench/README.md
+++ b/funcbench/README.md
@@ -1,100 +1,33 @@
 # funcbench
 
-A tool used as a github action to run a `go test -bench` and compare changes from a PR against another branch or commit.
+Benchmark and compare your Go code between commits or sub benchmarks. It uses `go test -bench` to run the benchmarks and uses [benchcmp](https://godoc.org/golang.org/x/tools/cmd/benchcmp) to compare them.
 
-The benchmark can be triggered by creating a comment which specifies a branch to compare. The results are then posted back as a PR comment.
-The benchmark can be also trigger and used as CLI command, without GitHub hook.
+funcbench currently supports two modes, Local and GitHub. Running it in the Github Mode also allows it to accept *a pull request number* and *a branch/commit* to compare against, which makes it suitable for automated tests.
 
-Comparison is done using [benchcmp](https://godoc.org/golang.org/x/tools/cmd/benchcmp).
-Arguments for the benchcmp are read from files created by previous action (for example [commentMonitor](/tools/commentMonitor/main.go)),
-which is responsible for the comment parsing.
+### Environment variables
+- `GITHUB_TOKEN`: Access token to post benchmarks results to respective PR.
 
-## Usage
+## Usage Examples
+> Clean git state is required.
 
-NOTE: Clean git state is required.
-Examples:
+|Usage|Command|
+|--|--|
+|Execute benchmark named `BenchmarkFuncName` regex, and compare it with `master` branch. | ``` ./funcbench -v master BenchmarkFuncName ``` |
+|Execute all benchmarks matching `BenchmarkFuncName.*` regex, and compare it with `master` branch.|```./funcbench -v master BenchmarkFuncName.*```|
+|Execute all benchmarks, and compare the results with `devel` branch.|```./funcbench -v devel . ```|
+|Execute all benchmarks matching `BenchmarkFuncName.*` regex, and compare it with `6d280faa16bfca1f26fa426d863afbb564c063d1` commit.|```./funcbench -v 6d280faa16bfca1f26fa426d863afbb564c063d1 BenchmarkFuncName.*```|
+|Execute all benchmarks matching `BenchmarkFuncName.*` regex on current code. Compare it between sub-benchmarks (`b.Run`) of same benchmark for current commit. Errors out if there are no sub-benchmarks.|```./funcbench -v . FuncName.*```|
+|Execute benchmark named `BenchmarkFuncName`, and compare `pr#35` with `master` branch.|```./funcbench --dryrun --github-pr="35" master BenchmarkFuncName```|
 
-* Execute benchmark named `FuncName` regex, and compare it with `master` branch.
-
- ```
- /funcbench -v master BenchmarkFuncName
- ```
-
-* Execute all benchmarks matching `FuncName.*` regex, and compare it with `master` branch.
-
+## Triggering with GitHub comments
+The benchmark can be triggered by creating a comment in a PR which specifies a branch to compare. The results are then posted back to the PR as a comment.
 ```
- /funcbench -v master FuncName.*
- ```
-
-* Execute all benchmarks, and compare the results with `devel` branch.
-
- ```
- /funcbench -v devel .
- ```
-
-* Execute all benchmarks matching `FuncName.*` regex, and compare it with `6d280faa16bfca1f26fa426d863afbb564c063d1` commit.
-
- ```
- /funcbench -v 6d280faa16bfca1f26fa426d863afbb564c063d1 FuncName.*
- ```
-
-* Execute all benchmarks matching `FuncName.*` regex on current code. Compare it between sub-benchmarks (`b.Run`) of same benchmark for current commit.
-Errors out if there are no sub-benchmarks.
-
- ```
- /funcbench -v . FuncName.*
- ```
-
-### GitHub
-
-Tests are triggered by posting a comment in a PR with the following format:
-
-`/funcbench <branch/commit> <Go test regex>`
-
-Specifying which tests to run are filtered by using the standard [Go regex RE2 language](https://github.com/google/re2/wiki/Syntax).
-
-* To test it locally, set `-w` flag or `WORKSPACE` environment variable to an empty directory where the source will be cloned.
-
-By default all benchmarks run without `-race` flag (#275).
-
-#### Example Github actions workflow file to pass in --input flag.
-
-> TODO: No longer using `issue_comment`, to be replaced with commentMonitor usage.
-
-```
-on: issue_comment // Workflow is executed when a pull request comment is created.
-name: Benchmark
-jobs:
-  commentMonitor:
-    runs-on: ubuntu-latest
-    steps:
-    - name: commentMonitor
-      uses: docker://prominfra/comment-monitor:latest
-      env:
-        COMMENT_TEMPLATE: 'The benchmark has started.' // Body of a comment that is created to announce start of a benchmark.
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} // Github secret token/
-      with:
-        args: '"^/funcbench ?(?P<BRANCH>[^ B\.]+)? ?(?P<REGEX>\.|Bench.*|[^ ]+)?'
-    - name: benchmark
-      uses: docker://prominfra/funcbench:latest
-      env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} // Github secret token/
+/funcbench <branch/commit> <benchmark function regex>
 ```
 
-#### Set up
-
-This tool is meant to be used with a Github action. The action itself is, to a large degree, unusable alone, as you need
-to combine it with another Github action that will provide necessary files to it. At this time, the only action it is
-supposed to work with, is [comment-monitor](https://github.com/prometheus/prombench/tree/master/tools/commentMonitor).
-
-- Create Github actions workflow file that is executed when an issue comment is created, `on = "issue_comment"`.
-- Add comment-monitor Github action as a first step.
-- Specify this regex `^/funcbench ?(?P<BRANCH>[^ B\.]+)? ?(?P<REGEX>\.|Bench.*|[^ ]+)?` in the `args` field of the comment-monitor.
-- Specify this Github action as a pre-built image, build from this source code, or just refer to this repository from the workflow file.
-- Provide a Github token as an environment variable to both comment-monitor and funcbench.
+The comment is handled by [comment-monitor](https://github.com/prometheus/test-infra/tree/master/tools/commentMonitor) and then the parsed arguments are handed over to funcbench(if using Github Actions) or to [prombench](https://github.com/prometheus/test-infra/tree/master/prombench) if using funcbench with GKE.
 
 ## Building Docker container.
-
 From the repository root:
 
 `docker build -t <tag of your choice> -f funcbench/Dockerfile .`
--- a/funcbench/README.md
+++ b/funcbench/README.md
@@ -1,13 +1,15 @@
 # funcbench
 
-Benchmark and compare your Go code between commits or sub benchmarks. It uses `go test -bench` to run the benchmarks and uses [benchcmp](https://godoc.org/golang.org/x/tools/cmd/benchcmp) to compare them.
+Benchmark and compare your Go code between commits or sub benchmarks. It automated the use of `go test -bench` to run the benchmarks and uses [benchcmp](https://godoc.org/golang.org/x/tools/cmd/benchcmp) to compare them.
 
-funcbench currently supports two modes, Local and GitHub. Running it in the Github Mode also allows it to accept *a pull request number* and *a branch/commit* to compare against, which makes it suitable for automated tests.
+funcbench currently supports two modes, Local and GitHub. Running it in the Github mode also allows it to accept *a pull request number* and *a branch/commit* to compare against, which makes it suitable for automated tests.
 
-### Environment variables
+## Environment variables
+
 - `GITHUB_TOKEN`: Access token to post benchmarks results to respective PR.
 
 ## Usage Examples
+
 > Clean git state is required.
 
 |Usage|Command|
@@ -20,6 +22,7 @@ funcbench currently supports two modes, Local and GitHub. Running it in the Gith
 |Execute benchmark named `BenchmarkFuncName`, and compare `pr#35` with `master` branch.|```./funcbench --dryrun --github-pr="35" master BenchmarkFuncName```|
 
 ## Triggering with GitHub comments
+
 The benchmark can be triggered by creating a comment in a PR which specifies a branch to compare. The results are then posted back to the PR as a comment.
 ```
 /funcbench <branch/commit> <benchmark function regex>
@@ -28,6 +31,7 @@ The benchmark can be triggered by creating a comment in a PR which specifies a b
 The comment is handled by [comment-monitor](https://github.com/prometheus/test-infra/tree/master/tools/commentMonitor) and then the parsed arguments are handed over to funcbench(if using Github Actions) or to [prombench](https://github.com/prometheus/test-infra/tree/master/prombench) if using funcbench with GKE.
 
 ## Building Docker container.
+
 From the repository root:
 
 `docker build -t <tag of your choice> -f funcbench/Dockerfile .`
--- a/funcbench/README.md
+++ b/funcbench/README.md
@@ -1,6 +1,6 @@
 # funcbench
 
-Benchmark and compare your Go code between commits or sub benchmarks. It automated the use of `go test -bench` to run the benchmarks and uses [benchcmp](https://godoc.org/golang.org/x/tools/cmd/benchcmp) to compare them.
+Benchmark and compare your Go code between commits or sub benchmarks. It automates the use of `go test -bench` to run the benchmarks and uses [benchcmp](https://godoc.org/golang.org/x/tools/cmd/benchcmp) to compare them.
 
 funcbench currently supports two modes, Local and GitHub. Running it in the Github mode also allows it to accept *a pull request number* and *a branch/commit* to compare against, which makes it suitable for automated tests.
 
--- a/funcbench/README.md
+++ b/funcbench/README.md
@@ -19,7 +19,7 @@ funcbench currently supports two modes, Local and GitHub. Running it in the Gith
 |Execute all benchmarks, and compare the results with `devel` branch.|```./funcbench -v devel . ```|
 |Execute all benchmarks matching `BenchmarkFuncName.*` regex, and compare it with `6d280faa16bfca1f26fa426d863afbb564c063d1` commit.|```./funcbench -v 6d280faa16bfca1f26fa426d863afbb564c063d1 BenchmarkFuncName.*```|
 |Execute all benchmarks matching `BenchmarkFuncName.*` regex on current code. Compare it between sub-benchmarks (`b.Run`) of same benchmark for current commit. Errors out if there are no sub-benchmarks.|```./funcbench -v . FuncName.*```|
-|Execute benchmark named `BenchmarkFuncName`, and compare `pr#35` with `master` branch.|```./funcbench --dryrun --github-pr="35" master BenchmarkFuncName```|
+|Execute benchmark named `BenchmarkFuncName`, and compare `pr#35` with `master` branch.|```./funcbench --nocomment --github-pr="35" master BenchmarkFuncName```|
 
 ## Triggering with GitHub comments
 
--- a/funcbench/bench.go
+++ b/funcbench/bench.go
@@ -79,19 +79,23 @@ func (b *Benchmarker) execBenchmark(pkgRoot string, commit plumbing.Hash) (out s
 	}
 
 	// TODO(bwplotka): Allow memprofiles.
+	// 'go test' flags: https://golang.org/cmd/go/#hdr-Testing_flags
 	extraArgs := []string{"-benchtime", b.benchTime.String()}
 	extraArgs = append(extraArgs, "-timeout", b.benchTimeout.String())
 	benchCmd := []string{"bash", "-c", strings.Join(append(append([]string{"cd", pkgRoot, "&&", "go", "test", "-run", "^$", "-bench", fmt.Sprintf("^%s$", b.benchFunc), "-benchmem"}, extraArgs...), "./..."), " ")}
 
-	b.logger.Println("Executing benchmark cmd:", benchCmd)
+	b.logger.Println("Executing benchmark command for", commit.String())
+	b.logger.Println(benchCmd)
 	out, err = b.c.exec(benchCmd...)
 	if err != nil {
 		return "", errors.Wrap(err, "benchmark ended with an error.")
 	}
 
-	fn := fileName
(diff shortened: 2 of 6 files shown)


funcbench/README.md
```diff
@@ -1,100 +1,33 @@
 # funcbench
 
-A tool used as a github action to run a `go test -bench` and compare changes from a PR against another branch or commit.
+Benchmark and compare your Go code between commits or sub benchmarks. It uses `go test -bench` to run the benchmarks and uses [benchcmp](https://godoc.org/golang.org/x/tools/cmd/benchcmp) to compare them.
```
bwplotka:
```suggestion
Benchmark and compare your Go code between commits or sub benchmarks. It automates the use of `go test -bench` to run the benchmarks and uses [benchcmp](https://godoc.org/golang.org/x/tools/cmd/benchcmp) to compare them across different commits using [git worktree](link)
```

funcbench/README.md
```diff
@@ -1,100 +1,33 @@
...
 
-The benchmark can be triggered by creating a comment which specifies a branch to compare. The results are then posted back as a PR comment.
-The benchmark can be also trigger and used as CLI command, without GitHub hook.
+funcbench currently supports two modes, Local and GitHub. Running it in the Github Mode also allows it to accept *a pull request number* and *a branch/commit* to compare against, which makes it suitable for automated tests.
```
bwplotka:
```suggestion
funcbench currently supports two modes, Local and GitHub. Running it in the Github mode also allows it to accept *a pull request number* and *a branch/commit* to compare against, which makes it suitable for automated tests.
```

funcbench/README.md
```diff
@@ -1,100 +1,33 @@
...
 
-Comparison is done using [benchcmp](https://godoc.org/golang.org/x/tools/cmd/benchcmp).
-Arguments for the benchcmp are read from files created by previous action (for example [commentMonitor](/tools/commentMonitor/main.go)),
-which is responsible for the comment parsing.
+### Environment variables
```
bwplotka:
```suggestion
## Environment variables
```

funcbench/README.md
```diff
@@ -1,100 +1,33 @@
...
-Specifying which tests to run are filtered by using the standard [Go regex RE2 language](https://github.com/google/re2/wiki/Syntax).
-
-* To test it locally, set `-w` flag or `WORKSPACE` environment variable to an empty directory where the source will be cloned.
-
-By default all benchmarks run without `-race` flag (#275).
-
-#### Example Github actions workflow file to pass in --input flag.
-
-> TODO: No longer using `issue_comment`, to be replaced with commentMonitor usage.
+|Usage|Command|
```
bwplotka, line 15: This is nice. But what I would recommend is:

* Put all of this into funcbench --help in code.
* Use embedmd as we do in Thanos to automatically generate docs from it! (: See: https://github.com/thanos-io/thanos/blob/cc0b7e62b8b2c6f0879e5bfdd9881a1f7d699fc4/scripts/genflagdocs.sh#L3

This way we can have only one place that we need to copy this and update on changes + it is available on `funcbench --help! :hugs:

bwplotka: What do you think?

funcbench/README.md
```diff
@@ -1,100 +1,33 @@
...
+|Execute all benchmarks matching `BenchmarkFuncName.*` regex, and compare it with `master` branch.|```./funcbench -v master BenchmarkFuncName.*```|
+|Execute all benchmarks, and compare the results with `devel` branch.|```./funcbench -v devel . ```|
+|Execute all benchmarks matching `BenchmarkFuncName.*` regex, and compare it with `6d280faa16bfca1f26fa426d863afbb564c063d1` commit.|```./funcbench -v 6d280faa16bfca1f26fa426d863afbb564c063d1 BenchmarkFuncName.*```|
+|Execute all benchmarks matching `BenchmarkFuncName.*` regex on current code. Compare it between sub-benchmarks (`b.Run`) of same benchmark for current commit. Errors out if there are no sub-benchmarks.|```./funcbench -v . FuncName.*```|
+|Execute benchmark named `BenchmarkFuncName`, and compare `pr#35` with `master` branch.|```./funcbench --dryrun --github-pr="35" master BenchmarkFuncName```|
 
+## Triggering with GitHub comments
+The benchmark can be triggered by creating a comment in a PR which specifies a branch to compare. The results are then posted back to the PR as a comment.
```
bwplotka, line 26: new lines are needed between title and text for md

funcbench/bench.go
```diff
@@ -79,19 +79,23 @@ func (b *Benchmarker) execBenchmark(pkgRoot string, commit plumbing.Hash) (out s
...
 
 	// TODO(bwplotka): Allow memprofiles.
+	// 'go test' flags: https://golang.org/cmd/go/#hdr-Testing_flags
 	extraArgs := []string{"-benchtime", b.benchTime.String()}
 	extraArgs = append(extraArgs, "-timeout", b.benchTimeout.String())
 	benchCmd := []string{"bash", "-c", strings.Join(append(append([]string{"cd", pkgRoot, "&&", "go", "test", "-run", "^$", "-bench", fmt.Sprintf("^%s$", b.benchFunc), "-benchmem"}, extraArgs...), "./..."), " ")}
 
-	b.logger.Println("Executing benchmark cmd:", benchCmd)
+	b.logger.Println("Executing benchmark command for", commit.String())
+	b.logger.Println(benchCmd)
```
bwplotka: Why two lines? This is one event, so you can print once. Just add `\n` char between  in the message (:

funcbench/env.go
```diff
@@ -62,25 +58,14 @@ func newLocalEnv(e environment) (Environment, error) {
 	if err != nil {
 		return nil, err
 	}
-
+	e.logger.Println("[Local Mode]")
```
bwplotka: Same here, would nice if you can format message to be as you want and just print once

funcbench/env.go
```diff
@@ -143,58 +109,27 @@ func newGitHubEnv(ctx context.Context, e environment, owner, repo string, prNumb
...
-		latestCommitHash: os.Getenv("GITHUB_SHA"),
-	}
-	return &c
-}
-
-func (c *gitHubClient) postComment(comment string) error {
-	issueComment := &github.IssueComment{Body: github.String(comment)}
-	_, _, err := c.client.Issues.CreateComment(context.Background(), c.owner, c.repo, c.prNumber, issueComment)
-	return err
+	e.logger.Println("[GitHub Mode]", gc.owner, ":", gc.repo)
```
bwplotka: ditto

funcbench/main.go
```diff
@@ -50,43 +51,67 @@ func (l *logger) FatalError(err error) {
 }
 
 func main() {
+	cfg := struct {
```
bwplotka, line 54: What about using something smarter than this: https://github.com/bwplotka/flagarize

((((:

funcbench/main.go
```diff
@@ -99,35 +124,46 @@ The default is 2 hours (2h).`).Default("2h").Duration()
...
-					compareTarget: *compareTarget,
-					home:          os.Getenv("HOME"),
-				})
+			// Setup Environment.
+			e := environment{
+				logger:        logger,
+				benchFunc:     cfg.benchFuncRegex,
+				compareTarget: cfg.compareTarget,
+			}
+			if cfg.ghPr == 0 {
```
bwplotka:
```suggestion
			if cfg.ghPR == 0 {
```

funcbench/main.go
```diff
@@ -99,35 +124,46 @@ The default is 2 hours (2h).`).Default("2h").Duration()
...
+				// Local Mode.
+				env, err = newLocalEnv(e)
 				if err != nil {
-					return errors.Wrap(err, "new env")
+					return errors.Wrap(err, "environment creation error")
```
bwplotka: We are in errors.Wrap... what else it than error? (: 
```suggestion
					return errors.Wrap(err, "environment create")
```

bwplotka: It's our opinion, so feel free to say no.

It's based on this we just are about to share: https://github.com/thanos-io/thanos/blob/389197775d5008fb8f52b1b3f9bd95da8bda9e1f/docs/contributing/coding-style-guide.md

funcbench/main.go
```diff
@@ -99,35 +124,46 @@ The default is 2 hours (2h).`).Default("2h").Duration()
...
 				}
-				logger.Printf("funcbench start [Local Mode]: Benchmarking current version versus %q for benchmark funcs: %q\n", *compareTarget, *benchFunc)
 			} else {
-				if *owner == "" || *repo == "" {
-					return errors.New("funcbench in GitHub Mode requires --owner and --repo flags to be specified")
+				// Github Mode.
+				ghClient, err := newGitHubClient(ctx, cfg.owner, cfg.repo, cfg.ghPr, cfg.dryrun)
+				if err != nil {
+					return errors.Wrapf(err, "could not create github client")
```
bwplotka: ditto

bwplotka: and everywhere else. It's clear we could not create ... it's already  wrapping error (: So it's just a spam

funcbench/main.go
```diff
@@ -99,35 +124,46 @@ The default is 2 hours (2h).`).Default("2h").Duration()
...
 				}
-				logger.Printf("funcbench start [GitHub Mode]: Benchmarking %q (PR-%d) versus %q for benchmark funcs: %q\n", fmt.Sprintf("%s/%s", *owner, *repo), *gitHubPRNumber, *compareTarget, *benchFunc)
 			}
 
 			// ( ◔_◔)ﾉ Start benchmarking!
-			return funcbench(ctx, env, newBenchmarker(logger, env, &commander{verbose: *verbose}, *benchTime, *benchTimeout, *resultCacheDir))
+			cmps, err := startBenchmark(ctx, env, newBenchmarker(logger, env, &commander{verbose: cfg.verbose}, cfg.benchTime, cfg.benchTimeout, cfg.resultsDir))
+			if err != nil {
+				if pErr := env.PostErr(fmt.Sprintf("%v. Benchmark failed, please check logs.", err)); pErr != nil {
+					return errors.Wrap(err, "could not log error")
```
bwplotka: This is confusing. I think this has to be logged with `pErr`... wrapping it does not really make sense. It does not relate to error.

funcbench/main.go
```diff
@@ -149,82 +185,78 @@ The default is 2 hours (2h).`).Default("2h").Duration()
...
 	}
 
-	var oldResult string
-
-	targetCommit, compareWithItself, err := compareTargetRef(ctx, env.Repo(), env.CompareTarget())
+	// Get info about target.
+	targetCommit, compareWithItself, err := getTargetInfo(ctx, env.Repo(), env.CompareTarget())
 	if err != nil {
-		return errors.Wrap(err, "compareTargetRef")
+		return nil, errors.Wrap(err, "failed to get target info")
```
bwplotka: `failed`? :open_mouth: Wow I expect it working if I have see error! ((((:

funcbench/main.go
```diff
@@ -149,82 +185,78 @@ The default is 2 hours (2h).`).Default("2h").Duration()
...
 
-		// 2a. Compare sub benchmarks. TODO.
-		cmps, err := bench.compareSubBenchmarks(newResult)
+		cmps, err := bench.compareSubBenchmarks(subResult)
 		if err != nil {
-			if pErr := env.PostErr("`benchcmp` failed."); pErr != nil {
-				return errors.Errorf("error: %v occured while processing error: %v", pErr, err)
-			}
-			return errors.Wrap(err, "compare sub benchmarks")
+			return nil, errors.Wrap(err, "comparing sub benchmarks failed")
```
bwplotka: so many of those... (:

funcbench/main.go
```diff
@@ -149,82 +185,78 @@ The default is 2 hours (2h).`).Default("2h").Duration()
...
+		return nil, errors.Wrap(err, "failed to get target info")
 	}
+	bench.logger.Println("Target:", targetCommit.String(), "Current Ref:", ref.Hash().String())
 
 	if compareWithItself {
-		bench.logger.Println("Target:", env.CompareTarget(), "is `.`; Assuming sub-benchmarks comparison.")
+		bench.logger.Println("Assuming sub-benchmarks comparison.")
+		subResult, err := bench.execBenchmark(wt.Filesystem.Root(), ref.Hash())
+		if err != nil {
+			return nil, errors.Wrap(err, fmt.Sprintf("failed to execute sub-benchmark"))
```
bwplotka: Unnecessary `fmt.Sprintf` function call  (:

funcbench/main.go
```diff
@@ -149,82 +185,78 @@ The default is 2 hours (2h).`).Default("2h").Duration()
...
-	bench.logger.Println("Target:", env.CompareTarget(), "is evaluated to be ", targetCommit.String(), ". Assuming comparing with this one (clean workdir will be checked.)")
-	// 2b. Compare with target commit/branch.
+	bench.logger.Println("Assuming comparing with target (clean workdir will be checked.)")
 
-	// 3. Best effort cleanup of worktree.
-	cmpWorkTreeDir := filepath.Join(wt.Filesystem.Root(), "_funcbench-cmp")
+	// Execute benchmark A.
+	newResult, err := bench.execBenchmark(wt.Filesystem.Root(), ref.Hash())
+	if err != nil {
+		return nil, errors.Wrap(err, fmt.Sprintf("failed to execute benchmark for A: %v", ref.Name().String()))
```
bwplotka:
```suggestion
		return nil, errors.Wrapf(err, "failed to execute benchmark for A: %v", ref.Name().String())
```

funcbench/main.go
```diff
@@ -149,82 +185,78 @@ The default is 2 hours (2h).`).Default("2h").Duration()
...
-	// 4. Benchmark in new worktree.
-	oldResult, err = bench.execBenchmark(cmpWorkTreeDir, targetCommit)
+	// Execute benchmark B.
+	oldResult, err := bench.execBenchmark(cmpWorkTreeDir, targetCommit)
 	if err != nil {
-		if pErr := env.PostErr(fmt.Sprintf("Go bench test for target %s failed", env.CompareTarget())); pErr != nil {
-			return errors.Errorf("error: %v occured while processing error: %v", pErr, err)
-		}
-		return errors.Wrap(err, "exec bench B")
+		return nil, errors.Wrap(err, fmt.Sprintf("failed to execute benchmark for B: %v", env.CompareTarget()))
```
bwplotka: ditto

funcbench/main.go
```diff
@@ -149,82 +185,78 @@ The default is 2 hours (2h).`).Default("2h").Duration()
...
 
-	// 5. Compare old vs new.
+	// Compare B vs A.
 	cmps, err := bench.compareBenchmarks(oldResult, newResult)
 	if err != nil {
-		if pErr := env.PostErr("`benchcmp` failed."); pErr != nil {
-			return errors.Errorf("error: %v occured while processing error: %v", pErr, err)
-		}
-		return errors.Wrap(err, "compare benchmarks")
+		return nil, errors.Wrap(err, "comparing benchmarks failed")
```
bwplotka: ditto

bwplotka (commented):
Nice! Really looks good just minor style nits.

Good work, thanks.

 I think we can improve error wrapping here a lot, but let me know if my comments make sense (:

geekodour:
> I think we can improve error wrapping here a lot, but let me know if my comments make sense (: 

Yes, it does make sense. I thought it would be nice to provide more context to the error messages and split logging into multiple lines as it'll be run locally to make it more like the cli apps :p sorry about the mess!

Looking into the suggested links..

funcbench/README.md
bwplotka: Hm?

bwplotka (approved):
LGTM just suggestion about embedmd :heart: 

Nice!

funcbench/README.md
geekodour: @krasi-georgiev mentioned doing this with prombench and with the other tools almost a year ago. I'll create an issue for this so we can work on this. :)

geekodour: should we put the usage info into the help command in this PR?

funcbench/main.go
geekodour: I'd like to try this with funcbench as it has all the keys that we're using and will make it look real clean :sparkles:  , but if we decide to use this, It'll be good to do it for other tools in test-infra aswell to have some consistency. But since flagarize is super new, wonder what @krasi-georgiev thinks?

funcbench/main.go
```diff
@@ -149,82 +185,78 @@ The default is 2 hours (2h).`).Default("2h").Duration()
...
-		bench.logger.Println("Target:", env.CompareTarget(), "is `.`; Assuming sub-benchmarks comparison.")
+		bench.logger.Println("Assuming sub-benchmarks comparison.")
+		subResult, err := bench.execBenchmark(wt.Filesystem.Root(), ref.Hash())
+		if err != nil {
+			return nil, errors.Wrap(err, "execute sub-benchmark")
+		}
 
-		// 2a. Compare sub benchmarks. TODO.
-		cmps, err := bench.compareSubBenchmarks(newResult)
+		cmps, err := bench.compareSubBenchmarks(subResult)
```
nevill, line 226: Is this necessary ? What are sub-benchmarks ? I think we run all the sub-benchmarks with `go test -bench=. ./...`.

nevill (commented):
1. The parameter name `--dryrun` is confusing. We actually run the benchmark, but not post results to github in fact. Would better be `--no-comments`.

2. I agree we should take care of wrapped errors. For example, we can get this error:
```
funcbech10:25:09 main.go:48: open _dev/funcbench/QmVuY2htYXJrUmFuZ2VRdWVyeS9leHByPWhpc3RvZ3JhbV9xdWFudGlsZS4q-d9613e5c466c6e9de548c4dae1b9aabf9aaf7c57.out: permission denied
execute benchmark for B: release-2.15.2
main.startBenchmark
    /home/nevill/test-infra/funcbench/main.go:251
main.main.func1
    /home/nevill/test-infra/funcbench/main.go:156
github.com/oklog/run.(*Group).Run.func1
    /home/nevill/go/pkg/mod/github.com/oklog/run@v1.0.0/group.go:38
runtime.goexit
    /home/nevill/golang-1.13.8/src/runtime/asm_amd64.s:1357
running command failed
main.main
    /home/nevill/test-infra/funcbench/main.go:183
runtime.main
    /home/nevill/golang-1.13.8/src/runtime/proc.go:203
runtime.goexit
    /home/nevill/golang-1.13.8/src/runtime/asm_amd64.s:1357
```
In fact, the actual place where error occurred is in bench.go:72.

3. I got a `git worktree remove` error on Debian 9 (git version 2.11), since only git 2.17+ can support this command, we'd better replace with `prune` command, or think of using `gopkg.in/src-d/go-git.v4` ?

funcbench/main.go
geekodour: sub-benchmark comparison is currently not implemented, what I assumed @bwplotka wanted to with sub-benchmarks was that you'd write benchmarks with two sub-benchmarks and they'll be compared. not sure though.

https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks

funcbench/main.go
```diff
@@ -50,43 +51,67 @@ func (l *logger) FatalError(err error) {
...
+		Short('d').Default("2h").DurationVar(&cfg.benchTimeout)
+
+	app.Arg("target", "Can be one of '.', branch name or commit SHA of the branch "+
+		"to compare against. If set to '.', branch/commit is the same as the current one; "+
+		"funcbench will run once and try to compare between 2 sub-benchmarks. "+
+		"Errors out if there are no sub-benchmarks.").
+		Required().StringVar(&cfg.compareTarget)
+	app.Arg("function-regex", "Function regex to use for benchmark."+
+		"Supports RE2 regexp and is fully anchored, by default will run all benchmarks.").
+		Default(".").
```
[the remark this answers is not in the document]
geekodour: I did put `".*"` initially, but decided to change it to "." as it's stated:
>  To run all benchmarks, use `'-bench .'` or `'-bench=.'`

in https://golang.org/cmd/go/#hdr-Testing_flags

Should we change it back?

geekodour:
> The parameter name --dryrun is confusing. We actually run the benchmark, but not post results to github in fact. Would better be --no-comments.

I think this flag will be useful if we add tests later because it can be used and is being used for more things than just to disable comments. If we want to see some output when running any funcbench command we can always use `-v`? we can still log the result but that's not the purpose of the Github mode I suppose. Please let me know if you still think to change this to `--no-comments` will be a good idea.

cc: @nevill

funcbench/main.go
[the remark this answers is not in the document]
geekodour: added the change. I forgot that the command was anchoring the provided regex. :p thanks!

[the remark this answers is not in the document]
nevill: The thing is we do `"-bench", fmt.Sprintf("^%s$", b.benchFunc)` in `execBenchmark`, which presumes that the input is always a regex, so `.` only represents ONE character here.

nevill:
> > The parameter name --dryrun is confusing. We actually run the benchmark, but not post results to github in fact. Would better be --no-comments.
> 
> I think this flag will be useful if we add tests later because it can be used and is being used for more things than just to disable comments. If we want to see some output when running any funcbench command we can always use `-v`? we can still log the result but that's not the purpose of the Github mode I suppose. Please let me know if you still think to change this to `--no-comments` will be a good idea.
> 

I am fine with the usage of this flag, but please give it an appropriate name, if `no-comments` is not. IMO, `dry-run` as a convention, is always NOT doing anything real.

nevill:
I see `githubClient` only has one method `postComment`, it's probably `dry-run` to `githubClient` itself. But this flag is a part of `funcbench`, which means to make `funcbench` dry run.

Here is a help message from `git rm --dry-run` for example.
```
-n, --dry-run
Don't actually remove any file(s). Instead, just show if they exist in the index and would otherwise be removed by the command.
```

funcbench/README.md
bwplotka: Nah, we can do later

funcbench/main.go
```diff
@@ -149,82 +185,85 @@ The default is 2 hours (2h).`).Default("2h").Duration()
...
-	if _, err := bench.c.exec("git", "worktree", "add", "-f", cmpWorkTreeDir, ref.Hash().String()); err != nil {
-		return errors.Wrapf(err, "failed to checkout %s in worktree %s", ref.String(), cmpWorkTreeDir)
+	// Best effort cleanup and checkout new worktree.
+	if err := os.RemoveAll(cmpWorkTreeDir); err != nil {
+		return nil, errors.Wrapf(err, "delete worktree at %s", cmpWorkTreeDir)
 	}
 
-	// 4. Benchmark in new worktree.
-	oldResult, err = bench.execBenchmark(cmpWorkTreeDir, targetCommit)
+	if _, err := bench.c.exec("git", "worktree", "prune"); err != nil {
```
bwplotka, line 247: I don't think this is nice. What you have locally 10000 important worktrees? We should remove just one

geekodour: I added this because @nevill mentioned that it's not working with git < 2.17, otherwise we can mention that funcbench requires git 2.17+  ?

But i tested this, `prune` seems to remove worktrees that no longer exists so it will not remove other worktrees. I'll test it again tomorrow just to confirm.

nevill: For `prune`, we have to `rm` the directory first, then prune.

In fact, I prefer `go-git` to do it, in that case we can remove `git` CLI as dependency in container image. Maybe another pull request to fix it.

nevill: As long as we use git >= 2.17, it's fine. I checked current alpine based image, it's v2.24.1.

funcbench/main.go
bwplotka: `t.Run`

bwplotka: It's not but would be nice

funcbench/main.go
bwplotka: go git has zero support of worktrees

bwplotka: It's not the best library :/  

Anyway, everything is better than malforming git repo for someone! (:

bwplotka: If prune works on not used ones, then fine I guess (:

funcbench/main.go
nevill: AFAIK, If we specify the sub-benchmarks via regex, it will be executed within `execBenchmark`. What will `compareSubBenchmarks` do ?

bwplotka: Choosing is not a problem. Comparing between subbenchmarks is. And I mean literally:

* Run now once on the current branch.
* /sub1 and /sub2 was ran.
* Compare sub1 vs sub2.

nevill: Oh, you mean comparing with different sub-benchmarks in the same run ?
e.g.
In Run A, we executed sub1, sub2; In Run B, we executed sub1, sub2 as well.
`compareSubBenchmarks` will compare A.sub1 to A.sub2.
`compareBenchmarks` will compare A.sub1 with B.sub1, and A.sub2 with B.sub2 ?

geekodour:
I added the suggested changes,

@krasi-georgiev can you take a final look at this PR?

funcbench/README.md
bwplotka: Can we add TODO/Github issue?

bwplotka (approved):
LGTM!

Thanks, good work :+1:

bwplotka:
Hope @krasi-georgiev would not mind if I merge this. (:

krasi-georgiev:
Happy to see such nice progress!

PR #348, opened 2020-04-14
use release in prombench restart command example

roidelapluie:
Signed-off-by: Julien Pivotto <roidelapluie@inuits.eu>


geekodour:
Thanks! I think we should also update it here:

https://github.com/prometheus/test-infra/blob/cc8457c5e1935fe0b0285c1d4fde1a2f1b4e5fcb/prombench/manifests/cluster-infra/7a_commentmonitor_configmap_noparse.yaml#L50

PR #349, opened 2020-04-17
Fix missing variables in `funcbench_resource_delete`

nevill:
`2_job.yaml` requires more variables.

### DIFF
--- a/prombench/Makefile
+++ b/prombench/Makefile
@@ -65,9 +65,12 @@ funcbench_resource_apply:
 		-v BRANCH:${BRANCH} -v BENCH_FUNC_REGEX:${BENCH_FUNC_REGEX} \
 		-f manifests/funcbench/benchmark
 
+# Removal of namespace should be at the end, after all other resources get removed.
 funcbench_resource_delete:
 	$(PROMBENCH_CMD) gke resource delete -a ${AUTH_FILE} \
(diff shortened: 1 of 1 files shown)


prombench/Makefile
```diff
@@ -68,6 +68,8 @@ funcbench_resource_apply:
 funcbench_resource_delete:
 	$(PROMBENCH_CMD) gke resource delete -a ${AUTH_FILE} \
 		-v ZONE:${ZONE} -v PROJECT_ID:${PROJECT_ID} -v CLUSTER_NAME:${CLUSTER_NAME} \
-		-v PR_NUMBER:${PR_NUMBER} \
+		-v PR_NUMBER:${PR_NUMBER} -v GITHUB_TOKEN:${GITHUB_TOKEN} \
+		-v GITHUB_ORG:${GITHUB_ORG} -v GITHUB_REPO:${GITHUB_REPO} \
+		-v BRANCH:${BRANCH} -v BENCH_FUNC_REGEX:${BENCH_FUNC_REGEX} \
 		-f manifests/funcbench/benchmark/2_job.yaml \
```
geekodour, line 75: Is there any order here? i.e `2_job.yaml` needs to be run first and then the namespace for `gke resource delete`? does deleting the nodepool not delete the job?

geekodour: If that's the case, it will be good to but a small comment next to it about the ordering.^

nevill: Comment added.

Issue #347, opened 2020-04-14
Generate automated help text for test-infra tools


geekodour:
Currently the usage example for most of the test-infra tools looks like the following in README files:
 ```
./<executable-name> --help
```
We can make `--help` flag more helpful by providing usage examples in it like [amGithubNotifier](https://github.com/prometheus/test-infra/blob/2de6c0e7370788fac314f7868c5f5a532cd1233c/tools/amGithubNotifier/main.go#L72) already does.

As discussed with @krasi-georgiev previously and with @bwplotka  in #346 it'll be a good idea to generate the help texts automatically and embed them into the readme files. This will mean we'll need to update help text in only one place.

Helpful links:
- https://github.com/thanos-io/thanos/blob/master/scripts/genflagdocs.sh

PR #350, opened 2020-04-20
Add funcbench support to comment monitor

nevill:


geekodour:
@nevill I think it'll be good to add one more thing to this PR as its related to comments, currently the github results show `Old` and `New`, it'll be good to have some indication to indicate which column belongs to PR when run with `--github-pr` flag. 

https://github.com/prometheus/test-infra/blob/3546498ff5f72b5fa6ec7d0d477aa5f7b0f2d41b/funcbench/bench.go#L155

Or else we can add it as another PR.

geekodour:
Think just updating the README about which column belongs to the PR will make it apparent aswell. i.e in our case the `New` column.

prombench/manifests/cluster-infra/7a_commentmonitor_configmap_noparse.yaml
```diff
@@ -48,3 +48,10 @@ data:
         **Other Commands:**
         To stop benchmark: `/prombench cancel`
         To restart benchmark: `/prombench restart v2.12.0`
+
+    - event_type: funcbench_start
+      regex_string: ^/funcbench (?P<BRANCH>\S+)? (?P<BENCH_FUNC_REGEX>.*)?
```
geekodour: https://golang.org/pkg/testing/#hdr-Benchmarks, `func BenchmarkXxx(*testing.B)` I think it'll be good to add the regex to check it aswell.

Maybe try this regex? 
```
(?mi)^/funcbench\s+(?P<BRANCH>[^\s]+)\s*(?P<BENCH_FUNC_REGEX>Benchmark[^\s]+)?\s*$
```

nevill: Well, this is tricky. But I guess for 80%, we do all the benchmarks.
When you specify it, you absolutely know what is going on there. Otherwise, you will see a failed result.
So, my idea is, let's keep it simple for now.

nevill:
> @nevill I think it'll be good to add one more thing to this PR as its related to comments, currently the github results show `Old` and `New`, it'll be good to have some indication to indicate which column belongs to PR when run with `--github-pr` flag.
> 
> https://github.com/prometheus/test-infra/blob/3546498ff5f72b5fa6ec7d0d477aa5f7b0f2d41b/funcbench/bench.go#L155
> 
> Or else we can add it as another PR.

I'd like to have another PR for it. Probably includes other changes to the benchmark comments.

prombench/manifests/cluster-infra/7a_commentmonitor_configmap_noparse.yaml
krasi-georgiev: > When you specify it, you absolutely know what is going on there. Otherwise, you will see a failed result.

What do you mean by this? Why will one see failed results?

geekodour: This regex will allow:
```
/funcbench my-branch BenchmarkFuncNameHere
```
and
```
/funcbench my-branch
```
and 
```
/funcbench my-branch

some other text in the comment.
```

I was thinking we could use these two ways, the latter one runs all benchmarks? If comment-monitor is not handling empty `BENCH_FUNC_REGEX` as expected this might fail. Will check this and get back; `comment-monitor` is also missing tests. :(

What comment format do you suggest to run all benchmarks?

krasi-georgiev: I think these ones you suggested are enough as a start

I can't remember exactly, but I think there was a way to run all bench tests for a given package. This would be also useful to add at some point so people can run let's say all bench tests for a single package - promql, tsdb  etc.

nevill: Sorry, I was responding in a hurry. 
> When you specify it, you absolutely know what is going on there. Otherwise, you will see a failed result.

I mean the `BENCH_FUNC_REGEX` part, even prefix with `Benchmark` in regex, it doesn't stop you making mistakes like `/funcbench master BenchmarkWithTypo.*`. You still can get a failed (or empty ?) benchmark result. PS, we can still `/funcbench branch .*`, right ?

Thanks @geekodour gave some good examples on valid commands to start the funcbench. 😺 
so, my suggestion on the regexp would be
`(?m)^/funcbench[[:blank:]]+(?P<BRANCH>\S+)[[:blank:]]*(?P<BENCH_FUNC_REGEX>\S*)?[[:blank:]]*$`
The testcase to test this regexp is (NOTICE: keep the spaces and newlines)
```
/funcbench master-branch .*

/funcbench evenwith/emoji😅 BenchmarkQuery.*

/funcbench master  BenchmarkQuery.*   
/funcbench your_branch benchmark.*

/funcbench the.dot BenchmarkQuery[Some=\S+  
comments

/funcbench master Benchmark?uery\S+

comments

/funcbench master   
comment comment

// followings are invalid

/funcbench     
comments

/funcbenchmaster BenchmarkQuery.*
comment
```

> This would be also useful to add at some point so people can run let's say all bench tests for a single package - promql, tsdb etc.

I guess this requires to specify the package name instead of `./...`(what we are doing now)

funcbench/README.md
```diff
@@ -24,10 +24,21 @@ funcbench currently supports two modes, Local and GitHub. Running it in the Gith
...
-/funcbench <branch/commit> <benchmark function regex>
+# test all the BenchmarkQuery.* on branch master vs this pull request you commented on
+/funcbench master BenchmarkQuery.*
+
+# test all the bencmarks on feature branch vs this pull request
+/funcbench feature-branch
+
+or
+
+/funcbench feature-branch . *
```
geekodour: I think it'll be nice if we format the comment instructions like this:
https://github.com/prometheus/test-infra/tree/master/prombench#trigger-tests-via-a-github-comment

It's more clear. Also I think `/funcbench feature-branch . * ` should be `
/funcbench feature-branch .*` without the space between `.` and `*`

prombench/manifests/cluster-infra/7a_commentmonitor_configmap_noparse.yaml
geekodour: So what is the final regex that we're using? I didn't see any change made to the regex? o.O 


> I mean the `BENCH_FUNC_REGEX` part, even prefix with Benchmark in regex, it doesn't stop you making mistakes like `/funcbench master BenchmarkWithTypo.*`. You still can get a failed (or empty ?) benchmark result. PS, we can still `/funcbench branch .*`, right ?

With the regex I suggested you won't be able to do the second one, but the funcbench cli will still be able to do that ofcourse. The idea of having `Benchmark` in the regex itself was to have it fail more to so that it does not run in vain, because once the regex is validated it'll be spinning up GKE nodes. There's always a probability of having a typo in the comment, cannot avoid it but we can always minimize.

`(?mi)^/funcbench\s+(?P<BRANCH>[^\s]+)\s*(?P<BENCH_FUNC_REGEX>Benchmark[^\s]+|\.\*)?\s*$`

^ with this modification, we explicitly allow `.*`

I am not good with regexes so feel free to improve on it, but it works :)

geekodour (commented):
Minor formatting changes and about regex.

funcbench/README.md
nevill: Oh, the space is a mistake :P

nevill:
> So what is the final regex that we're using? I didn't see any change made to the regex? o.O

I didn't update the regex in code, because I knew it will come more questions. :)
We can discuss on what kind of comment is valid, then wen can decide the final regex.
What about the following comments, are they all valid ?
```
/FuncBench branch benchmark.*
/FuncBench branch benchmark.+
/FuncBench branch bench.*
/FuncBench branch b.+
/FuncBench branch .+

/funcbench branch Bench.*
/funcbench branch Bench.+
/funcbench branch B.*
/funcbench branch B.+
/funcbench branch .+
```

geekodour:
The common naming convention for benchmarks afaik is prefixing it with `Benchmark`

More info:
- https://golang.org/pkg/testing/#hdr-Benchmarks
- https://github.com/golang/go/issues/33688#issuecomment-582554779

And funcbench CLI will always take in any regex that is given to it, if `prometheus/prometheus` is not using any other way to name/run benchmarks other than prefixing them with `Benchmark` then we can safely assume that all our benchmark functions will have `Benchmark` prefix, and avoid starting GKE node when someone runs something like `/funcbench branch-name FuncName.*`, it'll just quietly pass.

@krasi-georgiev  might have more information on it.

But if we really want to make sure that the GKE nodes are not spun up if there's an invalid regex/function name, then we can add another Github Action step to check that there are indeed benchmarks matching that regex and after that only start the prombench setup for funcbench.

Currently these tests will only be run by maintainers, so we can assume that these kind of mistake might not usually happen and not worry about that case.

My suggestions for valid comments are the same I mentioned above:
1. Run specific benchmark
```
/funcbench my-branch BenchmarkFuncNameHere
```
2. Run specific set of benchmarks
```
/funcbench my-branch BenchmarkFuncNam.*
```
3. Run all benchmarks
```
/funcbench my-branch
```
4. Run all benchmarks with the addition of `.*` for completeness
```
/funcbench my-branch .*
```

nevill:
I know we don't want to waste resources, but firstly, let's make things working as expected, then try to minimize the resource consumption.

I updated the regex in the code.
ONLY the followings should be valid:
```
/funcbench my-branch BenchmarkFuncNameHere
/funcbench my/branch  BenchmarkFuncName.*
/funcbench my_branch Benchmark.*
/funcbench mybranch .*
/funcbench release-1.0
```
Notice that, we shouldn't ignore the letter case.
We don't consider the usage of UNICODE characters (which are both valid in branch name and go func name). I guess we rarely use it in the project.

krasi-georgiev:
yep this format looks good and yes lets put it in production and will see if we need to worry about bench name checks

nevill:
Tested with empty BENCH_FUNC_REGEX, basically it will run all the benchmarks normally.

Tested with non-existence benchmark, it will return a comment like
```
comparing benchmarks: no repeated benchmarks. Benchmark failed, please check logs.. Benchmark did not complete, please check action logs.
```

funcbench/README.md
```diff
@@ -24,9 +24,19 @@ funcbench currently supports two modes, Local and GitHub. Running it in the Gith
 ## Triggering with GitHub comments
 
 The benchmark can be triggered by creating a comment in a PR which specifies a branch to compare. The results are then posted back to the PR as a comment.
-```
-/funcbench <branch/commit> <benchmark function regex>
-```
+
+The syntax is: `/funcbench <branch> <benchmark function regex>`
+Examples:
```
geekodour, line 30:
```suggestion

Examples:
```

funcbench/README.md
```diff
@@ -24,9 +24,19 @@ funcbench currently supports two modes, Local and GitHub. Running it in the Gith
...
+
+* `/funcbench master BenchmarkQuery.*` - test all the BenchmarkQuery.* on branch master vs this pull request you commented on
```
geekodour:
```suggestion
* `/funcbench master BenchmarkQuery.*` - compare benchmarks matching `BenchmarkQuery.*` for branch `master` vs  the PR.
```

funcbench/README.md
```diff
@@ -24,9 +24,19 @@ funcbench currently supports two modes, Local and GitHub. Running it in the Gith
...
+
+* `/funcbench feature-branch` or `/funcbench feature-branch .*` - test all the bencmarks on feature branch vs this pull request
```
geekodour:
```suggestion
* `/funcbench feature-branch` or `/funcbench feature-branch .*` - compare all benchmarks for `feature-branch` vs the PR.
```

geekodour (commented):
Minor formatting and README changes suggestions. Can merge after addressing this.

nevill:
@geekodour Updated.

geekodour:
Thanks! :)

Working on a PR #347 will format the README further.

PR #358, opened 2020-05-05
Adjust the format of the funcbench benchmark result

nevill:
This is a continuation of the work on `formatCommentToMD`.
The changes include:
1. A legend to result table.
2. A summary of benchmark result, [here](https://github.com/prometheus/test-infra/pull/355#issuecomment-623397733) suggested a collapsible feature we may use.

We can have more discussions here.
cc @geekodour


nevill:
This is updated, please review it @geekodour .

Issue #356, opened 2020-05-04
Keep funcbench and prombench complitely separate.


krasi-georgiev:
I think we need to keep funcbench and prombench completely separate with separate Makefiles. I know this will lead to few lines if duplicate code, but in the long run it would be easier to understand and maintain.

Since [gke is a package]("github.com/prometheus/test-infra/pkg/provider/gke") It will be simple enough to use it inside funcbench and extend it to support `funcbench gke resource ....`. 

Then funcbench will have its own docker image to use for the github actions - prominfra/funcbench. 

Then all [funcbench makefile targets](https://github.com/prometheus/test-infra/blob/master/prombench/Makefile#L49-L76) would be inside that docker image.

nevill:
Yes, I know we discussed about it, but I may come up with new ideas. :)

IMO `funcbench` should not contain `mainifests` or `prombench` binary. Because
1. when `funcbench` is running inside k8s, it doesn't require `prombench` binary
2. `funcbench` will somehow be used by other projects, they may have their own infrastructure, or their own manifests.
3. `funcbench` doesn't require `prombench` when building.

> As we discussed, we will rename `prombench` to `infra`, I use `infra` instead hereafter.

In fact, all the manifests are parts of our test-infra, they should stay in the same container along with `infra`. Image `infra` are only used in GitHub actions. Container images `tools/*`, `funcbench` are used in kubernetes clusters. 

We can keep the `Makefile` separate by adjusting the directory layout.
for example, in the `infra` image:
```
funcbench/
  - Makefile
  - manifests
prombench/
  - Makefile
  - manifests
cluster-infra/
  - manifests
```

geekodour:
If IUUC, this is the organization that you're suggesting:
```
├── ...
├── manifests/
│   ├── cluster-infra/
│   ├── README.md <- Contains setting up the Main Node
│   ├── cluster.yaml
│   ├── funcbench/ <-- Manifests and Makefile to deploy them with individual readme
│   └── prombench/ <-- Manifests and Makefile to deploy them with individual readme
├── ...
└── tools
    ├── ...
    ├── ...
    ├── funcbench <- /funcbench <- /tools/funcbench
    ├── infra <- this is the prombench cli
    └── ...
```

The `/tools/infra` docker image will have the `/manifests` inside it and have some kind of env var set which directory (`/`,`/manifests/funcbench`,`/manifests/prombench/` etc) should the `make` command be run from.

Is this right?

[...]

PR #362, opened 2020-05-08
[WIP] Use infra as the base docker image for funcbench and prombench

### LINKED ISSUE #357: Prombench is built from master on every push so no need to clone the repo dinamically.
Now that the CI builds and pushed all images on every push to master don't need the clone script https://github.com/prometheus/test-infra/blob/master/prombench/cmd/prombench/clone.sh

Can also remove the need for https://github.com/prometheus/test-infra/blob/master/Makefile , because the docker context can be inside the folder for each tool.

geekodour:
Closes #357 and #356 

~`prominfra/funcbench` now uses `prominfra/prombench` as the base image, this allows the `prominfra/funcbench` to access prombench executable from inside of the docker image. This allows the funcbench workflow file to just use the `prominfra/funcbench` and not depend on `prominfra/prombench`~

~We can probably work on extracting out the `prombench` cli into `prominfra`(a new tool/rename) in another PR as it will involve some reorganizing on documentation aswell. Wanted to keep this PR as short as possible.~

~The changes needed to be applied to current GitHub actions workflow files for this to work:
https://gist.github.com/geekodour/c0da839e3679e8a18cbe1a63e002f616~

Todo:
- [x] Moving `prometheus-builder` to `/tools`
- [ ] Make funcbench run on its own cluster
- [x] Not use root context for docker images
- [x] Test it with new workflow files
- [x] Add the circle-ci docker image build dependency for infra

This is an example run for both funcbench and prombench using images built from this PR:
https://github.com/prometheus-community/prometheus/pull/62

cc: @krasi-georgiev @nevill


nevill:
Sorry, I posted my new comment in #356.

funcbench/Dockerfile
```diff
@@ -7,10 +7,18 @@ COPY go.* ./
...
+WORKDIR /funcbench
+
+ENV PROMBENCH_CMD prombench
+
 COPY --from=builder /go/bin/funcbench /bin/
+COPY funcbench/ ./
+
+# Need 'eval' to prevent bash keywords be run as commands
+RUN echo -e '#!/bin/sh\ncd /funcbench\neval "$@"' >/usr/bin/docker_entrypoint
+RUN chmod u+x /usr/bin/docker_entrypoint
```
nevill: Is `docker_entrypoint` necessary?

geekodour: I could not find a better way to do thing like this: https://github.com/prometheus/prometheus/blob/9e265aba10056547fd920fe603b31eecfced5966/.github/workflows/prombench.yml#L35

There possibly is a better way but for now that's the way we are passing these to prombench docker image(using `eval`), so yes we need some kind of script. we could rename it something else ofcourse.

nevill: Is this method not good enough?
https://github.com/prometheus/test-infra/blob/9134f1ff2f5b9ebfec88e8e43c737fb71216bde7/pkg/provider/gke/gke.go#L386-L389

nevill: OK, I noticed #119 , you mentioned a `Cluster in 'FailedPrecondition' state` error. But seems `until make ...` is not a good way to fix, what if the previous operation removed the nodepools? Then fall into a dead loop?

geekodour: > what if the previous operation removed the nodepools? Then fall into a dead loop?

`until make all_nodepools_running; do echo "waiting for nodepools to be created"; sleep 10; done;`
This basically says, only start deleting nodepools once all nodepools are running, otherwise the github actions that's creating nodepools might create another nodepool as they are running independently.

what do you mean by previous operation? two `/prombench cancel` sent at the same time? In that case one of the nodepool will start deleting first, once it starts doing that `all_nodepools_running` will be returning false so the second one will just keep looping unless some other command independently creates the nodepools. This is the same as running `prombench delete` on a nodepool that does not exist. I think the current waiting mechanism is okay.

We could have some kind of timeout for this but since we're running on Github actions it takes care of it (the 6hrs timeout) and it's very rare that that'll happen so we can probably not worry about it now.

The prow setup followed something like setting a label to the github PR and if that label is set, don't  start the test but the issue itself didn't mean that the benchmark truly is running or not, so we changed it to this.

If we want a more rigid check we can for some metric (maybe the namespace metric) but this approach is working fine for now.

> Cluster in `'FailedPrecondition' state`

That's a gke error, we can probably improve on the logging in the `pkg/gke` but it's not creating any problem, for us it just means we're in progress for the deletion process. The comment I made had to do something with the order how the resource and nodepools are deleted `make clean` the current order is fine and working so far with GKE.

nevill: Thanks for the explanation @geekodour . I misunderstood the target `make all_nodepools_deleted`, which is not removing but checking. :P
Adding a label to the issue/PR sounds a good idea, we can track all the failed benchmarks before users notify us.

Back to the PR, I think we don't use `docker_entrypoint` here in funcbench, can we remove it here?

nevill (changes requested):

prombench/cmd/prombench/Dockerfile
```diff
@@ -10,17 +8,20 @@ COPY prombench/cmd/prombench/*.go ./
...
+COPY --from=builder /go/bin/prombench /usr/bin/prombench
+COPY prombench/ ./
 
-ENV PROMBENCH_DIR $TEST_INFRA_DIR/prombench
+ENV PROMBENCH_CMD prombench
 
-COPY --from=builder /go/bin/prombench /usr/bin/prombench
-COPY prombench/cmd/prombench/clone.sh /usr/bin/clone.sh
+# Need 'eval' to prevent bash keywords be run as commands
+RUN echo -e '#!/bin/sh\ncd /prombench\neval "$@"' >/usr/bin/docker_entrypoint
```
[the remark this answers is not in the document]
geekodour: Github actions unfortunately [does not respect `WORKDIR`](https://help.github.com/en/actions/building-actions/dockerfile-support-for-github-actions#workdir)

funcbench/Dockerfile
geekodour: Yes, we can remove them. I added them here for consistency between the docker images. I'll try simplifying this. thanks for the review :)

prombench/cmd/prombench/Dockerfile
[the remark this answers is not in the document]
geekodour: We could possibly [use this](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#defaultsrun), it was not working the last time I tried few months back when it was not GA. I'll try it again.

.circleci/config.yml
```diff
@@ -40,9 +40,16 @@ jobs:
...
+        container_image_name: infra
+        dockerfile_path: "infra/Dockerfile"
+        registry: docker.io
+        organization: "$DOCKER_ORG"
+        login_variable: DOCKER_LOGIN
+        password_variable: DOCKER_PASSWORD
     - prometheus/publish_images:
         container_image_name: prombench
-        dockerfile_path: "prombench/cmd/prombench/Dockerfile"
+        dockerfile_path: "prombench/Dockerfile"
```
krasi-georgiev, line 52: we need a depend on the infra docker build to make sure that the base image is always the latest.

infra/README.md
```diff
@@ -1,23 +1,23 @@
...
-Files passed to Prombench will be parsed using golang templates, to skip parsing and load the file as is, use `noparse` suffix.
+Files passed to `infra` will be parsed using golang templates, to skip parsing and load the file as is, use `noparse` suffix.
 
 Eg. `somefile.yaml` will be parsed, whereas `somefile_noparse.yaml` will not be parsed.
 
 ### Example for building the docker image
 From the repository root:
 ```
-$ make docker DOCKERFILE_PATH=prombench/cmd/prombench/Dockerfile DOCKER_IMAGE_NAME=prombench DOCKER_IMAGE_TAG=2.0.2
+$ make docker DOCKERFILE_PATH=infra/Dockerfile DOCKER_IMAGE_NAME=infra DOCKER_IMAGE_TAG=2.0.2
```
krasi-georgiev: I would say lets move pkg inside the infra folder and simplify this docker build.

Than can completely remove this paragraph as the docker build would become just `docker build .`

geekodour: other tools like scalar are using `pkg` so I think it's fine it to stay in the repository root. I've opened a PR to add dockerbuild context to the circleci orb (https://github.com/prometheus/circleci/pull/5) so that we can do build images without having the root repository context. Once that's done, we'll be able to replace all docker instructions(in READMEs) in the `test-infra` repo with just `docker build .` and the circleci config can use the `dockerbuild_context`.

krasi-georgiev:
feel free to merge when ready

prombench/cmd/prombench/Dockerfile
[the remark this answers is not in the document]
geekodour: I tested this, `workdir` does not work with docker images, it is supported when using `run` in github actions. Therefore we have to `cd` into the manifest directory anyway, so keeping the bash entrypoint will be a good idea.

geekodour:
@krasi-georgiev @nevill  I tested this PR end to end everything seems to be working as expected. Just the "Make funcbench run on its own cluster" is remaining which I think can be done in a separate PR.

Changes needed in the github action workflow files: https://gist.github.com/geekodour/525d76cad5fcc8ed06312ea9c3e4c818

Can this have another review?

nevill:
IIUC, we still miss the build process in CircleCI, we can do it via `promu`.

.circleci/config.yml
```diff
@@ -29,7 +29,26 @@ jobs:
           # don't limit this to the number of allocated cores, the job is
           # likely to get OOMed and killed.
           GOOPTS: "-p 2"
-  publish:
+  publish_infra:
+    executor: golang
+    steps:
+    - checkout
+    - setup_remote_docker:
+        version: 18.06.0-ce
```
nevill: set version to "19.03.8", because of the [update on base image](https://github.com/circleci/circleci-images/issues/487)

.circleci/config.yml
```diff
@@ -29,7 +29,26 @@ jobs:
...
+    - run:
+        name: Setup environment variables
+        command: |
+          echo "export DOCKER_IMAGE_TAG=master" >> $BASH_ENV
+          echo "export DOCKER_ORG=${DOCKER_ORG:-prominfra}" >> $BASH_ENV
+    - prometheus/publish_images:
+        container_image_name: infra
+        dockerfile_path: "infra/Dockerfile"
+        dockerbuild_context: "infra/"
```
krasi-georgiev, line 47: don't understand why you still need context? if promu builds the binary inside the tool folder and the docker file is there can just run docker build within that folder.

geekodour: the prometheus circle ci orb runs directly from repository root: https://github.com/prometheus/circleci/blob/c4c61ad3812f4e83b0d8661dc5dce0556e1fda05/src/commands/publish_images.yml#L44

if we want to build from the directory directly then we have to `cd` into the directory anyway. eg. `cd infra/` in this case, and do that for every other tool.

geekodour:
@nevill 
> IIUC, we still miss the build process in CircleCI, we can do it via promu.

The `build_and_test` is a requires for the both of the `publish_infra` and `publish_others` jobs. Will this not have the builds in the required directories?

.circleci/config.yml
krasi-georgiev: got it, so you are saying it is more simple this way

geekodour: um currently it's less work, because the orb already has the build and publish commands and the `/Makefile` has docker targets that the circleci orb uses. If we `cd` then we'll probably have to change a couple of things. Maybe have a script/make target to cd and build and push but I think I tried doing a variant of that in a PR previously but then we decided to use the orb.

krasi-georgiev: ok not a problem , lets move forward.

nevill:
> The `build_and_test` is a requires for the both of the `publish_infra` and `publish_others` jobs. Will this not have the builds in the required directories?

I guess we have to utilize orb's `prometheus/build`, `prometheus/setup_build_environment`, to keep the workspace data between job `build_and_test` and `publish_infra`. Right now we just call `make all` directly, without calling `persist_to_workspace`, `attach_workspace`.

PR #374, opened 2020-06-04
Stop using shallow clone

nevill:
This PR is to fix `object not found` error as in https://github.com/prometheus/prometheus/pull/7287#issuecomment-638212583.

It's exactly the same as reported [here](https://github.com/src-d/go-git/issues/1143#issuecomment-529676362).


krasi-georgiev:
lets give it a try.

PR #377, opened 2020-06-05
Try to get target reference from either branch names or tags.

### LINKED ISSUE #373: Add support for git tags and other commits for funcbench in GitHub Mode
Currently funcbench only supports branch name and commit hash as the target. It'll be nice to add support to[ comment like in this comment.](https://github.com/prometheus/prometheus/pull/7287#issuecomment-638246346)

`/funcbench <branch> <benchmark function regex>`

I also think when running funcbench in Github Mode it only allows comparing the PR to the master and not to even any other commit or branch because it clones the repo with depth1. (See logs below)

https://github.com/prometheus/test-infra/blob/3d826afdf0a60ae8440ca8681561ce8240805ade/funcbench/env.go#L87

```
λ ./funcbench --owner=prometheus-community --nocomment --github-pr="63" 8755434350ba9f8d3e5434826a9fcfb3a1327cb0 BenchmarkBufferedSeriesIterator
Enumerating objects: 26444, done.
Counting objects: 100% (26444/26444), done.
Compressing objects: 100% (16010/16010), done.
Total 26444 (delta 12021), reused 22272 (delta 8718), pack-reused 0
funcbech09:34:01 env.go:127: [GitHub Mode] prometheus-community : prometheus
Benchmarking PR - 63 versus: 8755434350ba9f8d3e5434826a9fcfb3a1327cb0
Benchmark func regex: BenchmarkBufferedSeriesIterator
funcbech09:34:01 main.go:223: Target: 8755434350ba9f8d3e5434826a9fcfb3a1327cb0 Current Ref: 4f8680e6fd14c7289e10a249e1d7128400886cfd
funcbech09:34:01 main.go:239: Assuming comparing with target (clean workdir will be checked.)
funcbech09:34:01 bench.go:88: Executing benchmark command for 4f8680e6fd14c7289e10a249e1d7128400886cfd
 [sh -c cd /tmp/funcbench/prometheus && go test -run ^$ -bench ^BenchmarkBufferedSeriesIterator$ -benchmem -benchtime 1s -timeout 2h0m0s ./...]
funcbech09:34:19 main.go:257: Checking out (in new workdir): /tmp/funcbench/prometheus/_funcbench-cmp commmit 8755434350ba9f8d3e5434826a9fcfb3a1327cb0
funcbech09:34:19 main.go:50: running command failed: checkout 8755434350ba9f8d3e5434826a9fcfb3a1327cb0 in worktree /tmp/funcbench/prometheus/_funcbench-cmp: error: exit status 128; Command out: fatal: invalid reference: 8755434350ba9f8d3e5434826a9fcfb3a1327cb0

λ cd ~/OpenSource/prometheus

λ ./funcbench 8755434350ba9f8d3e5434826a9fcfb3a1327cb0 BenchmarkBufferedSeriesIterator                                                      [master]
funcbech09:35:10 env.go:61: [Local Mode]
Benchmarking current version versus: 8755434350ba9f8d3e5434826a9fcfb3a1327cb0
Benchmark func regex: BenchmarkBufferedSeriesIterator
funcbech09:35:10 main.go:223: Target: 8755434350ba9f8d3e5434826a9fcfb3a1327cb0 Current Ref: 83619aa9ac566313942c624df6ee721f5b521032
funcbech09:35:10 main.go:239: Assuming comparing with target (clean workdir will be checked.)
funcbech09:35:10 bench.go:88: Executing benchmark command for 83619aa9ac566313942c624df6ee721f5b521032
 [sh -c cd /home/geekodour/OpenSource/prometheus && go test -run ^$ -bench ^BenchmarkBufferedSeriesIterator$ -benchmem -benchtime 1s -timeout 2h0m0s ./...]
funcbech09:35:42 main.go:257: Checking out (in new workdir): /home/geekodour/OpenSource/prometheus/_funcbench-cmp commmit 8755434350ba9f8d3e5434826a9fcfb3a1327cb0
funcbech09:35:43 bench.go:88: Executing benchmark command for 8755434350ba9f8d3e5434826a9fcfb3a1327cb0
 [sh -c cd /home/geekodour/OpenSource/prometheus/_funcbench-cmp && go test -run ^$ -bench ^BenchmarkBufferedSeriesIterator$ -benchmem -benchtime 1s -timeout 2h0m0s ./...]
Results:
benchmark                             8755434350ba9f8d3e5434826a9fcfb3a1327cb0 ns/op     new ns/op     delta
BenchmarkBufferedSeriesIterator-8     16.0                                               15.6          -2.50%

benchmark                             8755434350ba9f8d3e5434826a9fcfb3a1327cb0 MB/s     new MB/s           speedup
BenchmarkBufferedSeriesIterator-8     66817637610.15                                    78510541505.52     1.17x

benchmark                             8755434350ba9f8d3e5434826a9fcfb3a1327cb0 allocs     new allocs     delta
BenchmarkBufferedSeriesIterator-8     0                                                   0              +0.00%

benchmark                             8755434350ba9f8d3e5434826a9fcfb3a1327cb0 bytes     new bytes     delta
BenchmarkBufferedSeriesIterator-8     0                                                  0             +0.00%
funcbech09:36:10 main.go:191: exiting

```

cc: @krasi-georgiev  @nevill

nevill:
This fix is for #373. Now a tag name like `v2.18.1` can be set as target to funcbench.

### DIFF
--- a/funcbench/README.md
+++ b/funcbench/README.md
@@ -46,11 +46,12 @@ Flags:
                              longer than duration d, panic.
 
 Args:
-  <target>              Can be one of '.', branch name or commit SHA of the
-                        branch to compare against. If set to '.', branch/commit
-                        is the same as the current one; funcbench will run once
-                        and try to compare between 2 sub-benchmarks. Errors out
-                        if there are no sub-benchmarks.
+  <target>              Can be one of '.', tag name, branch name or commit SHA
+                        of the branch to compare against. If set to '.',
+                        branch/commit is the same as the current one; funcbench
+                        will run once and try to compare between 2
+                        sub-benchmarks. Errors out if there are no
+                        sub-benchmarks.
   [<bench-func-regex>]  Function regex to use for benchmark.Supports RE2 regexp
                         and is fully anchored, by default will run all
                         benchmarks.
--- a/funcbench/main.go
+++ b/funcbench/main.go
@@ -22,7 +22,6 @@ import (
 	"os/exec"
 	"os/signal"
 	"path/filepath"
-	"strings"
 	"syscall"
 	"time"
 
@@ -215,14 +214,7 @@ func startBenchmark(
 		return nil, errors.Wrap(err, "not clean worktree")
 	}
 
-	// Get info about target.
-	targetCommit, compareWithItself, err := getTargetInfo(ctx, env.Repo(), env.CompareTarget())
-	if err != nil {
-		return nil, errors.Wrap(err, "getTargetInfo")
-	}
-	bench.logger.Println("Target:", targetCommit.String(), "Current Ref:", ref.Hash().String())
-
-	if compareWithItself {
+	if env.CompareTarget() == "." {
 		bench.logger.Println("Assuming sub-benchmarks comparison.")
 		subResult, err := bench.exec(ctx, wt.Filesystem.Root(), ref.Hash())
 		if err != nil {
@@ -234,6 +226,18 @@ func startBenchmark(
 			return nil, errors.Wrap(err, "comparing sub benchmarks")
 		}
 		return cmps, nil
+	}
+
+	// Get info about target.
+	targetCommit := getTargetInfo(ctx, env.Repo(), env.CompareTarget())
+	if targetCommit == plumbing.ZeroHash {
+		return nil, fmt.Errorf("Cannot getTargetInfo from %s", env.CompareTarget())
+	}
+
+	bench.logger.Println("Target:", targetCommit.String(), "Current Ref:", ref.Hash().String())
+
+	if targetCommit == ref.Hash() {
+		return nil, fmt.Errorf("target: %s is the same as current ref %s (or is on the same commit); No changes would be expected; Aborting", targetCommit, ref.String())
 	}
 
 	bench.logger.Println("Assuming comparing with target (clean workdir will be checked.)")
@@ -286,39 +290,24 @@ func interrupt(logger Logger, cancel <-chan struct{}) error {
 }
 
 // getTargetInfo returns the hash of the target,
-// if target is the same as the current ref, set compareWithItself to true.
-// TODO return ZeroHash if we cannot find reference to target.
-func getTargetInfo(ctx context.Context, repo *git.Repository, target string) (ref plumbing.Hash, compareWithItself bool, _ error) {
-	if target == "." {
-		return plumbing.Hash{}, true, nil
-	}
-
-	currRef, err := repo.Head()
-	if err != nil {
-		return plumbing.ZeroHash, false, err
-	}
-
-	if target == strings.TrimPrefix(currRef.Name().String(), "refs/heads/") || target == currRef.Hash().String() {
-		return currRef.Hash(), true, errors.Errorf("target: %s is the same as current ref %s (or is on the same commit); No changes would be expected; Aborting", target, currRef.String())
-	}
-
+// if it cannot find from either branch names or tag names, plumbing.ZeroHash will return.
+func getTargetInfo(ctx context.Context, repo *git.Repository, target string) plumbing.Hash {
 	commitHash := plumbing.NewHash(target)
 	if !commitHash.IsZero() {
-		return commitHash, false, nil
+		return commitHash
 	}
 
-	branchRef, _ := repo.Reference(plumbing.NewBranchReferenceName(target), false)
-	if branchRef != nil {
-		return branchRef.Hash(), false, nil
+	brachRef, _ := repo.Reference(plumbing.NewBranchReferenceName(target), false)
+	if brachRef != nil {
+		return brachRef.Hash()
 	}
 
 	tagRef, _ := repo.Reference(plumbing.NewTagReferenceName(target), false)
 	if tagRef != nil {
-		return tagRef.Hash(), false, nil
+		return tagRef.Hash()
 	}
 
-	return plumbing.ZeroHash, false, err
-
+	return plumbing.ZeroHash
 }
 
 type commander struct {
--- a/funcbench/main.go
+++ b/funcbench/main.go
@@ -229,7 +229,7 @@ func startBenchmark(
 	}
 
 	// Get info about target.
-	targetCommit := getTargetInfo(ctx, env.Repo(), env.CompareTarget())
+	targetCommit := getTargetInfo(env.Repo(), env.CompareTarget())
 	if targetCommit == plumbing.ZeroHash {
 		return nil, fmt.Errorf("cannot find target %s", env.CompareTarget())
 	}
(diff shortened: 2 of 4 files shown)


nevill:
@geekodour PTAL

geekodour:
@nevill in reference to your comment from https://github.com/prometheus/test-infra/issues/373#issuecomment-639537578 , what change do we need in the eventmap? I think the regex would accept the tag name already.

geekodour (commented):
suggestion

funcbench/main.go
```diff
@@ -306,7 +306,11 @@ func getTargetInfo(ctx context.Context, repo *git.Repository, target string) (re
 		return commitHash, false, nil
 	}
 
-	targetRef, err := repo.Reference(plumbing.NewBranchReferenceName(target), false)
+	targetRef, _ := repo.Reference(plumbing.NewBranchReferenceName(target), false)
```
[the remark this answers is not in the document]
nevill: If using two new variables, it's like
```go
	var tagRef, branchRef *plumbing.Reference
	branchRef, _ = repo.Reference(plumbing.NewBranchReferenceName(target), false)
	if branchRef == nil {
		tagRef, err = repo.Reference(plumbing.NewTagReferenceName(target), false)
		if err != nil {
			return plumbing.ZeroHash, false, err
		}

		return tagRef.Hash(), false, err	
	}

	return branchRef.Hash(), false, nil
```

I'd prefer adding a comment to notice that it tries to find in branch names, then tag names. See my updated PR.

nevill:
In fact, I think we'd better change the signature of `getTargetInfo`. It doesn't need to return an error, if we don't find any reference, just return ZeroHash as reference. We can save some error checks in this way.

nevill:
> @nevill in reference to your comment from [#373 (comment)](https://github.com/prometheus/test-infra/issues/373#issuecomment-639537578) , what change do we need in the eventmap? I think the regex would accept the tag name already.

I meant to check if we can pass eventmap check. If yes, then everything is done. :)

funcbench/main.go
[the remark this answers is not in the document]
geekodour: I meant something like this:
```go
branchRef, _ := repo.Reference(plumbing.NewBranchReferenceName(target), false)
if branchRef != nil {
	return branchRef.Hash(), false, nil
}

tagRef, _ := repo.Reference(plumbing.NewTagReferenceName(target), false)
if tagRef != nil {
	return tagRef.Hash(), false, nil
}

return plumbing.ZeroHash, false, err
```
Think this is better than the nested if statements? o.O

> I think we'd better change the signature of getTargetInfo. It doesn't need to return an error

Yes, this sounds like a good idea, would you like to add the change in the same PR?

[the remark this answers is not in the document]
nevill: Yes, the way you suggested is great.
Now it's updated, `getTargetInfo` is also changed.

geekodour (commented):
Comment changes, otherwise LGTM. :+1: 

I think there's a bug in `getTargetInfo`, couple of reasons; git allows the same name for tag and branch, if a branch name has characters that are only in the hexadecimal set then it'll return the `commitHash` and never will go to `NewBranchReferenceName` part, that also has some additional catches. 

I think we need to have some kind of regex inside funcbench to differentiate between the three. eg. commit has to be the full commit and not the short commit id. For now differentiating between tag and branch can be left for now but might want to add some todo note about that aswell.

We can  probably do this in another PR aswell.

geekodour:
I just merged https://github.com/prometheus/test-infra/pull/379 the circleci tests for `make docs-check` were not running as expected. You'll need to run `make docs` for this PR as there are changes in the cli help flags.

funcbench/README.md
```diff
@@ -32,7 +32,7 @@ Flags:
       --nocomment            Disable posting of comment using the GitHub API.
       --owner="prometheus"   A Github owner or organisation name.
       --repo="prometheus"    This is the repository name.
-      --github-pr=GITHUB-PR  GitHub PR number to pull changes from and to post
+      --github-pr=GITHUB-PR  GitHub PR number to pull changes from and to pos
```
nevill: I don't know why `make docs` create a typo here, in the source it's `post`, can you take a look? @geekodour

geekodour: Could you try running `make build docs` ?might be some previous build in which the typo was there. I just tried running make docs on your PR, it's working as expected for me.

nevill: I tried several times, result the same typo.
I updated README manually. Let's get this PR done first.

geekodour:
LGTM, would like @krasi-georgiev to take a final look. :)

krasi-georgiev:
LGTM

I think it is now time to start adding tests for these behaviours, A good one here would be to pass some input data and check the expected error or output.

nevill:
Added with a new test `TestGetTargetInfo`.
Refactor on `getTargetInfo`, it utilizes [ResolveRevision](https://pkg.go.dev/gopkg.in/src-d/go-git.v4?tab=doc#Repository.ResolveRevision) now.

@geekodour please review it again :)

funcbench/main_test.go
```diff
@@ -15,6 +15,12 @@ package main
 
 import (
 	"testing"
+
+	fixtures "gopkg.in/src-d/go-git-fixtures.v3"
+	"gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/plumbing"
+	"gopkg.in/src-d/go-git.v4/plumbing/cache"
+	"gopkg.in/src-d/go-git.v4/storage/filesystem"
```
krasi-georgiev, line 23: this repo is archived, can't we use the new location?
https://github.com/go-git/go-git

krasi-georgiev:
> Added with a new test TestGetTargetInfo.

wow, that was quick, Thanks @nevill :)

funcbench/main_test.go
nevill: Yes, I will come with more PRs including that later. Since `go-git/go-git` also updates to v5.
It is enough for current PR right now.

funcbench/main.go
```diff
@@ -285,33 +289,14 @@ func interrupt(logger Logger, cancel <-chan struct{}) error {
...
-func getTargetInfo(ctx context.Context, repo *git.Repository, target string) (ref plumbing.Hash, compareWithItself bool, _ error) {
-	if target == "." {
-		return plumbing.Hash{}, true, nil
-	}
-
-	currRef, err := repo.Head()
+// getTargetInfo returns the hash of the target if found,
+// otherwise returns plumbing.ZeroHash.
+func getTargetInfo(repo *git.Repository, target string) plumbing.Hash {
+	hash, err := repo.ResolveRevision(plumbing.Revision(target))
```
geekodour, line 296: Great improvement! :partying_face: thanks for looking this up! might still want to add a todo note about the same target name for `tag` and `branch`, I tried running it a couple of times, it's always choosing the branch commit hash which can someday maybe change in go-git, so just a small note about it will be helpful. Think ready to merge after that :)

funcbench/main_test.go
```diff
@@ -49,3 +55,29 @@ BenchmarkBufferedSeriesIterator-8 0 0 +0.00%`
...
 	}
 }
+
+func TestGetTargetInfo(t *testing.T) {
+	_ = fixtures.Init()
+	f := fixtures.Basic().One()
+	sto := filesystem.NewStorage(f.DotGit(), cache.NewObjectLRUDefault())
+	r, err := git.Open(sto, f.DotGit())
+	if err != nil {
+		t.Errorf("error when open pepository: %s", err)
```
geekodour: typo:
```suggestion
		t.Errorf("error when open repository: %s", err)
```

geekodour (commented):
requesting small changes

funcbench/main_test.go
```diff
@@ -49,3 +55,29 @@ BenchmarkBufferedSeriesIterator-8 0 0 +0.00%`
...
+	}
+
+	datas := map[string]string{
```
geekodour: maybe a better name could be `targetHashPair` or simply `testCases`

nevill:
Updated

geekodour:
Thanks! :)

PR #378, opened 2020-06-05
Remove whitespace and post command syntax issues

### LINKED ISSUE #375: Post back error message when prombench / funcbench receive wrong input
Currently if a user input doesn't pass the regex check, everything is silently passing, user cannot know what is going on.

There should be a fix in comment monitor.

We need to post back an error message to tell the input has something wrong, when the comment is prefixed with `/prombench` or `/funcbench`.

geekodour:
* Added `extractCommand` to trim whitespaces
* Added `checkCommandPrefix` to check if it's a test-infra command and later
post a github comment in the PR if regex is not valid for the command.
* ~Added `COMMAND_PREFIXES` environment variable for specifying the command prefixes.~
* Added `--command-prefixes` flag for specifying the command prefixes.

Fixes: #375 

cc: @krasi-georgiev @nevill 

### DIFF
--- a/tools/commentMonitor/README.md
+++ b/tools/commentMonitor/README.md
@@ -41,7 +41,7 @@ For example, the following regex will create an argument named `RELEASE` with th
 #### Usage and examples:
 [embedmd]:# (commentMonitor-flags.txt)
 ```txt
-usage: commentMonitor --command-prefixes=COMMAND-PREFIXES [<flags>]
+usage: commentMonitor [<flags>]
 
 commentMonitor GithubAction - Post and monitor GitHub comments.
 
@@ -54,9 +54,8 @@ Flags:
       --eventmap="./eventmap.yml"
                         Filepath to eventmap file.
       --port="8080"     port number to run webhook in.
-      --command-prefixes=COMMAND-PREFIXES
-                        Comma separated list of command prefixes.
-                        Eg."/prombench,/funcbench"
+      --command-prefix=COMMAND-PREFIX ...
+                        Specify allowed command prefix. Eg."/prombench"
 
 ```
 ### Building Docker Image
--- a/tools/commentMonitor/README.md
+++ b/tools/commentMonitor/README.md
@@ -5,7 +5,6 @@ Currently it only works with [`issue_comment` event](https://developer.github.co
 
 ### Environment Variables:
 - `LABEL_NAME`: If set, will add the label to the PR.
-- `COMMAND_PREFIXES`: Comma separated list of command prefixes.
 - `GITHUB_TOKEN` : GitHub oauth token used for posting comments and settings the label.
 - Any other environment variable used in any of the comment templates in `eventmap.yml`.
 
@@ -20,9 +19,6 @@ Example content of the `eventmap.yml` file:
   regex_string: (?mi)^/prombench\s+cancel\s*$
   comment_template: |
     Benchmark cancel is in progress.
-```
-```shell
-$ export COMMENT_PREFIX='/funcbench,/prombench'
 ```
 
 If a GitHub comment matches with `regex_string`, then commentMonitor will trigger a [`repository_dispatch`](https://developer.github.com/v3/repos/#create-a-repository-dispatch-event) with the event type `event_type` and then post a comment to the issue with `comment_template`. The extracted out arguments will be passed to the [`client_payload`](https://developer.github.com/v3/repos/#example-5) of the `repository_dispatch` event.
@@ -45,7 +41,7 @@ For example, the following regex will create an argument named `RELEASE` with th
 #### Usage and examples:
 [embedmd]:# (commentMonitor-flags.txt)
 ```txt
-usage: commentMonitor [<flags>]
+usage: commentMonitor --command-prefixes=COMMAND-PREFIXES [<flags>]
 
 commentMonitor GithubAction - Post and monitor GitHub comments.
 
@@ -58,6 +54,9 @@ Flags:
       --eventmap="./eventmap.yml"
                         Filepath to eventmap file.
       --port="8080"     port number to run webhook in.
+      --command-prefixes=COMMAND-PREFIXES
+                        Comma separated list of command prefixes.
+                        Eg."/prombench,/funcbench"
 
 ```
 ### Building Docker Image
--- a/tools/commentMonitor/README.md
+++ b/tools/commentMonitor/README.md
@@ -5,6 +5,7 @@ Currently it only works with [`issue_comment` event](https://developer.github.co
 
 ### Environment Variables:
 - `LABEL_NAME`: If set, will add the label to the PR.
+- `COMMAND_PREFIXES`: Comma separated list of command prefixes.
 - `GITHUB_TOKEN` : GitHub oauth token used for posting comments and settings the label.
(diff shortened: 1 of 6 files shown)


tools/commentMonitor/README.md
```diff
@@ -14,12 +15,15 @@ Running commentMonitor requires the eventmap file which is specified by the `--e
...
 Example content of the `eventmap.yml` file:
-```
+```yaml
 - event_type: prombench_stop
   regex_string: (?mi)^/prombench\s+cancel\s*$
   comment_template: |
     Benchmark cancel is in progress.
 ```
+```shell
+$ export COMMENT_PREFIX='/funcbench,/prombench'
```
krasi-georgiev: I would much prefer to have these as cli args.
This way it would also be visible to the generated cli help readme.

tools/commentMonitor/main.go
```diff
@@ -155,10 +155,22 @@ func (c *commentMonitorConfig) webhookExtract(w http.ResponseWriter, r *http.Req
 			return
 		}
 
+		// Strip whitespace.
+		cmClient.extractCommand()
```
krasi-georgiev: It is my personal preference, but I would think this would be more readable like

```suggestion
		c.ghClient.commentBody = extractCommand(c.ghClient.commentBody)
```
the same with all other places

krasi-georgiev:
also it would be great to start adding tests. In this case it would be good to check that a wrong input returns expected errors and correct input parses it correctly.

geekodour:
> also it would be great to start adding tests. In this case it would be good to check that a wrong input returns expected errors and correct input parses it correctly.

Okay thanks, will add the tests in this PR itself. We could possibly also add the prefixes in the `eventmap.yml` itself but that'll change it's structure so maybe we can do that as an enhancement at some later point of time.

tools/commentMonitor/client.go
```diff
@@ -95,17 +72,17 @@ func (c commentMonitorClient) verifyUser(ctx context.Context, verifyUserDisabled
 }
 
 // Extract args if regexString provided.
-func (c *commentMonitorClient) extractArgs(ctx context.Context) error {
+func (c *commentMonitorClient) extractArgs(ctx context.Context, command string) error {
```
krasi-georgiev: I know this is not relevant to this or but let add `ctx` to the `githubClient` struct and use it from the `commentMonitorClient` instead of passing here as an arg to  `extractArgs`

tools/commentMonitor/client.go
```diff
@@ -115,6 +92,7 @@ func (c *commentMonitorClient) extractArgs(ctx context.Context) error {
 			return fmt.Errorf("%v: could not fetch SHA", err)
 		}
 
+		// TODO (geekodour) : We could run this in a seperate method.
 		err = c.ghClient.createRepositoryDispatch(ctx, c.eventType, c.allArgs)
```
krasi-georgiev: the same here , instead of passinge ctx lets just use it from `ghClient`

tools/commentMonitor/main.go
```diff
@@ -32,6 +32,7 @@ type commentMonitorConfig struct {
 	verifyUserDisabled bool
 	eventMapFilePath   string
 	whSecretFilePath   string
+	commandPrefixes    string
```
nevill: Better to declare `commandPrefixes` as `[]string`.
In this way, we don't need to split it on every request during `checkCommandPrefix`.

May `StringsVar` in kingpin help.

nevill (changes requested):

tools/commentMonitor/main.go
geekodour: Yes, that's better but what do you think of putting the prefixes in the eventmap file like this, might have to rename `eventmap.yml` to something like `config.yml`:

```yaml
prefixes:
  - /prombench
  - /funcbench
eventmap:
  - event_type: event2
    regex_string: some_regex
    comment_template: |
      Dummy funcbench start
  - event_type: event1
    regex_string: some_regex
    comment_template: |
      Cool
```

nevill: It makes sense to rename. But do you know the reason why we `loadConfig` on every request?
If it's a static part, we'd better preprocess it as much as possible, and not mix it with dynamical part, such as validation, extracting args.

geekodour: @nevill  : https://github.com/prometheus/test-infra/pull/307#discussion_r350567334

nevill: I would argue that, even eventmap.yml is cached, the function call to unmarshal is not.
Plus, if we use config.yml, should we put other configs webhooksecret / port etc into it either?
I think we'd keep prefixes as the input parameters from cli.

geekodour: Okay, will update it for multiple `--command-prefix` (possibly with some short flag) then for now, can work on the other issues in another PR if needed.

> if we use config.yml, should we put other configs webhooksecret / port etc into it either?

No I think the prefix is related to the github comment command and the regex, so they make sense to be somewhere near where the regexes for the command is defined. (currently `eventmap.yml`) but other things like webhook secret and port are specific to how we run comment-monitor. Maybe `config.yml` is not the best name for the kind of file I am trying to describe, we can still keep the name `eventmap.yml` but if we include prefixes in it, it won't be a good name for the content of the file imo.

krasi-georgiev: if we have a small number of cli args I would say to avoid the need of a config file.

on a similar note do we even need the eventmap parsing at all?

geekodour: Added the changes, 

>  do we even need the eventmap parsing at all?

It think we'll be adding more commands so I guess having a file for storing the event types and regex is a good idea. About loading it on each http request, I think we don't really need that unless we're making changes to the file often and we can't afford any downtime for commentMonitor which currently is not the case. Were you suggesting something else?

krasi-georgiev: > I guess having a file for storing the event types and regex is a good idea.

Could you mention few benefits of keeping this file?

I can't think of anything important to be worth keeping it.
IIRC it was needed only because of how GH actions work, but now that the tool doesn't run inside GH actions don't see a reason to keep it.

I would say up to 10-15 cli args it is absolutely fine and don't require having a config file.

geekodour: As per this PR (https://github.com/prometheus/test-infra/pull/207) we were just using the [`regex` from the command line](https://github.com/nikita-mk/prombench/blob/f918ceb908bf7c00b00756169fa94d8bc6792ce3/cmd/tools/commentMonitor/main.go#L23) when using it inside github actions and the command template was set via env var later on.

But as we move away from github actions for comment parsing needs, we need to specify three things for each event, i.e `event_type`,`regex_string` and `comment_template`, now `label`(after #381 ) I think a file to manage this is simpler than having to pass these long comment strings as env vars or command line arguments. When using github actions, we were using the `event.json` file which was basically a dump of the `IssueComment` event webhook response.

`eventmap.yml` used for comment-monitor is different:
https://github.com/prometheus/test-infra/blob/master/prombench/manifests/cluster-infra/7a_commentmonitor_configmap_noparse.yaml

krasi-georgiev: ooh , yes I see what you mean now, these are too long for cli args, no other comments from my side, feel free to merge when you think it is ready

PR #392, opened 2020-06-28
Added k8s secret for GITHUB_TOKEN.

geekodour:
Currently token is being passed as is:
https://github.com/prometheus/test-infra/blob/6f26e55f5899716114ac83d4d106b7b01edef4b8/funcbench/manifests/benchmark/2_job.yaml#L28

This PR creates a k8s secret and sets that to the appropriate env var.

### DIFF
--- a/funcbench/Makefile
+++ b/funcbench/Makefile
@@ -32,4 +32,5 @@ resource_delete:
 		-v GITHUB_ORG:${GITHUB_ORG} -v GITHUB_REPO:${GITHUB_REPO} \
 		-v BRANCH:${BRANCH} -v BENCH_FUNC_REGEX:${BENCH_FUNC_REGEX} \
 		-f manifests/benchmark/3_job.yaml \
+		-f manifests/benchmark/2_secrets.yaml \
 		-f manifests/benchmark/1_namespace.yaml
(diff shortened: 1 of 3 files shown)


funcbench/Makefile
```diff
@@ -28,8 +28,8 @@ resource_apply:
 resource_delete:
 	$(INFRA_CMD) gke resource delete -a ${AUTH_FILE} \
 		-v ZONE:${ZONE} -v PROJECT_ID:${PROJECT_ID} -v CLUSTER_NAME:funcbench-${PR_NUMBER} \
-		-v PR_NUMBER:${PR_NUMBER} -v GITHUB_TOKEN:${GITHUB_TOKEN} \
+		-v PR_NUMBER:${PR_NUMBER} \
 		-v GITHUB_ORG:${GITHUB_ORG} -v GITHUB_REPO:${GITHUB_REPO} \
 		-v BRANCH:${BRANCH} -v BENCH_FUNC_REGEX:${BENCH_FUNC_REGEX} \
-		-f manifests/benchmark/2_job.yaml \
+		-f manifests/benchmark/3_job.yaml \
```
nevill, line 34: Shall we also delete the secrets here?

geekodour: thanks, i'll add that. do we even need a `resource_delete` now since we're deleting the entire cluster. will check once.

nevill: Well, thought about it for a while. I'd say let's keep `resource_delete`, otherwise I'll be curious about why `resource_apply` doesn't have a corresponding removal task some day.

krasi-georgiev:
LGTM

PR #393, opened 2020-06-28
Update src-d/go-git.v4 to go-git/go-git/v5

nevill:
since go-git moved into a new GitHub org.


nevill:
@geekodour PTAL

geekodour:
Thanks!

PR #399, opened 2020-06-29
Sort by magnituted and show only changed

krasi-georgiev:
cc @bwplotka , any objections to this?


nevill:
Also change it in `func (g *GitHub) PostResults(ctx context.Context, cmps []BenchCmp)` around L137.

Issue #398, opened 2020-06-29
[funcbench] when a comment doesn't match any tests should also post a comment


krasi-georgiev:
At the moment if I post `/funcbench master TestDoesntExist` There is no feedback that this test doesn't exist. It would be useful to post a comment like: `Test expression ... didn't match any existing tests. Available command formats....`

nevill:
@krasi-georgiev This one is also can be closed since #400 gets merged.

krasi-georgiev:
thanks, in the future can use the PR keywords to close issues from PRs when merged. https://docs.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords#:~:text=issue%20will%20close.-,Closing%20an%20issue%20in%20the%20same%20repository,merged%20into%20the%20default%20branch.

nevill:
Yes I see. In fact I did it in the commit message [there](https://github.com/nevill/test-infra/commit/dd7601e81bd21a73ce37408d41f3bfc128751052), but seems it's squashed and gone away.

geekodour:
Closed by #400

PR #400, opened 2020-06-30
Use benchstat to compare the benchmark results

### LINKED ISSUE #389: Funcbench - make the minimum display unit in milisecond
In benchmark results, we think 200 ns/op is not a big difference as 100,000 ns/op( = 0.1 milisecond ). We should display them as 0 ms/op.

nevill:
This should fix #389, #398.

By using benchstat, the significance of benchmark results can scale based on the [Mann-Whitney U test](https://pkg.go.dev/golang.org/x/perf@v0.0.0-20200318175901-9c9101da8316/benchstat?tab=doc#UTest).
It reduces a lot noise in the results.

@geekodour @krasi-georgiev Please take a review.

PR includes:
- Will return an error with `didn't match any existing benchmarks` (#398)
- Use benchstat for comparing the benchmarks (#389)
- Calling benchstat.Table directly to format the result in markdown
- Refactor, stop passing context around too much


funcbench/bench.go
```diff
@@ -77,25 +76,28 @@ func (b *Benchmarker) benchOutFileName(commit plumbing.Hash) (string, error) {
...
 }
 
-func (b *Benchmarker) exec(ctx context.Context, pkgRoot string, commit plumbing.Hash) (out string, err error) {
+func (b *Benchmarker) exec(pkgRoot string, commit plumbing.Hash) (string, error) {
 	fileName, err := b.benchOutFileName(commit)
 	if err != nil {
 		return "", err
 	}
 
+	// try to get result from cache
```
krasi-georgiev: Please follow Golang commentary guidelines.
https://golang.org/doc/effective_go.html#commentary

krasi-georgiev (commented):
Nice!!!

I will let @geekodour review and approve this one

funcbench/env.go
```diff
@@ -128,19 +138,27 @@ func newGitHubEnv(ctx context.Context, e environment, gc *gitHubClient, workspac
...
+	}
+
 	legend := fmt.Sprintf("Old: `%s`\nNew: `PR-%d`", g.compareTarget, g.client.prNumber)
-	result := fmt.Sprintf("<details><summary>Click to check benchmark result</summary>\n\n%s\n%s\n%s</details>", legend, strings.Join(extraInfo, "\n"), formatCommentToMD(b.String()))
-	return g.client.postComment(ctx, result)
+	result := fmt.Sprintf(
+		"<details><summary>Click to check benchmark result</summary>\n\n%s\n%s\n%s</details>",
+		legend,
+		strings.Join(extraInfo, "\n"),
+		b.String(),
```
geekodour, line 165: For some reason the results are not getting posted in the comment: 
https://github.com/prometheus-community/prometheus/pull/69#issuecomment-652632042

nevill: I think this is a special case, you tested a benchmark existed in your PR, but not on branch master. I am testing against `BenchmarkRangeQuery` now, soon will post the benchmark result.

geekodour: Ah, so in that case if we actually ran funcbench on https://github.com/prometheus/prometheus/pull/7332  it would not show us the expected results. That's one of the usecase funcbench is supposed to be used for, no? So if there are new benchmarks added should it copy those to to the target branch/commit, ah i can already see this going little bit crazy. So unless, the PR is merged funcbench won't show expected results on https://github.com/prometheus/prometheus/pull/7332 , is that right?

cc: @krasi-georgiev

geekodour: also related to #398 , currently if there are no matching benchmarks, funcbench just runs and exits with 0.

geekodour: @nevill if we're not resolving this here, I think exiting with error if empty will be better than posting empty results.

funcbench/env.go
```diff
@@ -63,11 +64,17 @@ func newLocalEnv(e environment) (Environment, error) {
...
+func (l *Local) PostErr(string) error { return nil } // Noop. We will see error anyway.
 
-func (l *Local) PostResults(ctx context.Context, cmps []BenchCmp, extraInfo ...string) error {
+func (l *Local) PostResults(tables []*benchstat.Table, extraInfo ...string) error {
 	fmt.Println("Results:")
-	Render(os.Stdout, cmps, true, true, l.compareTarget)
+	var buf bytes.Buffer
+	benchstat.FormatText(&buf, tables)
+
+	if _, err := os.Stdout.Write(buf.Bytes()); err != nil {
```
geekodour: Could we not just use `fmt.Println` here?

geekodour (approved):

funcbench/env.go
nevill: For sure, newly created benchmarks cannot have anything to compare with. We can post comments like `no matching benchmark to compare with` back.

funcbench/env.go
nevill: Using `os.Stdout` to print a buffer is much straight forword.
`fmt.Println` will call `os.Stdout.Write` in the end. Plus, have to covert from []byte to string, and string back to []byte.

nevill: If we're annoyed with the error check, it can be saved here, just like normal `fmt.Println` doesn't check the error.

funcbench/main.go
```diff
@@ -248,38 +246,39 @@ func startBenchmark(
...
 	}
 
 	// Compare B vs A.
-	cmps, err := bench.compareBenchmarks(oldResult, newResult)
+	tables, err := bench.compareBenchmarks(oldResult, newResult)
 	if err != nil {
 		return nil, errors.Wrap(err, "comparing benchmarks")
 	}
-	return cmps, nil
+	return tables, nil
```
geekodour, line 290: Before returning could check if `tables` is empty and return the error that the benchmark might be missing in one or both of the branches. This will also resolve https://github.com/prometheus/test-infra/issues/398 if i am not wrong.

funcbench/main.go
```diff
@@ -307,10 +306,11 @@ func getTargetInfo(repo *git.Repository, target string) plumbing.Hash {
 
 type commander struct {
 	verbose bool
+	ctx     context.Context
```
geekodour, line 309: Is it okay to have the context in the struct for `commander`? what if we want to cancel the context from something that's using it. I am a little unsure about this. :(

nevill:
Now #398 is also solved. You may take a look at a sample [here](https://github.com/prometheus-community/prometheus/pull/69#issuecomment-655622006).

funcbench/main.go
```diff
@@ -151,27 +152,35 @@ func main() {
...
 			// ( ◔_◔)ﾉ Start benchmarking!
-			benchmarker := newBenchmarker(logger, env, &commander{verbose: cfg.verbose}, cfg.benchTime, cfg.benchTimeout, cfg.resultsDir)
-			cmps, err := startBenchmark(ctx, env, benchmarker)
+			benchmarker := newBenchmarker(logger, env, &commander{verbose: cfg.verbose, ctx: ctx}, cfg.benchTime, cfg.benchTimeout, cfg.resultsDir)
+			tables, err := startBenchmark(env, benchmarker)
 			if err != nil {
-				if pErr := env.PostErr(ctx, fmt.Sprintf("%v. Benchmark failed, please check logs.", err)); pErr != nil {
+				pErr := env.PostErr(
+					fmt.Sprintf(
+						"`%s`\n%s.",
```
geekodour, line 168:
```suggestion
						"```%s```\n%s.",
```

funcbench/bench.go
```diff
@@ -108,77 +108,30 @@ func (b *Benchmarker) exec(ctx context.Context, pkgRoot string, commit plumbing.
...
-		return nil, errors.Wrapf(err, "open %s", beforeFile)
-	}
-
-	after, err := parseFile(afterFile)
-	if err != nil {
-		return nil, errors.Wrapf(err, "open %s", afterFile)
-	}
-
-	// Use benchcmp library directly.
-	// TODO(bwplotka): benchstat is new thing - we might add choice for funcbench to choose from? (:
```
geekodour, line 136: Comparison: 

> note: this benchmark results are almost the same as there were not a lot of changes related to the benchmark being added.

Funcbench runs using this PR(`benchstat`):
- run1: https://github.com/prometheus-community/prometheus/pull/71#issuecomment-656452591
- run2: https://github.com/prometheus-community/prometheus/pull/71#issuecomment-656469104

Funcbench runs using master(`benchcmp`):
- run1: https://github.com/prometheus-community/prometheus/pull/71#issuecomment-656476903
- run2: https://github.com/prometheus-community/prometheus/pull/71#issuecomment-656482497

I think this PR adds some really nice stuff, but concerned about @bwplotka's comment on supporting both of them. I was wondering if we still need to keep compatibility with `benchcmp`? 

cc: @bwplotka

geekodour (commented):
Looks good :) Thanks!

waiting for @krasi-georgiev and @bwplotka 's final inputs.

funcbench/bench.go
bwplotka: benchstat is better but I use `-delta=none` (:

funcbench/main.go
```diff
@@ -215,13 +220,14 @@ func startBenchmark(
 		return nil, errors.Wrap(err, "get head")
 	}
 
-	if _, err := bench.c.exec(ctx, "sh", "-c", "git update-index -q --ignore-submodules --refresh && git diff-files --quiet --ignore-submodules --"); err != nil {
+	// TODO move it into env? since GitHub env doesn't need this check.
```
bwplotka, line 223: Why does not need? I think there is no harm, no?

bwplotka (approved):
Looks good, but I did not look very deeply (: 

Happy to merge this :+1: 
 `benchstat` FTW (:

funcbench/main.go
nevill: Because in GitHub env, the repo is always newly cloned, we presume that worktree is always clean.

Issue #397, opened 2020-06-29
[funcbench] for comments with incorrect syntacs post back a comment with examples


krasi-georgiev:
at the moment when I post a /funcbench comment with incorrect syntax it just replies with a comment that the syntax is incorrect. It would be a better experience to post back all available formats that are accepted.

geekodour:
This is posted back by the comment-monitor, it can post back the regex that it expects to be matched to but posting the available formats will need changing comment-monitor. should we just return the expected regex for now?

nevill:
I think this issue should be fixed in comment monitor instead of funcbench, prombench can also be affected by this.

krasi-georgiev:
how about having 

comment_template_success: .....
comment_template_failure: .....

the comment_template_failure needs to be auto-generated so that it doesn't diverge from the funbench cli and we should also have a test for this. Check that the generated examples match exactly what funcbench expects..

geekodour:
closed by #403

PR #403, opened 2020-07-04
Add support for help text and minor fixes for comment-monitor

### LINKED ISSUE #394: [funcbench] Deleting a comment triggers another test
when deleting a comment that was used to trigger a test it re-triggers the same test so it seems that the comment monitor doesn't distinguish between delete and create events.

geekodour:
Fixes #394 and #397 

* Removed `--command-prefix` flag, by adding `prefixes` to `config.yml`
* `config.yml` which replaces `eventmap.yml`
* Added help text for each prefix from the config.yml file (#397)
* `repository_dispatch` event dispatch sent only on creation of `issue_comment` event (#394 )

Todo:
- [x] Update the manifest file

cc :  @krasi-georgiev @nevill

### DIFF
--- a/tools/commentMonitor/README.md
+++ b/tools/commentMonitor/README.md
@@ -5,31 +5,38 @@ Currently it only works with [`issue_comment` event](https://developer.github.co
 
 ### Environment Variables:
 - `GITHUB_TOKEN` : GitHub oauth token used for posting comments and settings the label.
-- Any other environment variable used in any of the comment templates in `eventmap.yml`.
+- Any other environment variable used in any of the comment templates in `config.yml`.
 
 ### Setting up the webhook server
-Running commentMonitor requires the eventmap file which is specified by the `--eventmap` flag.
+To specify the config file for commentMonitor use the the `--config` flag.
 
-The `regex_string`, `event_type` and `comment_template` can be specified in the `eventmap.yml` file.
-
-Example content of the `eventmap.yml` file:
+Example `config.yml` file:
 ```yaml
-- event_type: prombench_stop
-  regex_string: (?mi)^/prombench\s+cancel\s*$
-  comment_template: |
-    Benchmark cancel is in progress.
-  label: prombench
+prefixes:
+  - prefix: /prombench
+    help_template: |
+      Get prombench syntax help here.
+  - prefix: /funcbench
+    help_template: |
+      Get funcbench syntax help [here](https://canbealink).
+eventmaps:
+  - event_type: prombench_stop
+    regex_string: (?mi)^/prombench\s+cancel\s*$
+    comment_template: |
+      Benchmark cancel is in progress.
+    label: prombench
 ```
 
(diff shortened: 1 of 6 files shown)


geekodour:
from @krasi-georgiev 's comment https://github.com/prometheus/test-infra/issues/397#issuecomment-651669315

`comment_template_failure` won't work because the eventmap is used only after the regex is matched, when it is not matched it'll never know which eventmap was it. hence added `help_template` which is specific to the prefix.

tools/commentMonitor/README.md
```diff
@@ -5,31 +5,38 @@ Currently it only works with [`issue_comment` event](https://developer.github.co
...
-    Benchmark cancel is in progress.
-  label: prombench
+prefixes:
+  - prefix: /prombench
+    help_template: |
+      Get prombench syntax help here.
+  - prefix: /funcbench
+    help_template: |
+      Get funcbench syntax help [here](https://canbealink).
+eventmaps:
```
krasi-georgiev, line 22:
```suggestion
events:
```

krasi-georgiev (commented):
Nice that would work!

We should also update funcbench to post back a comment when the bench arg doesn't match any tests but can be a separate PR just need to create a separate issue for this. I remember trying to start a tests and nothing happened, I didn't know whether the test took too long or nothing was triggered.

tools/commentMonitor/README.md
```diff
@@ -5,31 +5,38 @@ Currently it only works with [`issue_comment` event](https://developer.github.co
...
+Example `config.yml` file:
 ```yaml
-- event_type: prombench_stop
-  regex_string: (?mi)^/prombench\s+cancel\s*$
-  comment_template: |
-    Benchmark cancel is in progress.
-  label: prombench
+prefixes:
+  - prefix: /prombench
+    help_template: |
```
krasi-georgiev, line 17:
```suggestion
    help: |
```

tools/commentMonitor/README.md
```diff
@@ -5,31 +5,38 @@ Currently it only works with [`issue_comment` event](https://developer.github.co
...
+      Get prombench syntax help here.
+  - prefix: /funcbench
+    help_template: |
+      Get funcbench syntax help [here](https://canbealink).
+eventmaps:
+  - event_type: prombench_stop
+    regex_string: (?mi)^/prombench\s+cancel\s*$
+    comment_template: |
+      Benchmark cancel is in progress.
```
krasi-georgiev, line 26: I can't remember. Why do we need this? I thought the status checks block of github is changing to give feedback for this no?

 Update status to failure
  Update status to success

geekodour: We could do without this, but the flow from `github comment - comment-monitor sending repository_dispatch - github-action triggering` takes a while(sometimes github actions schedules the actions slowly sometime faster etc). Having comment monitor post this comment ensures that the commenter will get a quick response and they can move on to other work without worrying :)

geekodour:
> We should also update funcbench to post back a comment when the bench arg doesn't match any tests 

We already have #398 for that and #400 should fix it. :)

tools/commentMonitor/README.md
krasi-georgiev: go it

krasi-georgiev:
ok thanks, then just these small comments and feel free to merge so we can test it.

tools/commentMonitor/README.md
geekodour: I was wondering if this change is required because this is also a golang template like `comment_template`? please let me know.

krasi-georgiev:
LGTM

geekodour:
Tested, working as expected. Merging :)

Issue #385, opened 2020-06-19
Add funcbench logs to Github comment


geekodour:
Funcbench logs can only be seen in GKE stackdriver before the funcbench Job is deleted. It'll be nice to have these logs posted as part of the result or error comment posted back to Github.

roidelapluie:
Or in Loki

bwplotka:
FYI: We decided to remove `funcbench` for now from this repo, so this is no longer relevant. Thanks!

PR #405, opened 2020-07-05
Add support to specify package and refactors to funcbench

### LINKED ISSUE #396: [funcbench] allow running package wide benchmarks.
Make something like this possible in funcbnech

```go test -benchmem -run=^$ github.com/prometheus/prometheus/promql -bench .```

### LINKED ISSUE #404: Single quote CLI regexes when running in funcbench
https://github.com/prometheus/test-infra/blob/3374b562904be593592a03609dba71bf4dc3c14a/funcbench/bench.go#L54

This currently is not an issue, but if one tries to copy the command from the comment and tries to run it locally, the lack of quotes can be an issue.

### LINKED ISSUE #410: Funcbench fails to fetch github repo sometimes
It'll be nice to have some sort of retry timeout to how we're fetching the repository to run funcbench on. The subsequent runs after this comment were run without any changes and they just worked.

Example: https://github.com/prometheus-community/prometheus/pull/71#issuecomment-656461942

This seems more like a network issue, to me but unsure as it not always reproducible.

### PR DESCRIPTION
**Pull request by geekodour**
Fixes #396 
Fixes #404 
Fixes #410 

Adds ability to specify package with `PACKAGE_NAME`, also added some ui changes like posting the commit hash with the results plus some README improvements.

TODO:
- [x] Update workflow file
- [x] Update README and help text
- [x] Need to test this once

### CODE REVIEW

**Comment by nevill** [comment]
Do you think of omitting the `./` there in the comment, e.g. `/funcbench v2.16 Benchmark.* promql`

**Comment by geekodour** [comment]
@nevill 

> Do you think of omitting the ./ there in the comment, e.g. /funcbench v2.16 Benchmark.* promql


I think it'll be fine to have `./` because otherwise, think of a package path of `something/something` vs `./something/something` the second one is pretty obvious that it's taking the path to the package in the repository.

what do you suggest?

**Comment by nevill** [comment]
> what do you suggest?

OK, you example (also in the code) makes me recall that we talked about supporting `./...`, it makes sense to use `./`.

File: funcbench/README.md [hunk 1]
```diff
@@ -64,21 +66,30 @@ docker build -t prominfra/funcbench:master .
...
 
-- `/funcbench master BenchmarkQuery.*` - compare all the benchmarks mathching `BenchmarkQuery.*` for branch master vs the PR.
 
-- `/funcbench feature-branch` or `/funcbench feature-branch .*` - compare all the benchmarks on feature-branch vs the PR.
+|Command|Explanation|
+|---|--|
+|`/funcbench master BenchmarkQuery.*`| Compare all the benchmarks matching `BenchmarkQuery.*` for master vs the PR|
+|`/funcbench feature-branch` or `/funcbench tag-name .*`| Compare all the benchmarks on feature-branch/tag-name vs the PR|
+|`/funcbench master BenchmarkQuery.* ./tsdb` | Compare all the benchmarks matching `BenchmarkQuery.*` for master vs the PR in package `./tsdb`|
+|`/funcbench master Benchmark(?:Isolation.*|QuerierSelect) ./tsdb` | Compare all benchmarks matching `Benchmark(?:Isolation.*|QuerierSelect)` for master vs the PR|
```
geekodour: GitHub markdown is not rendering is as expected because of the `|`

funcbench/README.md
```diff
@@ -64,21 +66,30 @@ docker build -t prominfra/funcbench:master .
 ```
 
 ## Triggering with GitHub comments
+<!-- If you change the heading, please change the anchor at 7a_commentmonitor_configmap_noparse.yaml aswell. -->
 
 The benchmark can be triggered by creating a comment in a PR which specifies a branch to compare. The results are then posted back to the PR as a comment. The Github Actions workflow for funcbench [can be found here](https://github.com/prometheus/prometheus/blob/master/.github/workflows/funcbench.yml).
 
-The syntax is: `/funcbench <branch> <benchmark function regex>`
+The syntax is: `/funcbench <branch/tag/commit> <benchmark function regex>`.
```
```suggestion
The syntax is: `/funcbench <branch|tag|commit> <benchmark function regex>`.
```

a bit more obvious that it is an OR, otherwise branch/tag/commit looks like a path

funcbench/README.md
```diff
@@ -64,21 +66,30 @@ docker build -t prominfra/funcbench:master .
...
-  ```
-  /funcbench old_branch .*
-
-  The old_branch performs poorly, I bet mine are much better.
-  ```
+> **Notes:**
+>
+> - Editing/Deleting a comment will not re-trigger the workflow of starting/stopping a benchmark. Only creating a comment starts a benchmark.
+> - In case of funcbench, it automatically cleans up. So, no explicit stop command required.
+> - Multiline comments are allowed:
```
```suggestion
> - Multiline comment lines are allowed:
```

funcbench/env.go
```diff
@@ -89,10 +103,24 @@ type GitHub struct {
...
+
+	var r *git.Repository
+	var err error
+	// Retry 10 times.
+	for i := 1; i <= 10; i++ {
+		if err := os.RemoveAll(filepath.Join(workspace, gc.repo)); err != nil {
+			return nil, err
+		}
+		e.logger.Println("Cloning ", gc.owner, ":", gc.repo, " is in progress. Checking in ", provider.GlobalRetryTime)
+		time.Sleep(provider.GlobalRetryTime)
```
krasi-georgiev: This is a different Retry and shouldn't be mixed with when applying resources.

provider.GlobalRetryTime - this is how long to wait for retrying to apply a resource
and the retry here is how long to wait for cloning a repo. 

I would say lets keep these separate and have this hard coded or a separate env just for this package.

Also always think about dependencies. There is absolutely no reason why would the funcbench package to depend on the provider package.

funcbench/env.go
```diff
@@ -157,7 +185,12 @@ func (g *GitHub) PostResults(tables []*benchstat.Table, extraInfo ...string) err
 		return err
 	}
 
-	legend := fmt.Sprintf("Old: `%s`\nNew: `PR-%d`", g.compareTarget, g.client.prNumber)
+	legend := fmt.Sprintf("Old: `%s`/`%s`\nNew: `PR-%d`/`%s`",
```
krasi-georgiev: I would say always use %v, this way if you change the env type don't have to think about updating the message. Fewer things to think about.

prombench/README.md
```diff
@@ -94,6 +94,7 @@ cat $AUTH_FILE | base64 -w 0
 ## Usage
 
 ### Trigger tests via a Github comment.
+<!-- If you change the heading, please change the anchor at 7a_commentmonitor_configmap_noparse.yaml aswell. -->
```
krasi-georgiev:
```suggestion
<!-- If you change the heading, also change the anchor in the comment monitor config map. -->
```

no need to be that specific. Error prone as filenames change quite often and there is no way you would remember to update the filename here as well.

funcbench/main.go
```diff
@@ -332,7 +343,12 @@ func (c *commander) exec(command ...string) (string, error) {
 	if err := cmd.Run(); err != nil {
 		out := b.String()
 		if c.verbose {
-			out = ""
+			_, ok := c.env.(*GitHub)
+			if ok {
+				out = fmt.Sprintf("<details><summary>Error:</summary>\n```\n%s\n```\n</details>", out)
```
krasi-georgiev: Why not just:
````

if out != "" {
     out = fmt.Sprintf("<details><summary>Error:</summary>\n```\n%s\n```\n</details>", out)
}
```

krasi-georgiev (commented):
not related to this PR, but could you add some comments for the  env interface to explain each method and that it represents if funcbench is run in github or local mode or environment.

funcbench/main.go
geekodour: Since it'll be slightly complicated to have loki for funcbench since it runs on a different cluster( #385 ) I thought it'll be nice to have the command line error on the github comment aswell.

Previously, it just set `out` to an empty string when run in verbose mode. (we run with verbose flag on when running funcbench on Github mode)

krasi-georgiev: sure I understand the logic, but `_, ok := c.env.(*GitHub)` looks like a design flow. 

Maybe extend the interface to help you identify when running in local mode or github mode?

geekodour: Yes, that seemed a bit odd. I updated this PR to include the error output in verbose mode aswell. It doesn't add a lot of noise so it should be fine.

Plus now it will always then give us the error message(run in verbose flag set or not) when running in GitHub mode.

Will test this and post example comments.

krasi-georgiev:
LGTM

PR #406, opened 2020-07-06
Converting ingress service type as deployment var and adding a infra CLI command to show deployment vars

imrajdas:
Motivation:
- As KIND doesn't support Loadbalance type service, That's why it converts ingress service type as deployment var (PROVIDER_SERVICE) and assigning that deployment var in NewGKEClient function of gke.go 

Changes:
- Adding NGINX_SERVICE variable for the load balancer service type
- Adding `gke info deploymentvars` command to show all default deployment variables.


imrajdas:
cc @geekodour

imrajdas:
Logs-
[log.txt](https://github.com/prometheus/test-infra/files/4880617/log.txt)

pkg/provider/gke/gke.go
```diff
@@ -77,6 +77,10 @@ type GKE struct {
 
 // NewGKEClient sets the GKE client used when performing GKE requests.
 func (c *GKE) NewGKEClient(*kingpin.ParseContext) error {
+
+	// Set PROVIDER_SERVICE Deployment variable to  LoadBalancer
+	c.DeploymentVars["PROVIDER_SERVICE"] = "LoadBalancer"
```
geekodour:
```suggestion
	c.DeploymentVars["NGINX_SERVICE_TYPE"] = "LoadBalancer"
```

In the provider specific docs, we could just add a section about what are the deployment variables that are automatically passed for this specific provider. 

It'll be good if we could have these deployment variables overridable when uses passes the same using the `-v` flag. Maybe also have a provider specific flag to list out all the set deployment variables. (This way we won't have to update readme manually and we could just use embedmd https://github.com/prometheus/test-infra/pull/353)

What I think will be good is to have a default `DeploymentVars` and have it set to the default value, like for eg. in this example only nginx controller k8s service type needs to be changed when running in KIND (to `CLusterIP`).

But for EKS and GKE, the type `LoadBalancer` works, so we could have some defaults and only modify them when creating the provider that needs the change(in our case KIND)

geekodour (commented):
What I was suggesting was something like  
```
c.DeploymentVars["PROVIDER_SERVICE"] = "GKE" 
``` 
and use something like 
```
{{ if (eq .PROVIDER_SERVICE "GKE") }}
```
in the manifest file, but the approach in this PR actually looks better i think :) . I've commented some ways this can be improved.

cc @krasi-georgiev @nevill

geekodour:
@rajdas98  Also ingress in [kind](https://kind.sigs.k8s.io/docs/user/ingress/) uses 
https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/static/provider/kind/deploy.yaml

And [GKE](https://kubernetes.github.io/ingress-nginx/deploy/#gce-gke) currently uses
https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/static/provider/cloud/deploy.yaml

Difference being :
```diff
λ diff gke_deploy.yml kind_deply.yml 
287,288c287
<   type: LoadBalancer
<   externalTrafficPolicy: Local
---
>   type: NodePort
322a322,325
>   strategy:
>     rollingUpdate:
>       maxUnavailable: 1
>     type: RollingUpdate
343d345
<             - --publish-service=ingress-nginx/ingress-nginx-controller
349a352
>             - --publish-status-address=localhost
390a394
>               hostPort: 80
393a398
>               hostPort: 443
404a410,415
>       nodeSelector:
>         ingress-ready: 'true'
>       tolerations:
>         - effect: NoSchedule
>           key: node-role.kubernetes.io/master
>           operator: Equal
406c417
<       terminationGracePeriodSeconds: 300
---
>       terminationGracePeriodSeconds: 0
```

So, is just changing the service type enough?

geekodour:
So as far as i can see there should be three levels to this:

- default deployment vars (which get set provider agnostic)

- provider deployment vars which has a different value than default deployment var Eg. (Loadbalancer in gke and Nodeport in kind) Here the default could be Loadbalancer as it's used by both gke and eks and we need to replace it only for KIND

- the normal deployment vars at use, this should be able to override both default and provider specific deployment vars.

infra/README.md
```diff
@@ -23,6 +23,9 @@ Commands:
   help [<command>...]
     Show help.
 
+  gke show deploymentvars
+    gke show deploymentvars -a service-account.json
```
geekodour: Maybe `infra gke info` will be a better command for this :) what do you think? also would be nice if we could do this without having to pass the auth file. This should become easier after https://github.com/prometheus/test-infra/pull/413/

infra/infra.go
```diff
@@ -38,13 +38,17 @@ func main() {
...
 		StringVar(&g.Auth)
 	k8sGKE.Flag("file", "yaml file or folder  that describes the parameters for the object that will be deployed.").
-		Required().
 		Short('f').
 		ExistingFilesOrDirsVar(&g.DeploymentFiles)
 	k8sGKE.Flag("vars", "When provided it will substitute the token holders in the yaml file. Follows the standard golang template formating - {{ .hashStable }}.").
 		Short('v').
 		StringMapVar(&g.DeploymentVars)
 
+	k8sGKEShow := k8sGKE.Command("show", "show GKE related infomation").
```
geekodour: We should probably just have one command, also a short flag would be nice. :sparkles:

pkg/provider/gke/gke.go
```diff
@@ -44,8 +44,12 @@ import (
 
 // New is the GKE constructor.
 func New() *GKE {
+	DeploymentVars := make(map[string]string)
+	// Set PROVIDER_SERVICE Deployment variable to  LoadBalancer
```
geekodour: Does `PROVIDER_SERVICE` need to be something different here ? Also please end comments with a full stop :)

pkg/provider/gke/gke.go
```diff
@@ -577,3 +581,12 @@ func (c *GKE) ResourceDelete(*kingpin.ParseContext) error {
 	}
 	return nil
 }
+
+// GetDefaultDeploymentVars shows default deployment var
+func (c *GKE) GetDefaultDeploymentVars(parseContext *kingpin.ParseContext) error {
+	fmt.Print("Key : Value \n")
```
geekodour: We can just call this function info, that way we would be able to add more provider specific info later if we want. In place of `Key: Value`, just writing `DeploymentVars` would do imo. but feel free to keep it this way if you feel strongly about this :)

prombench/.gitignore
```diff
@@ -1 +1,2 @@
 /prombench
+.env
```
geekodour: This should go inside `.git/info/exclude` this way you exclude env files that are specific to your dev environment.

geekodour (commented):
Nice work! :+1: Thanks!

Some suggestions and improvements.

prombench/.gitignore
imrajdas: Removed from .gitignore

pkg/provider/gke/gke.go
imrajdas: DOne

infra/README.md
imrajdas: Done

pkg/provider/provider.go
```diff
@@ -136,3 +137,13 @@ func MergeDeploymentVars(ms ...map[string]string) map[string]string {
 	}
 	return res
 }
+
+// GetDefaultDeploymentVars shows default deployment var
+func (dr *DeploymentResource) GetDefaultDeploymentVars(parseContext *kingpin.ParseContext) error {
```
geekodour:
```suggestion
func (dr *DeploymentResource) GetDeploymentVars(parseContext *kingpin.ParseContext) error {
```

pkg/provider/provider.go
```diff
@@ -136,3 +137,13 @@ func MergeDeploymentVars(ms ...map[string]string) map[string]string {
 	}
 	return res
 }
+
+// GetDefaultDeploymentVars shows default deployment variables.
+func (dr *DeploymentResource) GetDefaultDeploymentVars(parseContext *kingpin.ParseContext) error {
+	fmt.Print("-------------------\n   DeploymentVars   \n------------------- \n")
+	for key, value := range dr.FlagDeploymentVars {
```
geekodour: I don't see we have this method set for `DeploymentResource`, this method made more sense in the provider specific files that you added in your previous commit. It will also have sort of a clean separation between `FlagDeploymentVars`, `DefaultDeploymentVars` and `providerDeploymentVars`

Why do you think  we should have this here? :thinking:

pkg/provider/gke/gke.go
```diff
@@ -43,6 +43,10 @@ import (
 
 // New is the GKE constructor.
 func New(dr *provider.DeploymentResource) *GKE {
+
+	// Set SERVICEACCOUNT_CLIENT_EMAIL to LoadBalancer.
+	dr.FlagDeploymentVars["NGINX_SERVICE_TYPE"] = "LoadBalancer"
```
geekodour: We have a new function to set this now. See `SetupDeploymentResources` from #413

infra/infra.go
```diff
@@ -47,6 +47,10 @@ func main() {
 		Short('a').
 		StringVar(&g.Auth)
 
+	k8sGKEShow := k8sGKE.Command("info", "show GKE related infomation")
+	k8sGKEShow.Command("deploymentvars", "gke info deploymentvars").
+		Action(dr.GetDefaultDeploymentVars)
```
geekodour: We probably should have only one command `infra gke info`, we can call the `GetDefaultDeploymentVars` action for now, but later we can extend it to print out other useful things :)

geekodour (commented):
Some questions and suggestions. :)

pkg/provider/gke/gke.go
```diff
@@ -43,6 +43,10 @@ import (
 
 // New is the GKE constructor.
 func New(dr *provider.DeploymentResource) *GKE {
+
+	// Set NGINX_SERVICE_TYPE to LoadBalancer.
+	dr.DefaultDeploymentVars["NGINX_SERVICE_TYPE"] = "LoadBalancer"
```
geekodour: You should be setting default deployment vars here:

https://github.com/prometheus/test-infra/blob/623686db4f315d316d6ba3bce3c58dc288b14c6c/pkg/provider/provider.go#L45-L51

pkg/provider/provider.go
```diff
@@ -136,3 +137,13 @@ func MergeDeploymentVars(ms ...map[string]string) map[string]string {
 	}
 	return res
 }
+
+// GetDefaultDeploymentVars shows default deployment variables.
+func (dr *DeploymentResource) GetDefaultDeploymentVars(parseContext *kingpin.ParseContext) error {
+	fmt.Print("-------------------\n   DeploymentVars   \n------------------- \n")
+	for key, value := range dr.DefaultDeploymentVars {
```
geekodour: This will only show default deploymentvars, we want to see all the deployment vars that'll be passed when parsing the manifest files. Therefore this function should be inside the specific provider code.(in our case `gke.go`) 

So we want to see `FlagDeploymentVars`+`DefaultDeploymentVars`+if we set any provider specific vars inside `SetupDeploymentResources`

We'd want to get info about:
https://github.com/prometheus/test-infra/blob/623686db4f315d316d6ba3bce3c58dc288b14c6c/pkg/provider/gke/gke.go#L67

https://github.com/prometheus/test-infra/blob/623686db4f315d316d6ba3bce3c58dc288b14c6c/pkg/provider/gke/gke.go#L133

If there are no FlagDeploymentvars or provider specific  deployment it'll simply show the defaultdeploymentvars :)

geekodour (commented):
Some questions and suggestions :) 

I think once this is merged, we don't have any other blockers for local?

pkg/provider/provider.go
imrajdas: Done

infra/infra.go
imrajdas: Done

pkg/provider/gke/gke.go
imrajdas: Done

pkg/provider/provider.go
imrajdas: Done

pkg/provider/provider.go
```diff
@@ -43,10 +43,14 @@ type DeploymentResource struct {
...
 func NewDeploymentResource() *DeploymentResource {
+	// Set NGINX_SERVICE_TYPE to LoadBalancer.
+	DefaultDeploymentVars := map[string]string{}
+	DefaultDeploymentVars["NGINX_SERVICE_TYPE"] = "LoadBalancer"
+
 	return &DeploymentResource{
 		DeploymentFiles:       []string{},
 		FlagDeploymentVars:    map[string]string{},
-		DefaultDeploymentVars: map[string]string{},
+		DefaultDeploymentVars: DefaultDeploymentVars,
```
geekodour: why not just do:
```suggestion
		DefaultDeploymentVars: map[string]string{
		  "NGINX_SERVICE_TYPE" : "LoadBalancer",
		}
```
?

pkg/provider/provider.go
```diff
@@ -43,10 +43,14 @@ type DeploymentResource struct {
 
 // NewDeploymentResource returns DeploymentResource with default values.
 func NewDeploymentResource() *DeploymentResource {
+	// Set NGINX_SERVICE_TYPE to LoadBalancer.
```
geekodour: I think we can get rid of this comment now as this is not telling us anything different from what the map declaration is telling. :)

pkg/provider/gke/gke.go
```diff
@@ -579,3 +579,13 @@ func (c *GKE) ResourceDelete(*kingpin.ParseContext) error {
 	}
 	return nil
 }
+
+// GetDeploymentVars shows deployment variables.
+func (c *GKE) GetDeploymentVars(parseContext *kingpin.ParseContext) error {
+	fmt.Print("-------------------\n   DeploymentVars   \n------------------- \n")
+	for key, value := range c.DeploymentVars {
+		fmt.Print(key, " : ", value, "\n")
```
geekodour:
```suggestion
		fmt.Println(key, " : ", value)
```

geekodour (commented):
Everything looks good :partying_face:  just small nits. 

Will merge after changes are resolved :)

pkg/provider/provider.go
imrajdas: Okay

pkg/provider/provider.go
imrajdas: This is also possible. Thanks

PR #412, opened 2020-07-18
Removing unused service-account secret from 1b_secret

imrajdas:
- To support the PR https://github.com/prometheus/test-infra/pull/390, we need to remove this unused service-account secret from 1b_secret

- [x] Tested in GKE https://gist.github.com/rajdas98/c811dd665398bbf7ca00dabfd19267e8

### DIFF
--- a/prombench/Makefile
+++ b/prombench/Makefile
@@ -1,9 +1,5 @@
 INFRA_CMD        ?= ../infra/infra
 
-ifeq ($(AUTH_FILE),)
-AUTH_FILE = /etc/serviceaccount/service-account.json
-endif
-
 .PHONY: deploy clean
 deploy: nodepool_create resource_apply
 # GCP sometimes takes longer than 30 tries when trying to delete nodepools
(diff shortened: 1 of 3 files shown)


imrajdas:
cc @geekodour

geekodour:
Could we also remove this?

https://github.com/prometheus/test-infra/blob/2c519dd4c7012bf84c6a2cdb590fafbd3e6771a0/prombench/Makefile#L4

imrajdas:
AUTH_FILE is used by other make steps in Makefile. If we remove this other commands will not work

geekodour:
We don't need the default value anymore, since we're providing it in our command runs.

It was used for this:
https://github.com/prometheus/test-infra/blob/f097e236c498833ffce85bef0042cb8041da1426/components/prow/manifests/prow_internals_2.yaml#L231

It should work, can you check?

imrajdas:
> We don't need the default value anymore, since we're providing it in our command runs.
> 
> It was used for this:
> https://github.com/prometheus/test-infra/blob/f097e236c498833ffce85bef0042cb8041da1426/components/prow/manifests/prow_internals_2.yaml#L231
> 
> It should work, can you check?

Okay, Will remove it.

geekodour:
could you remove this line from the readme aswell? :) as it's no longer used.
```
-v GKE_AUTH="$(cat $AUTH_FILE | base64 -w 0)" \
```

imrajdas:
> could you remove this line from the readme aswell? :) as it's no longer used.
> 
> ```
> -v GKE_AUTH="$(cat $AUTH_FILE | base64 -w 0)" \
> ```

Removed

geekodour:
@rajdas98 please get rid of this line aswell :) 

```
Note: Use -v GKE_AUTH="$(echo $AUTH_FILE | base64 -w 0)" if you're passing the data directly into $AUTH_FILE
```

PR #390, opened 2020-06-24
Adding KIND provider for the prombench

imrajdas:
Proposed Changes
- Adding kind provider to deploy KIND and prombench.
- Adding new commands to infra CLI to do prombench operations using KIND
- Adding a doc to use prombench with KIND

Issue: https://github.com/prometheus/test-infra/issues/333

PR #384, opened 2020-06-17
EKS support for prombench

weastel:
This PR incorporates the following changes:-
1. Add Support for rangeVars
2. EKS commands for creation/deletion of clusters, creation/deletion of nodegroups and resource apply/delete
3. Temporary documentations for EKS

PR #413, opened 2020-07-19
Refactor Infra, pkg/provider and GKE provider

geekodour:
Since we're on our way to add KIND support (#390) and EKS support (#384) for different test-infra components, it makes sense to allow the `DeploymentVars` to be variable among different providers aswell.

This PR adds some minor refactoring and adds the support for `DeploymentVars` overriding and adds `DefaultDeploymentVars` aswell. So that the `DefaultDeploymentVars` can be overridden by `-v`. This can also be easily extended to add support for provider specific `DeploymentVars` (#406)

Here are the things that this PR adds:
- `gke.NewGKEClient` be called lazily inside `infra`
-  Remove `GKE.setProjectID` and add `GKE.checkDeploymentVarsAndFiles` to validate input `DeploymentVars` and `DeploymentFiles`
- Add support for default DeploymentVars by adding `provider.DeploymentResource`
- Add `provider.MergeDeploymentVars` and tests which is basically the overriding idea
- Separate `DeploymentVars` into `FlagDeploymentVars` and `DefaultDeploymentVars`

Todo:
- [x] E2E test of this. (I did partial test, working as expected)

cc : @rajdas98 @krasi-georgiev @nevill   @weastel

### DIFF
--- a/infra/README.md
+++ b/infra/README.md
@@ -17,7 +17,13 @@ usage: infra [<flags>] <command> [<args> ...]
 The prometheus/test-infra deployment tool
 
 Flags:
-  -h, --help  Show context-sensitive help (also try --help-long and --help-man).
+  -h, --help           Show context-sensitive help (also try --help-long and
+                       --help-man).
+  -f, --file=FILE ...  yaml file or folder that describes the parameters for the
+                       object that will be deployed.
+  -v, --vars=VARS ...  When provided it will substitute the token holders in the
+                       yaml file. Follows the standard golang template formating
+                       - {{ .hashStable }}.
 
 Commands:
   help [<command>...]
--- a/infra/infra.go
+++ b/infra/infra.go
@@ -40,7 +40,8 @@ func main() {
 		StringMapVar(&dr.FlagDeploymentVars)
 
 	g := gke.New(dr)
-	k8sGKE := app.Command("gke", `Google container engine provider - https://cloud.google.com/kubernetes-engine/`)
+	k8sGKE := app.Command("gke", `Google container engine provider - https://cloud.google.com/kubernetes-engine/`).
+		Action(g.SetupGKEDeploymentResources)
 	k8sGKE.Flag("auth", "json authentication for the project. Accepts a filepath or an env variable that inlcudes tha json data. If not set the tool will use the GOOGLE_APPLICATION_CREDENTIALS env variable (export GOOGLE_APPLICATION_CREDENTIALS=service-account.json). https://cloud.google.com/iam/docs/creating-managing-service-account-keys.").
 		PlaceHolder("service-account.json").
 		Short('a').
--- a/infra/infra.go
+++ b/infra/infra.go
@@ -20,6 +20,7 @@ import (
 	"path/filepath"
 
 	"github.com/pkg/errors"
+	"github.com/prometheus/test-infra/pkg/provider"
 	"github.com/prometheus/test-infra/pkg/provider/gke"
 	"gopkg.in/alecthomas/kingpin.v2"
 )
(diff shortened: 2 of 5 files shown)


pkg/provider/gke/gke.go
```diff
@@ -43,9 +42,15 @@ import (
 )
 
 // New is the GKE constructor.
-func New() *GKE {
+func New(dr provider.DeploymentResource) *GKE {
```
geekodour: There's a bug here, `dr` get passed when it's corresponding flags not even parsed.

pkg/provider/gke/gke.go
```diff
@@ -124,13 +125,26 @@ func (c *GKE) NewGKEClient(*kingpin.ParseContext) error {
 	}
 	c.clientGKE = cl
 	c.ctx = context.Background()
+
+	return nil
+}
+
+// SetupGKEDeploymentResources Sets up DeploymentVars and DeploymentFiles
+func (c *GKE) SetupGKEDeploymentResources(*kingpin.ParseContext) error {
```
geekodour:
```suggestion
func (c *GKE) SetupDeploymentResources(*kingpin.ParseContext) error {
```

infra/infra.go
```diff
@@ -20,33 +20,36 @@ import (
...
-		Required().
-		Short('f').
-		ExistingFilesOrDirsVar(&g.DeploymentFiles)
-	k8sGKE.Flag("vars", "When provided it will substitute the token holders in the yaml file. Follows the standard golang template formating - {{ .hashStable }}.").
-		Short('v').
-		StringMapVar(&g.DeploymentVars)
 
 	// Cluster operations.
 	k8sGKECluster := k8sGKE.Command("cluster", "manage GKE clusters").
+		Action(g.NewGKEClient).
```
krasi-georgiev, line 52: why not put it after the `SetupGKEDeploymentResources` to avoid needing to repeat it for `k8sGKE.Command("nodepool",` as well?

pkg/provider/gke/gke.go
krasi-georgiev: yep good idea.

pkg/provider/gke/gke.go
```diff
@@ -178,19 +194,19 @@ func (c *GKE) K8SDeploymentsParse(*kingpin.ParseContext) error {
...
-		}
-		v, ok := d["project_id"].(string)
-		if !ok {
-			log.Fatal("Couldn't get project id from the auth file")
+// checkDeploymentVarsAndFiles checks whether the requied deployment vars are passed.
+func (c *GKE) checkDeploymentVarsAndFiles() error {
+	reqDepVars := []string{"PROJECT_ID", "ZONE", "CLUSTER_NAME"}
+	for _, k := range reqDepVars {
+		v := c.DeploymentVars[k]
+		if v == "" {
```
```suggestion
		if _,ok == c.DeploymentVars[k]; !ok {
```

infra/README.md
```diff
@@ -17,7 +17,13 @@ usage: infra [<flags>] <command> [<args> ...]
 The prometheus/test-infra deployment tool
 
 Flags:
-  -h, --help  Show context-sensitive help (also try --help-long and --help-man).
+  -h, --help           Show context-sensitive help (also try --help-long and
+                       --help-man).
+  -f, --file=FILE ...  yaml file or folder that describes the parameters for the
+                       object that will be deployed.
+  -v, --vars=VARS ...  When provided it will substitute the token holders in the
```
krasi-georgiev, line 24: is this readme auto generated as well?

krasi-georgiev (commented):
LGTM with some small nits

infra/infra.go
geekodour: @rajdas98 is planning to add an `info` command (provider specific #406 ) which will list all the provider specific deployment vars. If we add `g.NewGKEClient` with `SetupGKEDeploymentResources`, then the `info` command will require passing the `-a` flag. To avoid that, just calling `g.NewGKEClient` in the commands where it is actually required makes this possible. 

But yes, it is some repetition, but couldn't find a way around :(

infra/README.md
geekodour: Yes this is auto generated

infra/infra.go
krasi-georgiev: ok no problem

geekodour:
~Will test this once, will merge if~ tested. everything seems alright.

PR #422, opened 2020-08-02
Remove verifyUserDisabled and add verify_user to commentMonitor config

geekodour:
Added to allow funcbench to be run by any GitHub user.


tools/commentMonitor/client.go
```diff
@@ -31,6 +31,7 @@ type commentMonitorClient struct {
 	events             []webhookEvent
 	prefixes           []commandPrefix
 	prefixHelpTemplate string
+	prefixVerifyUser   bool
```
krasi-georgiev: Why are these called "prefixHelpTemplate" and "prefixVerifyUser", why not just helpTemplate and verifyUser?

PR #428, opened 2020-08-21
Adding descriptions to prombench dashboard

imrajdas:
Google doc for the panel descriptions: https://docs.google.com/document/d/1FKLUkCcaTgC72Dh-Kz66228PykCwG4zy2Tw7zD_491k/edit

- Adding descriptions to prombench dashboard
Fixes- https://github.com/prometheus/test-infra/issues/328  https://github.com/prometheus/test-infra/issues/305


prombench/manifests/cluster-infra/grafana_dashboard_dashboards_noparse.yaml
```diff
@@ -2349,6 +2350,7 @@ data:
           "dashLength": 10,
           "dashes": false,
           "datasource": "prometheus-meta",
+          "description": "This dashboard indicates how many time series each instance holds",
```
geekodour:
```suggestion
          "description": "This graph indicates how many time series each instance holds",
```

This is a little opinionated but maybe we can completely omit the words "This dashboard"/"This graph" because one will be able to see the description by hovering over the `i` tag on that particular grafana panel only. In some cases maybe having this will make sense :) 

Same for the other cases.

prombench/manifests/cluster-infra/grafana_dashboard_dashboards_noparse.yaml
```diff
@@ -2896,6 +2903,7 @@ data:
           "dashLength": 10,
           "dashes": false,
           "datasource": "prometheus-meta",
+          "description": "This dashboard shows two metrics\n- prometheus_tsdb_head_min_time: Minimum timestamp of the head block\n- prometheus_tsdb_head_max_time: Maximum timestamp of the head block\n",
```
geekodour: I don't think it's the best idea to simply have the metric names in the description. :) One could look that up in the graph editor anyway.

A more friendly description (based on your current description) would be something like "_Plots of minimum and maximum timestamp of the head block in reference to current time_", you could go on further to explain what the timestamp of the head block may signify.

prombench/manifests/cluster-infra/grafana_dashboard_dashboards_noparse.yaml
```diff
@@ -3381,6 +3393,7 @@ data:
           "dashLength": 10,
           "dashes": false,
           "datasource": "prometheus-meta",
+          "description": "This dashboard shows the following metrics.\n- Total WAL corruptions\n- Total TSDB reload failures\n- Head series not found\n- Total Compactions failures\n- Total retentions cutoffs failures",
```
geekodour: The problems can already be seen in the graph legends, so we might just want to keep it short something like "_Various tsdb related checks_", or in this case, since the title is pretty much self explanatory, we may not even need a description here.

Or the other idea could be to mention cause and effects of these tsdb issues in the description.

Interested in hearing what others have to say about this. :)

geekodour (commented):
nice :) Could you please link the google doc in the PR description which has the Panel title along with the description in tabular format? It seems that GitHub diff cuts the panel title and it becomes very hard to see where this description actually belongs to. 

I've added some general dashboard related comments.

imrajdas:
Here is the document- https://docs.google.com/document/d/1FKLUkCcaTgC72Dh-Kz66228PykCwG4zy2Tw7zD_491k/edit

bboreham:
@geekodour do you think we could merge this?

bboreham (approved):
Approving on the basis (a) previous comments seem to have been addressed and (b) the added text seems reasonable to me.

Thanks!

PR #432, opened 2020-08-27
WIP: Adding dashboard directory for grafana dashboards and script to combine the dashboards

### LINKED ISSUE #420: Autogenerate grafana_dashboard_dashboards_noparse.yaml
Currently it's somewhat of a pain to add modifications to dashboards because these are JSON inside yaml files, formatting issues etc etc.

It'll be nice if we can have some `dashboard` directory somewhere where we store dashboards in json. and we can run some script to auto generate grafana_dashboard_dashboards_noparse.yaml from them. then we can apply it as we're currently applying it.

related #364 

https://github.com/prometheus/test-infra/blob/master/prombench/manifests/cluster-infra/grafana_dashboard_dashboards_noparse.yaml

imrajdas:
- fixes: https://github.com/prometheus/test-infra/issues/420
- Changes:
   - Moving grafana-dashboard config to dashboards directory with separate json
   - Adding apply/delete for grafana-dashboard config
   - Combing both dashboards JSON


imrajdas:
cc @geekodour. dashboard's configmap is successfully created in this way. I check the text diff with the previous config map. It looks the same(except the JSON format changes).

But after applying all resources, the dashboards are not showing in the grafana. I will fix this problem later. Let me know if this approach is good or not.

pkg/provider/provider.go
```diff
@@ -138,3 +144,75 @@ func MergeDeploymentVars(ms ...map[string]string) map[string]string {
 	}
 	return res
 }
+
+func getK8sClientSet() (*kubernetes.Clientset, error) {
```
geekodour, line 148: We already have the clientset at `pkg/provider/k8s/k8s.go` :)

geekodour (commented):
Thanks for the PR! 

It'll be good to keep the provider code limited to auth provider and k8s code. Having to do with anything dashboards and grafana should not be in the scope of the provider code.

I think this change is also hardcoding filenames and names of the configmaps which has quite a few issues on itself :( It would be better if we maintain the flexibility to choose filenames and other parameters.

In #420 I simply meant  generating `grafana_dashboard_dashboards_noparse.yaml` file from grafana dashboard exports(json files), Ideally would be done by script or in some generic way to deal with generating one configmap from multiple normal files. Nothing to do with the provider.

That way rest of our code will remain the same and nothing changes :) What I think this PR is trying to do is specifically apply grafana dashboard files, the `infra` cli is less concered about what the content of the `ConfigMap` is (in our case grafana dashboard files) :)

imrajdas:
Thanks @geekodour for reviewing this PR. I will come up with a better solution using bash script.

PR #440, opened 2020-09-30
Update path to infra executable in Grafana dashboard automation

geekodour:
Signed-off-by: Hrishikesh Barman <hrishikeshbman@gmail.com>

PR #442, opened 2020-09-30
Update path to executable in $PATH for infra in GH Grafana workflow

geekodour:
Fix added in #440 was incorrect (used local path instead of path to executable in dockeimage)


roidelapluie:
:crossed_fingers:

PR #467, opened 2022-08-07
Update build

SuperQ:
* Update to Go 1.18.
* Update Go modules to 1.17 format.
* Bump some Go dependencies.
* Add dependabot.
* Sync some files from upstream Prometheus.

Signed-off-by: SuperQ <superq@gmail.com>


SuperQ:
For the gke proto use, there's no specified replacement for the deprecated API. It just says "Do not use". :frowning_face: So I added an ignore for now.

Nexucis (approved):
A bit sad for "the do not use" without any pointer to replace it :/

PR #487, opened 2022-11-18
Synchronize common files from prometheus/prometheus

prombot:
Propagating changes from prometheus/prometheus default branch.


metalmatze:
No idea why golangci-lint is failing :thinking:

SuperQ:
@metalmatze Fixed it, extended the timeout to match the config in prometheus/prometheus.

PR #525, opened 2023-03-20
Update load-generator requirements

SuperQ:
Update all requirements to the latest to fix conflicts.
* Enable dependabot for Python requirements.
* Remove obsolete requirements.


SuperQ:
```
ERROR: Cannot install -r /usr/src/app/requirements.txt (line 12) and urllib3==1.26.5 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested urllib3==1.26.5
    requests 2.20.1 depends on urllib3<1.25 and >=1.21.1
```

geekodour:
Sorry for the late response @SuperQ , haven't been keeping up with the project.

- Just took a look at this, seems like this specifies some old dependencies that are not even used.
- Moreover, think it'll be better to specify a more exact version on the Dockerfile maybe.
- These changes can be added later, just made a note here. 
- Thanks for the Dependabot change!

Otherwise changes LGTM. Please merge.


`requirements.txt` with only things used in `main.py`

```
PyYAML==6.0
certifi==2022.12.7
charset-normalizer==3.1.0
idna==3.4
prometheus-client==0.16.0
requests==2.28.2
urllib3==1.26.15
```

```diff
1d0
< DateTime==5.1
4,6c3
< cffi==1.15.1
< chardet==5.1.0
< cryptography==39.0.2
---
> charset-normalizer==3.1.0
9,11d5
< pyOpenSSL==23.0.0
< pycparser==2.21
< pytz==2022.7.1
13d6
< six==1.16.0
15d7
< zope.interface==6.0
```

SuperQ:
I think using requirements.txt is better than in Dockerfile, as you may want to build locally.

SuperQ:
Ok, cleaned up the requirements.txt.

geekodour:
> I think using requirements.txt is better than in Dockerfile, as you may want to build locally.

Oh sorry, I meant the python version. I tested this locally [with `3.10.9`](https://hub.docker.com/layers/library/python/3.10.9/images/sha256-cfa3b79333c4e56fc675b6800445b6dcbb3e6cd4d52f2a9ade944ab73dadc6a1?context=explore) but `FROM python:3` will just get the most recent version which should be fine too.

PR #533, opened 2023-04-28
moved loader=yaml.FullLoader to correct place

Zanabazer:
Loader type is an argument for yaml.load function not the open function

### DIFF
--- a/tools/load-generator/main.py
+++ b/tools/load-generator/main.py
@@ -118,7 +118,7 @@ def main():
     namespace = sys.argv[1]
     pr_number = sys.argv[2]
 
-    config = yaml.load(open("/etc/loadgen/config.yaml", 'r', Loader=yaml.FullLoader).read())
+    config = yaml.load(open("/etc/loadgen/config.yaml", 'r').read(), Loader=yaml.FullLoader)
 
     print("loaded configuration")
 


bboreham:
Do you have some way to try this before merge?

Zanabazer:
I'm not familiar enough with how the prombench seems to load the Docker containers with files. From what I can tell it seems to pull them from the repo when loading the Docker containers so it doesn't take the local file when you edit main.py or even the Dockerfile in tools/load-generator/Dockerfile. The only way I could think to test this was to get it into the main repo. If there is a different flow or I am mistaken on how prombench seems to be populating the docker containers with the files I could try it but I've tried manually editing both the file and running as well changing the Dockerfile to completely use a different file and neither of those changes seemed to propagate to the containers.

From grepping through the repo that is the file that needs to be changed and that function was updated. Following the pyyaml wiki https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation it shows the syntax of the function call to be

yaml.load(input, Loader=yaml.FullLoader)

I didn't put the correct syntax in the first pull request.

bboreham (approved):

bboreham:
OK, I don't see how it can be any worse; I'll merge it and see if prombench works any better.

bboreham:
Queries now working in prombench, thanks!

PR #595, opened 2023-08-25
prombench: Added parca for continous profiling support.

### LINKED ISSUE #552: Add Continuous Profiling to Prombench for meta observability
I would be useful to continuously profile benchmarked Prometheus-es to understand where the bottlenecks are (CPU, mem)

Julien: "I would also like to have profiling with Parca as part of Prombench."

bwplotka:
Fixes https://github.com/prometheus/test-infra/issues/552

### DIFF
--- a/prombench/manifests/cluster-infra/8b_parca.yaml
+++ b/prombench/manifests/cluster-infra/8b_parca.yaml
@@ -84,13 +84,6 @@ spec:
           - "--cors-allowed-origins=*"
           - "--debuginfod-upstream-servers=https://debuginfod.systemtap.org"
           - "--debuginfod-http-request-timeout=5m"
-        livenessProbe:
-          exec:
-            command:
-              - /grpc_health_probe
-              - -v
-              - -addr=:7070
-          initialDelaySeconds: 5
         name: parca
         terminationMessagePolicy: FallbackToLogsOnError
         volumeMounts:
--- a/prombench/manifests/cluster-infra/8b_parca.yaml
+++ b/prombench/manifests/cluster-infra/8b_parca.yaml
@@ -1,0 +1,147 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: parca
+data:
+  # TODO(bwplotka): Add config reloader, otherwise pod has to manually deleted.
+  parca.yaml: |-
+    "object_storage":
+      "bucket":
+        "config":
+          "directory": "/var/lib/parca"
+        "type": "FILESYSTEM"
+    scrape_configs:
+      - job_name: prometheus
+        scrape_interval: 30s
+        scrape_timeout: 40s
+        scheme: http
+        
+        kubernetes_sd_configs:
+          - role: endpoints
+              
+        tls_config:
+          ca_file: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+          insecure_skip_verify: true
+        bearer_token_file: /var/run/secrets/kubernetes.io/serviceaccount/token
+            
+        relabel_configs:
+          - action: keep
+            source_labels: [__meta_kubernetes_service_label_app]
+            regex: prometheus|prometheus-meta
+          - source_labels: [__meta_kubernetes_service_label_prometheus]
+            target_label: prometheus
+          # Each prometheus is configured with external path option, so it serves on certain hardcoded path. Adjust it.
+          - source_labels: [__meta_kubernetes_namespace]
+            regex: prombench-(.*)
+            target_label: pr_number
+          - source_labels: [__profile_path__]
+            target_label: __init_profile_path
+          - source_labels: [__meta_kubernetes_service_label_app, __init_profile_path]
+            regex: prometheus-meta;(.*)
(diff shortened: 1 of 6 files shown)


bwplotka:
~Almost there~ Done

* [X] Prometheus listens for profiles on dynamic path (due to web prefix being applied to pprof paths too - might require Prometheus code change)
* [X] Update comment tool to give exact path for interesting profiles

bwplotka:
image

prombench/manifests/cluster-infra/8b_parca.yaml
```diff
@@ -0,0 +1,147 @@
...
+          - "--debuginfod-upstream-servers=https://debuginfod.systemtap.org"
+          - "--debuginfod-http-request-timeout=5m"
+        livenessProbe:
+          exec:
+            command:
+              - /grpc_health_probe
+              - -v
+              - -addr=:7070
+          initialDelaySeconds: 5
+        name: parca
```
bboreham, line 87: Should this container have CPU and memory requests?
Also probably ephemeral-storage.

prombench/manifests/cluster-infra/8b_parca.yaml
```diff
@@ -0,0 +1,147 @@
...
+        args:
+          - /parca
+          - "--http-address=:7070"
+          - "--config-path=/etc/parca/parca.yaml"
+          - "--path-prefix=/profiles"
+          - "--log-level=info"
+          - "--cors-allowed-origins=*"
+          - "--debuginfod-upstream-servers=https://debuginfod.systemtap.org"
+          - "--debuginfod-http-request-timeout=5m"
+        livenessProbe:
```
bboreham: You are absolutely certain that killing the process is the best thing to do?

bboreham (commented):
Thanks for this.  I had a couple of questions.

prombench/manifests/cluster-infra/8b_parca.yaml
bwplotka: Probably yes, but none of containers really have here, plus we have dedicated nodes, so no need for K8s scheduling really. Let's do it once it will be necessary, WDYT?

prombench/manifests/cluster-infra/8b_parca.yaml
bwplotka: What do you mean?

bboreham: That's what livenessProbe does.  I agree that in some situations this is appropriate, but I want you to be sure.
https://twitter.com/bboreham/status/1178083887991398400

prombench/manifests/cluster-infra/8b_parca.yaml
```diff
@@ -0,0 +1,147 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: parca
+data:
+  # TODO(bwplotka): Add config reloader, otherwise pod has to manually deleted.
```
nicolastakashi, line 6: Worth maybe try to mimic helm feature of add a annotation if the checksum of the file?
Something like `sha256sum configmap.yaml | cut -d ' ' -f 1` might get the checksum and we just need to sed the deployment with an annotation `checksum/config`

prombench/manifests/cluster-infra/8b_parca.yaml
bboreham: I guess it's ok to fix this in a separate PR.
It is necessary - kubelet may decide to evict pods that are over their request, and CPU requests feed into the Linux scheduler.

prombench/manifests/cluster-infra/8b_parca.yaml
bwplotka: Nice! The other option is HUP signal or `/reload`. I think Parca supports this (hot reload). Also probably they would be fine with native way of doing this as proposed in https://github.com/prometheus/prometheus/issues/9783 at some point. Let's leave it for another PR (:

prombench/manifests/cluster-infra/8b_parca.yaml
bwplotka: Right, I took it from their recommended YAML, but happy to change. I think there is no logical reason for it (no known Parca state of being stuck that would require process kill), but who knows. Will remove it for now.

prombench/manifests/cluster-infra/8b_parca.yaml
```diff
@@ -0,0 +1,140 @@
...
+        - mountPath: /var/lib/parca
+          name: storage
+        ports:
+        - name: parca-web
+          containerPort: 7070
+      volumes:
+      - name: config
+        configMap:
+          name: parca
+      # TODO(bwplotka): Make it persistent at some point.
```
metalmatze, line 101: Do other components of the prom test infra use GCS? That might be even nicer.

bwplotka: No, I don't think so. Let's make it in another PR if needed.

metalmatze (approved):

PR #609, opened 2023-10-29
Update test-infra go modules

jesusvazquez:
Bumping go modules


SuperQ:
The monthly Dependabot run should take care of these.

jesusvazquez:
@SuperQ yes, this is related to a thread in prometheus-dev were we're investigating why infra can no longer authenticate against gcp https://cloud-native.slack.com/archives/C01AUBA4PFE/p1698501889156349

We need to use the new gcp auth plugin or refactor the code so the k8s provider is given already configured credentials. 

I was thinking it would be better if we did that refactor with the dependencies up to date thus the attempt at updating them beforehand.

SuperQ (approved):

bboreham (approved):

PR #638, opened 2024-01-15
Parca: update to 0.20.0

bboreham:
I saw some faults with the UI, and apparently it has been rewritten between 0.18 and 0.20.

Didn't see any breaking changes listed, but I have no idea whether it will work.


bwplotka (approved):
Let's merge and wait for somebody (us) to deploy & test.

Nothing will be deployed automatically after this merge.

PR #660, opened 2024-04-01
Grafana: Fix `pr-number` variable

aknuds1:
Rename `pr-number` Grafana dashboard variable to `prNumber`, since dashes are not allowed in Grafana variable names. As a result, the variable wasn't working. Also, switch to newer variable expansion syntax (`${prNumber}` instead of `[[prNumber]]`, as the latter is [deprecated](https://grafana.com/docs/grafana/latest/dashboards/variables/variable-syntax/#variable-syntax)).

The variable's JSON changed a bit apart from the name, since I regenerated it within Grafana.

NB: I've tested locally, not in GitHub Actions. Not sure why I only discovered this problem locally. When deploying prombench locally, the Grafana version is 9.4.3.

### DIFF
--- a/prombench/manifests/cluster-infra/grafana_dashboard_dashboards_noparse.yaml
+++ b/prombench/manifests/cluster-infra/grafana_dashboard_dashboards_noparse.yaml
@@ -1573,7 +1573,7 @@ data:
               "steppedLine": false,
               "targets": [
               {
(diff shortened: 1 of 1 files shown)


jesusvazquez (approved):
LGTM nice find Arve 🙏

PR #714, opened 2024-08-03
Update README.md

Sweetdevil144:
The `go install... ` command in Readme was slightly incorrect I've updated the command to install the package correctly while pulling the latest version.


jesusvazquez (approved):
LGTM thanks for the addition!

PR #744, opened 2024-09-05
written a go code for generating data in open_metric formate. 

kushalShukla-web:
Here in this PR, I have written Go code to generate data in the OpenMetrics format. Once this PR is approved by the maintainers, I will proceed to create block data using `promtool`.


bboreham (commented):
Thanks!  It's fine as a starting-point; let's carry on a bit with the idea because I think some changes will be needed before we're done.

* Prometheus code has a copyright notice at the top of every file.
* Add one or two sentences what the function is all about.

PR #746, opened 2024-09-15
Generated Data in Block formate using promtool for now

kushalShukla-web:
This pull request aims to confirm whether I am proceeding on the right Direction Specifically, it addresses the following: 

- [x] Generation of data for PromBench.
- [ ] Configuration of the Loadgen Querier to utilize the newly generated data.
@bboreham @kakkoyun


kushalShukla-web:
Update 2.0.0 
 1.   In this PR, I have commented out the prometheus.yml file as it is not required at the moment. 
 2.   I have used hostPath volumes to retrieve data from the local system. Hi @bboreham, can we  use Persistent Volumes instead of hostPath volumes ? but as the prometheus pods are running on a separate nodes so if pods terminated still they are going to be created on the same node . 
 3.  Currently, the path is hardcoded for my system. I need to make it more dynamic (I tried but did not came up with the answer) or use Persistent Volumes instead.
 4. Hi @bboreham  i need little bit your guidance here .

bboreham:
Thanks for this. 

I don't think we want large amounts of binary data in the git repo.  We can set up a Google Cloud block volume and put the data there, or maybe a Google Cloud Storage bucket.

Your `hostPath` volume is fine to illustrate; depending on the point above we will mount the Google Cloud volume or not.

I don't think we want large amounts of config commented out either.

kakkoyun:
This needs a rebase @kushalShukla-web

kushalShukla-web:
I think @bboreham @kakkoyun  we can close this pull request.

kakkoyun:
> I think @bboreham @kakkoyun we can close this pull request.

@kushalShukla-web Let's add a reasoning about why we closed this. It could be helpful for future readers.

kushalShukla-web:
Because we don't want to store TSDB data in the test-infra repository, we are not proceeding with this approach.

Issue #705, opened 2024-07-23
PromBench Idea: exercise more parts of PromQL


bboreham:
PromBench runs synthetic load defined in [6_loadgen.yaml](https://github.com/prometheus/test-infra/blob/34a189da3778d9035d430e3213f5c9dd18d520a7/prombench/manifests/prombench/benchmark/6_loadgen.yaml).
These cover selectors, `rate`, `histogram_quantile` and `sum`, but leave out most of PromQL.

For instance we could add:
* binary arithmetic operators (joins)
* logical operators (`and`, `or`, `unless`)
* `topk`

HarshitNagpal29:
@bboreham PTAL https://github.com/prometheus/test-infra/pull/723

PR #747, opened 2024-09-18
Add additional PromQL operators to synthetic load

kushalShukla-web:
This PR enhances the synthetic load generation by incorporating additional PromQL operators in the 6_loadgen.yaml file:

- Added binary arithmetic operators (joins) to cover more complex query types.
- Included logical operators (and, or, unless) for better query testing coverage.
- Added `topk` function to test query performance with ranked results.


kushalShukla-web:
Solves #705

prombench/manifests/prombench/benchmark/6_loadgen.yaml
```diff
@@ -48,6 +48,28 @@ data:
         - expr: histogram_quantile(0.99, sum by(path, le) (rate(codelab_api_request_duration_seconds_bucket{method="POST"}[5m])))
         - expr: histogram_quantile(0.99, sum by(path, method, le) (rate(codelab_api_request_duration_seconds_bucket{method="POST"}[5m])))
         - expr: histogram_quantile(0.99, sum by(instance, le) (rate(codelab_api_request_duration_seconds_bucket{method="POST"}[5m])))
+      - name: arthematic_operation
+        interval: 10s
+        type: instant
+        queries:
+        - expr: sum(node_cpu_seconds_total)/sum(container_memory_rss)
```
bboreham: This is not terribly good as a test of operator performance, since it matches one series against one other series, both of which have no labels.

prombench/manifests/prombench/benchmark/6_loadgen.yaml
```diff
@@ -48,6 +48,28 @@ data:
...
+        - expr: rate(node_cpu_seconds_total[5m]) * 5
```
bboreham: This is better, at 256 series when tested.

prombench/manifests/prombench/benchmark/6_loadgen.yaml
```diff
@@ -48,6 +48,28 @@ data:
...
+        - expr: sum(go_gc_heap_goal_bytes)/sum(loadgen_query_duration_seconds_created)
```
bboreham: This has the same structural problem as the first one.

prombench/manifests/prombench/benchmark/6_loadgen.yaml
```diff
@@ -48,6 +48,28 @@ data:
...
+      - name: logic_operator
+        interval: 10s
+        type: instant
+        queries:
+        - expr: node_cpu_seconds_total{mode="nice"} and node_cpu_seconds_total{namespace="default"}
```
bboreham: Nodes don't have a namespace, so this is short-cutted to return nothing.

bboreham (commented):
Hi, I commented on a few, but they are all likely problemmatic.
I recommend you change tack to use metrics exposed by the fake webserver, since there are a lot of them and prombench can control them.

kushalShukla-web:
> Hi, I commented on a few, but they are all likely problemmatic. I recommend you change tack to use metrics exposed by the fake webserver, since there are a lot of them and prombench can control them.

Actually i have tested this on one of the Pull Request where the prombench was running.

prombench/manifests/prombench/benchmark/6_loadgen.yaml
kushalShukla-web: So Here for improving the performance i have to use some labels something like this ? 
`sum by (instance)(node_cpu_seconds_total) / sum by (instance)(container_memory_rss)`

bboreham:
Ok but my recommendation remains the same.

prombench/manifests/prombench/benchmark/6_loadgen.yaml
kushalShukla-web: I have replaced the query with `sum by (container, pod) (rate(container_cpu_usage_seconds_total[5m])) / sum by (container, pod) (container_spec_cpu_quota)`, which calculates the CPU utilization of containers relative to their allocated CPU quota. The query is structured with the labels container and pod.

prombench/manifests/prombench/benchmark/6_loadgen.yaml
kushalShukla-web: I replaced the query with `sum by(instance) (rate(node_cpu_seconds_total[5m])) / avg by(instance) (node_load1)`, which is also structured with the label instance.

kushalShukla-web:
Hi @bboreham,

I have introduced some new metrics from Prometheus in PromBench:

1.     container_cpu_usage_seconds_total
2.     container_spec_cpu_quota
3.     storage_operation_duration_seconds_sum

Let me know if any further changes are needed in this PR.

bboreham:
> I have replaced some of the existing metrics with new ones from PromBench, such as:

How many of those did you count?

kushalShukla-web:
> > I have replaced some of the existing metrics with new ones from PromBench, such as:
> 
> How many of those did you count?

These are some of the new unique metrics that  i have written 

- container_cpu_usage_seconds_total
- container_spec_cpu_quota
- node_network_receive_bytes_total
- storage_operation_duration_seconds_sum
- storage_operation_duration_seconds_count
- kubelet_runtime_operations_duration_seconds_sum
- kubelet_runtime_operations_duration_seconds_count
- node_filesystem_avail_bytes
- container_ulimits_soft
- container_memory_working_set_bytes
- container_memory_usage_bytes
- container_sockets

kushalShukla-web:
Hi @bboreham all the queries having value above 2000 , and codelabz metric is having 52000 different variations .

prombench/manifests/prombench/benchmark/6_loadgen.yaml
```diff
@@ -48,6 +48,33 @@ data:
...
+        - expr: codelab_api_request_duration_seconds_sum{status="200"} or codelab_api_request_duration_seconds_sum{status="500"}
+        - expr: codelab_api_request_duration_seconds_bucket{status="200"} and codelab_api_request_duration_seconds_bucket{method="GET"}
+        - expr: codelab_api_request_duration_seconds_count{method="POST"} and codelab_api_request_duration_seconds_count{status="500"}
+        - expr: codelab_api_request_duration_seconds_bucket{status="200"} or codelab_api_request_duration_seconds_bucket{method="GET"}
+      - name: topk
+        interval: 10s
+        type: instant
+        queries:
+        - expr: topk(2000, sum(rate(go_gc_duration_seconds_count[5m])) by (instance, job))
+        - expr: topk(10000, sum(codelab_api_request_duration_seconds_bucket) by (method,job))
```
bboreham: `topk(10000` is not realistic; nobody is going to scroll down 10,000 lines of screen output to find something.
`k` should be more like 10, or perhaps 100.
Also I don't think there are 10,000 combinations of method and job.
Also it is not valid to sum histogram buckets.

prombench/manifests/prombench/benchmark/6_loadgen.yaml
```diff
@@ -48,6 +48,33 @@ data:
...
+        - expr: rate(go_gc_duration_seconds_sum{job=~"fake-webservers-[0-9]+"}[5m]) / rate(go_gc_duration_seconds_count{job=~"fake-webservers-[0-9]+"}[5m])        
+        - expr: sum by (instance, job) (rate(codelab_api_request_errors_total[5m])) / sum by (instance, job) (rate(go_memstats_mallocs_total[10m]))
+      - name: logic_operator
+        interval: 10s
+        type: instant
+        queries:
+        - expr: codelab_api_request_duration_seconds_bucket{method="GET"} or codelab_api_request_duration_seconds_bucket{method="POST"}
+        - expr: codelab_api_request_duration_seconds_sum{status="200"} or codelab_api_request_duration_seconds_sum{status="500"}
+        - expr: codelab_api_request_duration_seconds_bucket{status="200"} and codelab_api_request_duration_seconds_bucket{method="GET"}
+        - expr: codelab_api_request_duration_seconds_count{method="POST"} and codelab_api_request_duration_seconds_count{status="500"}
```
bboreham: I don't see much point in doing multiple expressions that are essentially the same.
`or` is different to `and`, but after that you could have an `/`, taking the ratio of errors to all requests for instance.

prombench/manifests/prombench/benchmark/6_loadgen.yaml
```diff
@@ -48,6 +48,33 @@ data:
...
+        - expr: codelab_api_request_duration_seconds_bucket{status="200"} or codelab_api_request_duration_seconds_bucket{method="GET"}
+      - name: topk
+        interval: 10s
+        type: instant
```
bboreham, line 66: Could we have some range queries too?

bboreham (commented):
Thanks, this is getting better.

I am still interested to think about the cardinality you are expecting for each operator in each query.

At the end of the day there should be a balance across different kinds of load, so we can justify that prombench is a realistic test, and also we want the prometheus under load to be able to keep up.

prombench/manifests/prombench/benchmark/6_loadgen.yaml
bboreham: I realized a couple of things since this comment: you have `/` in "arithmetic operation" above, but this `and` is never going to return anything because the labels on each side are different. We want the benchmark queries to make sense.

bboreham:
I trimmed down the newly-added queries a bit:
* Slow down arithmetic_operation and logic_operator.
* Took out a few queries to avoid overloading the server.
* Stop querying `_bucket` series directly; those should be used by `histogram_quantile` or similar.
* Use more realistic `k` parameters to `topk`.

I also trimmed down some pre-existing queries for balance, to retain about the same overall load on the server as before.

bboreham (approved):

PR #743, opened 2024-09-03
Replace load-generator with Go implementation

HarshitNagpal2901:
Replace python load-generator with a Go implementation.
Add additional PromQL tests.
Fixes https://github.com/prometheus/test-infra/issues/688
Fixes https://github.com/prometheus/test-infra/issues/705

PR #723, opened 2024-08-21
added-binary-arithmetic-operators-logical-operators-topk

HarshitNagpal29:
Modified the 6_loadgen.yaml file to include tests for the additional PromQL features mentioned in the issue:

1. Binary Arithmetic Operators (Joins)
2. Logical Operators (and, or, unless)
3. TopK

PR #748, opened 2024-09-20
Replace load-generator with Go implementation

bboreham:
This is #743 with one more commit to avoid crashing on empty duration string - instant queries typically have empty start and end times.

Also use Prometheus `model.ParseDuration` to give more flexibility.

And revert the change to `6_loadgen.yaml`, which is in #723.


SuperQ (approved):
Thanks!

PR #726, opened 2024-08-29
Replace the kubernetes.io/ingress.class into spec.ingressClassName.

kushalShukla-web:
kubernetes.io/ingress.class is deprecated .

PR #770, opened 2024-10-18
Fix up ingressClassName setting

bboreham:
Setting it as an annotation was a misunderstanding in #726.

I have tested this change on `ingress-prometheus` in Prombench on GKE.

### DIFF
--- a/prombench/manifests/cluster-infra/3b_prometheus-meta.yaml
+++ b/prombench/manifests/cluster-infra/3b_prometheus-meta.yaml
@@ -257,9 +257,9 @@ kind: Ingress
 metadata:
   name: ingress-prometheus-meta
   annotations:
-    spec.ingressClassName: "nginx"
     nginx.ingress.kubernetes.io/ssl-redirect: "false"
 spec:
+  ingressClassName: nginx
   rules:
   - http:
       paths:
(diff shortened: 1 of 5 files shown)


prombench/manifests/cluster-infra/3b_prometheus-meta.yaml
```diff
@@ -258,6 +258,7 @@ metadata:
   name: ingress-prometheus-meta
   annotations:
     spec.ingressClassName: "nginx"
```
machine424: I'm no longer using nginx-ingress controller, but I wasn't aware of this annotation.
I think it was meant to be added under `spec:`

SuperQ: Yes, this should be down under `spec:`
```
spec:
  ingressClassName: nginx
```

See: https://kubernetes.io/docs/concepts/services-networking/ingress/#the-ingress-resource

bboreham: Updated PR.

machine424 (approved):

bwplotka (commented):
Thanks!

PR #771, opened 2024-10-20
Update loki to v3.2.1

bboreham:
Config based on https://github.com/grafana/loki/blob/5d78a3a3fd1f/cmd/loki/loki-local-config.yaml

All previously-stored logs data is lost when you do this upgrade; v1.4 is too old to convert from.

Also improve a couple of things in the container spec.

### DIFF
--- a/prombench/manifests/cluster-infra/6b_loki_stateful_set.yaml
+++ b/prombench/manifests/cluster-infra/6b_loki_stateful_set.yaml
@@ -49,8 +49,13 @@ data:
     server:
       http_listen_port: 3100
       grpc_listen_port: 9096
-      log_level: debug
+      log_level: info
       grpc_server_max_concurrent_streams: 1000
+
+    compactor:
+      retention_enabled: true
+    limits_config:
+      retention_period: 2184h
 ---
 apiVersion: v1
 kind: Service
--- a/prombench/manifests/cluster-infra/6b_loki_stateful_set.yaml
+++ b/prombench/manifests/cluster-infra/6b_loki_stateful_set.yaml
@@ -114,11 +114,6 @@ spec:
             - name: http-metrics
               containerPort: 3100
               protocol: TCP
-          livenessProbe:
-            httpGet:
-              path: /ready
-              port: http-metrics
-            initialDelaySeconds: 45
           readinessProbe:
             httpGet:
               path: /ready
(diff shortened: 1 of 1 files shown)


prombench/manifests/cluster-infra/6b_loki_stateful_set.yaml
```diff
@@ -7,39 +7,50 @@ metadata:
...
-    storage_config:
-      boltdb:
-        directory: /data/loki/index
-      filesystem:
-        directory: /data/loki/chunks
-    chunk_store_config:
-      max_look_back_period: 0s
-    table_manager:
-      retention_deletes_enabled: true
-      retention_period: 2184h
```
machine424: I see those were discarded, the new defaults are more suitable?

prombench/manifests/cluster-infra/6b_loki_stateful_set.yaml
```diff
@@ -7,39 +7,50 @@ metadata:
...
+      grpc_listen_port: 9096
+      log_level: debug
```
machine424: I assume it's not a leftover (ok for me to keep it if it's not chatty)

machine424 (commented):
lgtm, just some questions

prombench/manifests/cluster-infra/6b_loki_stateful_set.yaml
aknuds1: Yeah, maybe this makes sense in the [config](https://github.com/grafana/loki/blob/5d78a3a3fd1f/cmd/loki/loki-local-config.yaml) this was copied from?

prombench/manifests/cluster-infra/6b_loki_stateful_set.yaml
aknuds1: Curious about the same as @machine424.

aknuds1 (commented):
Looks like a good idea, but wondering about what @machine424 pointed out.

prombench/manifests/cluster-infra/6b_loki_stateful_set.yaml
bboreham: What I read [here](https://github.com/grafana/loki/blob/baaaa83c78c03c6b9257afddc0854daec928a755/docs/sources/operations/storage/retention.md) is that the new defaults are not more suitable.

However I could not find a simple explanation of how to configure retention.

prombench/manifests/cluster-infra/6b_loki_stateful_set.yaml
bboreham: It is indeed copied from the example.  I am not really qualified to say which setting is best.

bboreham:
I got some advice on how to configure retention.
Also reduced log level to info.

machine424 (approved):
Thanks, let's give this a try.

PR #782, opened 2024-11-11
Enhance Load Generator to Query Pre-Generated Data Blocks

kushalShukla-web:
changed Load generator to query from downloaded blocks. Here i have added "current" and "absolute" time to query data between current data (head_blocks) and absolute data (Downloaded) blocks. 
Things left - i have to configure to add  instant query on downloaded blocks !


kushalShukla-web:
Instant query is still missing; currently, it does not perform instant queries on downloaded blocks.

tools/load-generator/main.go
```diff
@@ -139,27 +174,50 @@ func (q *Querier) run(wg *sync.WaitGroup) {
...
+	if q.qtype == "range" && timeMode == "absolute" {
+		fmt.Println("absolute second time checking !!")
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			log.Fatal(err)
+		}
+		defer resp.Body.Close()
+
+		// Print the body content as a string
+		fmt.Println(string(body), "Hi from body side!!!")
```
kakkoyun: It needs to a bit of clean up.

tools/load-generator/main.go
```diff
@@ -120,16 +132,59 @@ func NewQuerier(groupID int, target, prNumber string, qg QueryGroup) *Querier {
...
+	var bucketConfig BucketConfig
+	err = yaml.Unmarshal(data, &bucketConfig)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing YAML: %w", err)
+	}
+
+	return &bucketConfig, nil
+}
+
+func configstate(v *BucketConfig, err error) *configState {
```
bboreham: It is potentially confusing to have two identifiers which differ only in the case of some letters.

tools/load-generator/main.go
```diff
@@ -120,16 +132,59 @@ func NewQuerier(groupID int, target, prNumber string, qg QueryGroup) *Querier {
...
+
+func (q *Querier) run(wg *sync.WaitGroup, timeBound *configState) {
 	defer wg.Done()
 	fmt.Printf("Running querier %s %s for %s\n", q.target, q.name, q.url)
 	time.Sleep(20 * time.Second)
 
 	for {
 		start := time.Now()
-
+		runBlockMode := "current"
```
bboreham, line 179: I think the idea here is to alternate between current and absolute?
At least, put a comment saying that.

tools/load-generator/main.go
```diff
@@ -120,16 +132,59 @@ func NewQuerier(groupID int, target, prNumber string, qg QueryGroup) *Querier {
...
+}
+
+func configstate(v *BucketConfig, err error) *configState {
+	var absolutetime int64
+	if v != nil {
+		absolutetime = v.MaxTime
+	}
+	return &configState{
+		bucketConfig: v,
+		Err:          err,
```
bboreham: I think the idea is we carry `err` around, and act like this config doesn't exist if `err` is non-nil.

This at least deserves a comment.  Maybe an alternative would work, just having `nil` for this `configState` object, if there would be an error, and checking that?

tools/load-generator/main.go
```diff
@@ -120,16 +132,59 @@ func NewQuerier(groupID int, target, prNumber string, qg QueryGroup) *Querier {
 	}
 }
 
-func (q *Querier) run(wg *sync.WaitGroup) {
+// Function to load `minTime` and `maxTime` from bucket-config.yml
+func loadKeyConfig() (*BucketConfig, error) {
+	filePath := "/config/bucket-config.yml"
```
bboreham: This should probably be a parameter, a CLI flag.

kakkoyun (approved):

kushalShukla-web:
same like  PR https://github.com/prometheus/test-infra/pull/779 .

PR #794, opened 2024-11-27
Prombench: tweak rw-sink resource usage

bboreham:
Since the sink program handles a lot of messages but doesn't do anything with them, it does a lot of garbage-collection.
Tell the Go runtime we would like it to use more memory, which reduces the CPU by about half.

Adjust requests to match.

image

### DIFF
--- a/prombench/manifests/prombench/benchmark/7_prometheus_remote_write.yaml
+++ b/prombench/manifests/prombench/benchmark/7_prometheus_remote_write.yaml
@@ -19,8 +19,14 @@ spec:
         image: quay.io/bwplotka/sink:latest
         resources:
           requests:
-            cpu: "560m"
-            memory: "250Mi"
+            cpu: "2"
+            memory: "512Mi"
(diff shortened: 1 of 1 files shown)


bwplotka (approved):
Wow, amazing thanks!


How much traffic do you send it to it, so it consumes 4 CPU? Just Prometheus we "prombench"? I think we might want to limit concurrency to give more predictable resource usage, sink does not need to that "fast" in responding perhaps?

bboreham:
This is 1M series scraped every 5s, on two Prometheus.  So 400K samples/sec?

I'm not very sure what you mean about limiting concurrency.  We use 8-core nodes, so it won't be more than that.

bwplotka:
I mean we can add limit of the concurrency on sink, so it receives slower, but with stable CPU use. But you are right that a lot of it was through GC. 

I will also play with memory regions (new proposal) to reduce GC. Thanks!

PR #796, opened 2024-11-27
Prombench: slow scraping from 5s to 15s

bboreham:
More realistic for end-user workloads; less expensive in cloud resources.


kakkoyun (approved):
LGTM

PR #797, opened 2024-11-27
Prombench: add Kubernetes resource requests

bboreham:
Best to tell the system what is needed.

Specific levels based on observing at GKE, after changing scraping to 15s per #796.

I left some containers un-set, e.g. the init container that builds Prometheus.


kakkoyun (approved):
LGTM

PR #810, opened 2024-12-14
Added Bucket-config as a configmap in a Block-sync init container.

kushalShukla-web:
1. Following [prometheus/test-infra#779](https://github.com/prometheus/test-infra/pull/779) and [prometheus/prometheus#15183](https://github.com/prometheus/prometheus/pull/15183), the placement of the `bucket-config` file in the Prometheus repository has created some inconsistencies, such as:

   - Different data for different PRs.
   - The `bucket-config` file is required in multiple parts of PromBench, such as the Load Generator to query downloaded block data. This leads to downloading the entire Prometheus repository, which is not an efficient solution.

2. To address these issues, I have moved all bucket configuration into a ConfigMap.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -158,7 +158,7 @@ require (
 	go.opentelemetry.io/otel/trace v1.31.0 // indirect
 	golang.org/x/crypto v0.31.0 // indirect
 	golang.org/x/mod v0.17.0 // indirect
-	golang.org/x/net v0.32.0 // indirect
+	golang.org/x/net v0.33.0 // indirect
 	golang.org/x/sync v0.10.0
 	golang.org/x/sys v0.28.0 // indirect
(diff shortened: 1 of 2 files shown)


prombench/manifests/prombench/benchmark/3b_prometheus-test-pr_deployment.yaml
```diff
@@ -93,6 +104,12 @@ spec:
           path: /mnt/disks/ssd0
       - name: prometheus-executable
         emptyDir: {}
+      - name: bucket-config   # Define the Secret volume
+        secret:
+          secretName: bucket-secret
+      - name: key
```
kakkoyun:
```suggestion
      - name: blocksync-location-key
```
The above is just an example.
Let's be more specific. `key` is too generic.

PR #811, opened 2024-12-14
Updated LoadGenerator to use Configmap to query downloaded data.

kushalShukla-web:
In the previous PR [#782](https://github.com/prometheus/test-infra/pull/782), the Load Generator was using the bucket-config file from the Prometheus repository. Instead, it will now be stored as a ConfigMap inside the test-infra repository.


kakkoyun (approved):

kushalShukla-web:
Hi @bboreham @bwplotka , pr looks Mergeable

prombench/manifests/prombench/benchmark/6_loadgen.yaml
```diff
@@ -138,13 +139,18 @@ spec:
...
         ports:
         - name: loadgen-port
           containerPort: 8080
       volumes:
       - name: config-volume
         configMap:
           name: prometheus-loadgen
+      - name: blocksync-location-key
+        configMap:
+          name: blocksync-config
```
bboreham, line 162: Some extra scheme is needed to ensure this ConfigMap exists at the time this pod is started.

bboreham (changes requested):
I think the basic idea is that someone will follow the instructions in #840 to create the `blocksync-config` ConfigMap.
However, as noted in my review of that PR, this has to be done by hand for each Prombench run, because they are all in a different namespace.

prombench/manifests/prombench/benchmark/6_loadgen.yaml
kushalShukla-web: Yeah, but the copy command will copy the ConfigMap into the PR namespace, so I don't think it's going to be an issue.

PR #821, opened 2025-01-31
Improved sink debugging flow.

bwplotka:
This improves PRW benchmark flow by:

* ~Adding PRW 2.0 traffic which we should care about with the next Prometheus releases. We could drop PRW 1.0 communication given no development on that path, but since ppl still use it, let's keep it.~ I removed it for now as it depends on `metadata-wal-records` feature which is problematic currently: https://github.com/prometheus/prometheus/issues/15911
* Updating sink with the latest issue handling features.
* Creating different configmap for each Prometheus in the same file. This is to ensure we have pass a different PRW header so in the `sink` we will know which issue belongs to what Prometheus (e.g. prometheus from the PR is suddenly not sending help).
* Meta Prom scrapes now sink


ArthurSens:
Would you like to try running a custom benchmark from this branch to make sure it works?

bwplotka:
I didn't plan this, no - wanted to spend my time somewhere else (low risk, low impact, can be fixed in seconds) 🙃 
 
But given you pointed out, let's do this: https://github.com/prometheus/prometheus/pull/15914

bwplotka:
Well glad you did asked for it

> 2025-01-31 13:09:29.996 GMT
time=2025-01-31T13:09:29.996Z level=ERROR source=main.go:1465 msg="Failed to apply configuration" err="invalid remote write configuration, if you are using remote write version 2.0 the `--enable-feature=metadata-wal-records` feature flag must be enabled"

 (:

bwplotka:
Removed PRW2 comms for now as it depends on metadata-wal-records feature which is problematic currently: https://github.com/prometheus/prometheus/issues/15911

bwplotka:
Tested https://github.com/prometheus/prometheus/pull/15914 works now.

ArthurSens (approved):
Thanks! Idea for the future: adding the new sink metrics/logs to a separate dashboard :)

kakkoyun (commented):
LGTM

bboreham:
Why did the filenames change? 
e.g. `3b_prometheus-test-pr_deployment.yaml` -> `3_prometheus-test-pr_deployment.yaml`

PR #833, opened 2025-02-07
Fix (tmp) for prometheus-builder:master causing npm install freeze on prombench runs.

bwplotka:
This custom image is locally build from https://github.com/prometheus/golang-builder/pull/296 that we agree to pause until Prometheus release. However, Prometheus release needs working prombench. This is a tmp solution until we have an official image.

Mitigates https://github.com/prometheus/test-infra/issues/832


Nexucis (approved):
awesome, thank you for doing that @bwplotka !
Hope it will unlock the situation

bwplotka:
Testing https://github.com/prometheus/prometheus/pull/15707#issuecomment-2642472143

PR #840, opened 2025-02-08
Added Bucket-config as a configmap in a Block-sync init container.

kushalShukla-web:
1. Following [prometheus/test-infra#779](https://github.com/prometheus/test-infra/pull/779) and [prometheus/prometheus#15183](https://github.com/prometheus/prometheus/pull/15183), the placement of the `bucket-config` file in the Prometheus repository has created some inconsistencies, such as:

   - Different data for different PRs.
   - The `bucket-config` file is required in multiple parts of PromBench, such as the Load Generator to query downloaded block data. This leads to downloading the entire Prometheus repository, which is not an efficient solution.

2. To address these issues, I have moved all bucket configuration into a ConfigMap.


kakkoyun:
This needs a reabase @kushalShukla-web

kushalShukla-web:
i will rebase this .

kushalShukla-web:
i dont understand why lintin showing error ? yaml file looks fine to me.

kushalShukla-web:
Hi @kakkoyun  and @bboreham  @bwplotka  pr looks mergeable.

prombench/docs/gke.md
```diff
@@ -96,8 +96,71 @@ Run Prombench tests in [Google Kubernetes Engine (GKE)](https://cloud.google.com
...
+
+ - **Option 1: To Download Data from a Specific Directory** 
+
+ Create a ConfigMap with the following structure:
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: blocksync-config
+  namespace: prombench-{{ .PR_NUMBER }}
```
bboreham, line 144: This is one of the places where having a dependency on the PR number is a problem.

prombench/docs/gke.md
```diff
@@ -96,8 +96,71 @@ Run Prombench tests in [Google Kubernetes Engine (GKE)](https://cloud.google.com
...
+data:
+  bucket-config.yml: |
+    path: your-directory-name
+    minTime: block-starting-time
+    maxTime: block-ending-time
+```
+Replace the values of ```directory```,```minTime```, and ```maxTime``` with your desired configuration. Here ```minTime``` , ```maxTime``` are the starting and ending time of TSDB block.
```
bboreham: Nit: state what format the times should be in.  E.g. Unix timestamp.

prombench/docs/gke.md
```diff
@@ -96,8 +96,71 @@ Run Prombench tests in [Google Kubernetes Engine (GKE)](https://cloud.google.com
...
-3. **Deploy the Kubernetes Objects**:
+3. **Setting Up Benchmarking Data**    
+ When setting up a benchmarking environment, it’s often useful to have pre-generated data available.This data can help speed up testing and make benchmarks more realistic by simulating actual workloads.
+
+In this setup, you have two choices:
+
+Here’s how each option works:
+- **Option 1: Download data from object storage**
+
+   To download data from object storage, create a Kubernetes secret with exact named `bucket-secret` and file name `object-config.yml`  with the necessary credentials as per your object storage. This secret enables access to the stored data.
```
bboreham, line 107: When we run at GKE we use integrated authorization to allow pods to access the bucket; no secret needed. I guess it's ok to have a blank one for consistency, but I think we should mention this mode.

bboreham (changes requested):
Requiring the Secret and ConfigMap to be placed in a namespace including the PR number implies the operator has to create new ones every time a PR is benchmarked. And this has to be done in a short window of time after the namespace is created by automation but before it times out.
That isn't workable.

prombench/docs/gke.md
kushalShukla-web: do i have to mention this for both GKE and EKS ?

prombench/docs/gke.md
kushalShukla-web: yes bryan you are right !

prombench/docs/gke.md
kushalShukla-web: okay

Issue #704, opened 2024-07-23
Idea: include rules in PromBench load


bboreham:
We could create some recording rules and alerting rules to exercise that part of Prometheus from PromBench.

Sweetdevil144:
Hi @bboreham . I'd like to work on this issue for now.

Can you explain some 'recording rules' which would be good to go as a beginning approach?

bboreham:
Hi, thanks for your interest. 
If you are coming up to speed, the first step would be to get familiar with Prometheus and recording rules. 
Then, to start with you could use some of the queries in the load-generator in this repo. 

The main work for this issue is in configuration.

kushalShukla-web:
Hi @bboreham, I would like to work on this issue.

bboreham:
OK, that's great.  I think this is simple enough that you can just make a PR; no design doc required.  But if you've thought about it and want to propose a design, please do.

We might need to remove some queries from the load-generator to leave overall load balanced.

PR #854, opened 2025-02-22
 included rules in PromBench as a configmap

kushalShukla-web:
1. Added rules in PromBench to enable testing for that part of the benchmark.
2. Removed some queries from the load generator to balance the overall load.
3. Renamed the fake web server file for better file structure.


kushalShukla-web:
partially solves #704

kushalShukla-web:
Hi @bboreham  what do you think ?

PR #874, opened 2025-04-24
Included copy command to copy blocksync-config and bucket-secret from default to pr namespace

kushalShukla-web:
1. Included copy command so that testinfra binary can copy (blocksync-config and bucket-secret) from default namespace to pr-namespace that can be used by block-sync later.
Signed-off-by: Kushal Shukla <kushalshukla110@gmail.com>


kushalShukla-web:
cc: @bboreham , @bwplotka

kushalShukla-web:
I have tested this with only kind.

PR #887, opened 2025-06-13
Switch Parca storage backend to GCS

### LINKED ISSUE #630: Stabilize and extend Parca retention
I don't think we set any retention, but maybe it OOMs at some point? We see inconsistent Parca retentions, let's improve it.

E.g. running prombench for 4 days, yet only ~1h or data:

image

cc @bboreham @kakkoyun

metalmatze:
## Summary
- Migrates Parca's storage backend from temporary filesystem to Google Cloud Storage
- Enables persistent storage of profiling data across pod restarts
- Addresses the TODO comment about making storage persistent
- Partially fixes #630

## Changes
- Updated Parca ConfigMap to use GCS bucket configuration instead of filesystem
- Added volume mount for GCS service account credentials
- Removed temporary emptyDir storage volume
- Added documentation explaining the required external secret setup

## Configuration Requirements
This deployment requires a Kubernetes secret named `parca-gcs-service-account` containing a GCS service account key JSON file. The secret should be created separately with appropriate permissions to read/write to the configured GCS bucket (`prometheus-parca-profiles`).

### DIFF
--- a/prombench/manifests/cluster-infra/8b_parca.yaml
+++ b/prombench/manifests/cluster-infra/8b_parca.yaml
@@ -1,15 +1,31 @@
+# Parca deployment with GCS storage backend
+# 
+# IMPORTANT: This deployment requires a Kubernetes secret named 'parca-gcs-service-account'
+# containing a GCS service account key JSON file. The secret must have the following structure:
+#
+# apiVersion: v1
+# kind: Secret
+# metadata:
+#   name: parca-gcs-service-account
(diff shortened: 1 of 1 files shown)


metalmatze:
I'm not sure I can create the necessary secret. I would need some more time to check this. 
The config should be good from what I can see. The code was generated by Claude, but I double-checked. It may need some finagling once we have the secret.

PR #937, opened 2026-02-06
feat: Improved prombench dashboard

bwplotka:
See preview http://prombench.prometheus.io/grafana/d/afbm2esx87klce/prombench-atbwplotka?orgId=1&var-RuleGroup=All&var-prNumber=18013

Thanks @bboreham for brainstorming session on what to improve! 

Changes:
* Dashboard starts with `Acceptance` rows for Mem, CPU and HTTP latency for quick, ideally single number that shows high level look on important performance numbers.
* Functional check (collapsed) is a set of old panels that help to tell if load, sample/s, series match (are we consuming same amount of data)
   * load generator panel has step 61s to avoid aliasing
* Performance Details is a set of old panels for more drill downs.  Added extra GC metrics.
* Query Latency: details row drill down stages of query engine as we had in the past
* Panels we do not understand or they don't work or they were useful only for a single issue are now in `Attic` panel. We can remove them later unless they will be objections

This dashboard could be better and more perfect, but let's just try iterate etc (:

### DIFF
--- a/prombench/manifests/cluster-infra/dashboards/prombench.json
+++ b/prombench/manifests/cluster-infra/dashboards/prombench.json
@@ -1,4092 +1,6050 @@
 {
-  "__requires": [
-    {
-      "type": "grafana",
-      "id": "grafana",
-      "name": "Grafana",
-      "version": "5.0.0"
-    },
-    {
-      "type": "datasource",
-      "id": "prometheus",
-      "name": "Prometheus",
-      "version": "5.0.0"
-    }
-  ],
-  "annotations": {
-    "list": [
-      {
-        "builtIn": 1,
-        "datasource": "-- Grafana --",
-        "enable": true,
-        "hide": true,
-        "iconColor": "rgba(0, 211, 255, 1)",
-        "name": "Annotations & Alerts",
-        "type": "dashboard"
-      }
-    ]
-  },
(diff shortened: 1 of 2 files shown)


bboreham (approved):
Thanks for moving this forward.

bwplotka:
Unfortunately our dashboard is too large for a config map 🙃 

```
kubectl apply -f prombench/manifests/cluster-infra/grafana_dashboard_dashboards_noparse.yaml 
The ConfigMap "grafana-dashboards" is invalid: metadata.annotations: Too long: may not be more than 262144 bytes
➜  test-infra git:(dashboard) ✗ ls -l prombench/manifests/cluster-infra/grafana_dashboard_dashboards_noparse.yaml 
-rw-r--r--@ 1 bwplotka  primarygroup  336526  6 Feb 15:26 prombench/manifests/cluster-infra/grafana_dashboard_dashboards_noparse.yaml
➜  test-infra git:(dashboard) ✗ ls -lh prombench/manifests/cluster-infra/grafana_dashboard_dashboards_noparse.yaml 
-rw-r--r--@ 1 bwplotka  primarygroup   329K  6 Feb 15:26 prombench/manifests/cluster-infra/grafana_dashboard_dashboards_noparse.yaml
```

bwplotka:
TODO shard dashboards due to K8s limits. I guess minifying might help too ;p

roidelapluie:
#941

bwplotka:
Updated, thanks to @roidelapluie !

PR #938, opened 2026-02-06
feat: Configure Parca to gather fpgprof profiles

bwplotka:


bwplotka:
Turning off, it's not very useful with delta and on continuation basis... kills perf too 
image

PR #940, opened 2026-02-09
feat: add support for building the "release" Prometheus from source

bwplotka:
This PR adds ability to compare Prometheus PR against anything in Prometheus repo or quay.io/prometheus/prometheus

It does change the default semantics when doing the typical `/prombench main`. In the past it was taking the image with the `main` tag (while PR Prometheus always rebuild from source, from the latest git commit on the GitHub PR), now it builds from source for the given `string` that can be git SHA, branch name or even PR number. Extra flag `bench.use-registry=true` has to be added to force using docker image (previous behavior).

# Motivation

* Reduce unknowns: 
  * It's not uncommon to do `/prombench main` and realize you didn't rebase recently and a few commits are in the REF vs current PR. Even with recent rebase you will likely have a different commits as there's a lag between `main` docker tag and the latest `main` commit. **We could autodetect the shared commit vs main or even merge with main and then benchmark in the next iteration**.
  * There's slight difference in Go version I noticed when building from scratch (PR) vs taking from the image.

image

* Enables benchmarking on feature branches (e.g. measuring difference in perf between feature branch and a PR)
* Enable comparing across two PRs

# Next Steps

* [ ]  Once approved if we want this, I have to deploy and test this.

### DIFF
--- a/tools/prometheus-builder/build.sh
+++ b/tools/prometheus-builder/build.sh
@@ -22,7 +22,7 @@ fi
 
 # Fetch from quay if requested.
 if [[ "${USE_REGISTRY}" == "true" ]]; then
-    echo ">> USE_PRE_BUILD is enabled."
+    echo ">> USE_REGISTRY is enabled."
     echo ">> Attempting to download binary from quay.io/prometheus/prometheus:${REFERENCE}"
 
     IMAGE="quay.io/prometheus/prometheus:${REFERENCE}"
--- a/tools/prometheus-builder/build.sh
+++ b/tools/prometheus-builder/build.sh
@@ -1,34 +1,82 @@
 #!/bin/bash
 
+# This script created a Prometheus binary to use in benchmarks and place it in $VOLUME/prometheus.
+# It uses $REFERENCE variable to build or fetch the binary.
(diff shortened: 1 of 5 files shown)


krajorama:
The release doc will have to be updated to reflect the change https://github.com/prometheus/prometheus/blob/main/RELEASE.md

tools/prometheus-builder/build.sh
```diff
@@ -1,34 +1,82 @@
...
+fi
+
+if [[ -z ${REFERENCE} || -z ${VOLUME_DIR} || -z ${GITHUB_ORG} || -z ${GITHUB_REPO} ]]; then
+    echo "ERROR:: environment variables not set correctly, requires REFERENCE (or PR_NUMBER), VOLUME_DIR, GITHUB_ORG, GITHUB_REPO"
     exit 1;
 fi
 
+# Fetch from quay if requested.
+if [[ "${USE_REGISTRY}" == "true" ]]; then
+    echo ">> USE_PRE_BUILD is enabled."
```
krajorama:
```suggestion
    echo ">> USE_REGISTRY is enabled."
```

krajorama (approved):
Looks reasonable.

bwplotka:
TODO deploy and test before merging

PR #942, opened 2026-02-25
Update node-metrics dashboard.

SuperQ:
* Use `$__rate_interval` to better auto-scale graph data.
* Use `rate()` instead of `irate()` to avoid misleading graphs.
* Fix node-metrics dashboard panel types.


prombench/manifests/cluster-infra/dashboards/prombench.json
```diff
@@ -143,7 +143,7 @@
                         "uid": "P86B77AEF8F12A5FA"
                     },
                     "editorMode": "code",
-                    "expr": "sum(rate(go_memstats_alloc_bytes_total{job=\"prometheus\",namespace=\"prombench-${prNumber}\",prometheus=~\"test-pr.*\"}[10m])) - sum(rate(go_memstats_alloc_bytes_total{job=\"prometheus\",namespace=\"prombench-${prNumber}\",prometheus=~\"(test+).*\", prometheus!~\"test-pr.*\"}[10m]))",
```
bwplotka, line 146: Those average graphs are specifically 10m as per description for stable and predictable results. We could change other sections (you only change 10m ones?), but all those `10m` ideally stay as 10m.

>         "description": "Delta of number of bytes allocated per second (go_memstats_alloc_bytes_total), even if freed. PR minus Reference, so negative values indicate how much less PR allocates.\n\n10m lower than our periodic scrape load scale ups and downs cycles (~30m interval) but large enough to hide noise.",

bwplotka (changes requested):
For node metrics it's fine, but for "acceptance" sections in prombench.json, we specifically made it 10m with @bboreham - can we keep those unchanged?

We can change everything else in prombench.json to dynamic interval, just not those acceptance sections.

SuperQ:
Sure, I can put the 10m graphs back. Maybe what we should do is add acceptance test recording rules and use those.

SuperQ:
Ping @bwplotka

bwplotka (approved):
Thanks!

Remember that we need to manually redeploy ):

PR #947, opened 2026-03-05
prombench: fix initContainer ready URL to include web route prefix

roidelapluie:
The PR Prometheus is started with --web.external-url which sets the route prefix to /<PR_NUMBER>/prometheus-pr. The /-/ready endpoint is therefore served at /<PR_NUMBER>/prometheus-pr/-/ready, not /-/ready.

### DIFF
--- a/prombench/manifests/prombench/benchmark/3_prometheus-test-release_deployment.yaml
+++ b/prombench/manifests/prombench/benchmark/3_prometheus-test-release_deployment.yaml
@@ -769,7 +769,7 @@ spec:
         - /bin/sh
         - -c
         - |
-          until wget -qO- http://prometheus-test-pr-{{ .PR_NUMBER }}/-/ready; do
+          until wget -qO- http://prometheus-test-pr-{{ .PR_NUMBER }}/{{ .PR_NUMBER }}/prometheus-pr/-/ready; do
             sleep 1
           done
       containers:


bwplotka (approved):

prombench/manifests/prombench/benchmark/3_prometheus-test-release_deployment.yaml
```diff
@@ -769,7 +769,7 @@ spec:
         - /bin/sh
         - -c
         - |
-          until wget -qO- http://prometheus-test-pr-{{ .PR_NUMBER }}/-/ready; do
+          until wget -qO- http://prometheus-test-pr-{{ .PR_NUMBER }}/{{ .PR_NUMBER }}/prometheus-pr/-/ready; do
```
bwplotka, line 772: hm, why not using K8s DNS?

bwplotka (approved):
LGTM, just doing it through local network likely makes more sense?

prombench/manifests/prombench/benchmark/3_prometheus-test-release_deployment.yaml
roidelapluie: what do you mean?

roidelapluie: This is using k8s dns, but prometheus API has an prefix, even when called internally.

PR #946, opened 2026-03-05
prombench: add initContainer to wait for PR Prometheus readiness

roidelapluie:
Before starting the release Prometheus, wait until the PR Prometheus instance is ready by polling its /-/ready endpoint.

PR #948, opened 2026-03-06
Set CPU limit and make init containers really wait for each other

roidelapluie:
Setting limit:4 on CPU and starting pods at the same time would make workflows more comparable.

#946 unforunately was a no-op since the apply job applied resources one after each other.


prombench/manifests/prombench/benchmark/5_prometheus-test-pr_deployment.yaml
```diff
@@ -777,6 +777,20 @@ spec:
         volumeMounts:
         - name: prometheus-executable
           mountPath: /prometheus-builder
+      - name: sync-with-release-prometheus
+        image: quay.io/prometheus/busybox:latest
+        imagePullPolicy: Always
+        command:
```
bwplotka, line 787: Please add comment - it's not super obvious what's this script is doing

```suggestion
        # Open connection on Prometheus port and attempt to connect the second instance.
        # When the connection is possible, unblock the init container.
        # This is to synchronize Prometheus instances for more reliable comparisons.
        command:
```

prombench/manifests/prombench/benchmark/5_prometheus-test-release_deployment.yaml
```diff
@@ -762,14 +762,18 @@ spec:
       securityContext:
         runAsUser: 0
       initContainers:
-      - name: wait-for-pr-prometheus
+      - name: sync-with-pr-prometheus
         image: quay.io/prometheus/busybox:latest
         imagePullPolicy: Always
         command:
```
bwplotka, line 772: Please add comment - it's not super obvious what's this script is doing

```suggestion
        # Open connection on Prometheus port and attempt to connect the second instance.
        # When the connection is possible, unblock the init container.
        # This is to synchronize Prometheus instances for more reliable comparisons.
        command:
```

bwplotka (approved):
👍🏽  Can you add some quick commentary? Otherwise LGTM.

prombench/manifests/prombench/benchmark/5_prometheus-test-pr_deployment.yaml
bwplotka, line 787: hm, what about synchronizing other resources? 

Also maybe let's add TODO to add a bash script for this we can maintain/test later

roidelapluie: With the new "waves" system this should behave as expected.

bwplotka (approved):
Amazing, thanks!

PR #949, opened 2026-03-06
prombench: use variables for cleanup file references in Makefile

roidelapluie:
Introduce override variables for the cleanup namespace and cluster-role-binding filenames so they are defined in one place. Update the README to reflect the current filenames and add a check_config target to validate the referenced files exist.

### DIFF
--- a/prombench/Makefile
+++ b/prombench/Makefile
@@ -1,6 +1,11 @@
 INFRA_CMD  ?= ../infra/infra
 PROVIDER   ?= gke
 
+# Files used in resource_delete for cleanup. These must match actual filenames
+# in the benchmark directory.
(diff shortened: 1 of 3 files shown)


prombench/Makefile
```diff
@@ -1,6 +1,11 @@
 INFRA_CMD  ?= ../infra/infra
 PROVIDER   ?= gke
 
+# Files used in resource_delete for cleanup. These must match actual filenames
+# in the benchmark directory.
+override CLEANUP_NAMESPACE_FILE            := 1_namespace.yaml
```
roidelapluie, line 6: override because that should not be overriden by users ; reducing surface

SuperQ (approved):

PR #950, opened 2026-03-06
prombench: move node-exporter to wave 6 after prometheus deployments

roidelapluie:
The node-exporter deployments for test-main and test-pr use podAffinity requiring prometheus pods to already exist. Moving node-exporter from wave 4 to wave 6 ensures prometheus (wave 5) is deployed first.


bwplotka (approved):
Thanks, got hit by this

PR #951, opened 2026-03-09
prombench: fix scaler broken by wave extraction and add wait parameter

roidelapluie:
extractWave previously returned an error for filenames without an N_ prefix, causing every ResourceApply call in the scaler to fail silently. The scaler's config file (/etc/scaler/webserver.yaml) has no wave prefix by design, so change extractWave to return 0 instead of an error for files without a numeric prefix, matching the function's documented behavior.

Also add a wait bool parameter to ResourceApply so callers can opt out of the per-wave readiness check. The scaler passes false since it only needs to patch replica counts and sleep; the infra providers (GKE, EKS, KIND) pass true to preserve the existing wave-ordered readiness behavior.


bwplotka (approved):
Amazing thanks!

PR #956, opened 2026-03-12
prombench: add support for extra flags on Prometheus instances

roidelapluie:
Add --pr.extra-flags and --main.extra-flags options to the /prombench command, allowing callers to pass arbitrary flags to the PR and main/release Prometheus instances respectively.

Use github.com/google/shlex for proper shell tokenization of the command line, replacing the naive strings.Split approach. Also fix flag value parsing so that flag values containing = signs are handled correctly.

### DIFF
--- a/pkg/provider/provider.go
+++ b/pkg/provider/provider.go
@@ -127,17 +127,37 @@ func RetryUntilTrue(name string, retryCount int, fn func() (bool, error)) error 
 	return fmt.Errorf("Request for '%v' hasn't completed after retrying %d times", name, retryCount)
 }
 
+// prometheusArgs builds a YAML flow-sequence injection string starting with
+// baseArg, followed by any space-separated flags from extraFlags. The returned
+// value is intended to be placed inside a double-quoted YAML string: the
+// embedded "," separators will break into additional flow-sequence items after
+// template rendering, e.g. "--log.level=info","--flag1","--flag2".
+func prometheusArgs(baseArg, extraFlags string) string {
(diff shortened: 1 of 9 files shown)


pkg/provider/provider.go
```diff
@@ -127,17 +127,37 @@ func RetryUntilTrue(name string, retryCount int, fn func() (bool, error)) error
...
+		// k8s objects can't have dots(.) so we add a custom function to allow normalising the variable values.
 		"normalise": func(t string) string {
 			return strings.ReplaceAll(t, ".", "-")
 		},
-		"split": func(rangeVars, separator string) []string {
-			return strings.Split(rangeVars, separator)
+		// prArgs returns a YAML flow-sequence injection string for the PR prometheus
+		// instance, including --log.level=info and any PR_EXTRA_FLAGS.
+		"prArgs": func(vars map[string]string) string {
+			return prometheusArgs("--log.level=info", vars["PR_EXTRA_FLAGS"])
```
bwplotka, line 155: Kind of confusing we hardcode info here. Why not removing this and allowing this option to set it? (info is default)

	a.Flag(LevelFlagName, LevelFlagHelp).
		Default("info").HintOptions(promslog.LevelFlagOptions...).
		SetValue(config.Level)

bwplotka (approved):
We have already a way for this - bench.version.

But this is bit faster indeed, so LGTM. Just please remove log level setting, we don't need that and would unblock new use case

PR #957, opened 2026-03-16
prombench: handle rolling restarts in Prometheus sync init containers

roidelapluie:
Skip peer sync when the other Prometheus is already up (rolling restart case) by checking /-/ready. Otherwise, synchronize on port 9090 instead of 9099 and remove the dedicated sync service port.


krajorama (approved):
LGTM, but please test

PR #1045, opened 2026-09-10
Add remote write receiver benchmark

jcreixell:
- Uses avalanche to generate remote write synthetic load
  - Enables RW receiver and relabel configs for assessing the performance of different candidate implementations
  - Specifically designed to test https://github.com/prometheus/prometheus/pull/19675


bwplotka (changes requested):
We can't merge it to master as this feature is not merged yet (PR version will be used for all prombench experiments).

However, we can merge it to another branch and use [`--bench.version`](https://github.com/prometheus/prometheus/pull/19675#issuecomment-5678929444) to use it if you want.

Created `rwreceive` branch if you want.

jcreixell:
Thanks @bwplotka, I just created https://github.com/prometheus/test-infra/pull/1048 so we can run the tests

